PR #914, opened 2013-03-16
there are no Deprecation Warnings

tshepang:
This is practically a revert of commit 50fb1030b, or at least not anymore. I tested with Python 3.3 and 2.7.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,9 +15,6 @@ import sys
 import os
 import datetime
 import inspect
-import warnings
-
-warnings.simplefilter('ignore', DeprecationWarning)
 
 # skip raw nodes
 from sphinx.writers.text import TextTranslator


hyperknot:
Just run into a similar issue, but instead of opening a ticket I thought of asking it here first, as it might be related.

With Python 2.7, there are absolutely no deprecation warnings displayed when using Pyramid. It took me some effort to manually somehow make them appear. My only solution was to use

    python -Wd `which pserve` development.ini

instead of simply pserve. That `which` in there is super-hacky though, and non-Windows as well.

So my problem is that deprecation warnings simply do not appear in 2.7, unless the user puts a lot of effort to make sure that they are displayed. But the point of a warning is that you'd like to warn unsuspecting users, which is not happening here, as I'd guess 99% of Python 2.7 users simply never see any warning.

mmerickel:
We could make an effort to turn deprecation warnings on by default in pserve, silenced only by a flag. It feels weird to go against the interpreter defaults however.

hyperknot:
Then maybe printing of warnings should use a different method. (I think Django does something similar.)

Also, there is a problem of having the following lines always printed:
```
/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/Users/user/.virtualenvs/gigamacro_web/lib/python2.7/site-packages/zope': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/Users/user/.virtualenvs/gigamacro_web/lib/python2.7/site-packages/repoze': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/Users/user/.virtualenvs/gigamacro_web/lib/python2.7/site-packages/paste': missing __init__.py
  file, filename, etc = imp.find_module(subname, path)
```


This appears 2x when using `--reload`.

mmerickel:
I'm open to suggestions on alternative methods. Please open a new issue to propose something though.

hyperknot:
Sorry, I've just checked and edited my answer, I think Django uses the default warning framework, but it's displayed by default.

PR #1028, opened 2013-06-03
Add a parent_domain option for auth_tkt policy

wichert:
This change adds a new `parent_domain` option to `AuthTktAuthenticationPolicy` which sets the authentication cookie as a wildcard cookie on the parent domain. This is useful if you have multiple sites sharing the same domain.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -15,6 +15,10 @@ Features
   ``pyramid.config.Configurator.add_static_view``. This allows
   externally-hosted static URLs to be generated based on the current protocol.
 
+- The ``AuthTktAuthenticationPolicy`` has a new ``parent_domain`` option to
+  set the authentication cookie as a wildcard cookie on the parent domain. This
+  is useful if you have multiple sites sharing the same domain.
+
 - The ``AuthTktAuthenticationPolicy`` now supports IPv6 addresses when using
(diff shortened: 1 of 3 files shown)


landreville:
Just wanted to note that this has the same problem as an earlier pull request I submitted: https://github.com/Pylons/pyramid/pull/450. When the domain has a multi-part public suffix such as "example.co.uk" it will set the cookie on ".co.uk" instead of the correct "example.co.uk".

Here is an example test that fails:
```
def test_remember_parent_domain_multipart_suffix(self):
        helper = self._makeOne('secret', parent_domain=True)
        request = self._makeRequest()
        request.domain = 'example.co.uk'
        result = helper.remember(request, 'other')
        self.assertEqual(len(result), 1)

        self.assertEqual(result[0][0], 'Set-Cookie')
        self.assertTrue(result[0][1].endswith('; Domain=example.co.uk; Path=/'))
        self.assertTrue(result[0][1].startswith('auth_tkt='))
```

bertjwregeer:
@landreville Unfortunately there is not much that Pyramid can do. We don't want to depend on hardcoding the list of "top-level" domain names, nor do we want to depend on packages that do this already.

mmerickel:
This is really the responsibility of the developer to make sure their app is not doing this. They know where it will be hosted, and have control over the settings.

PR #1362, opened 2014-06-13
Update util.py

amyalichkin:


mmerickel:
Thanks for this change. We fixed it finally in #1556.

PR #1370, opened 2014-07-01
Query string arg names starting with equal sign

timka:
I use %3D (URL encoded equal sign) prefix in argument name for filter parametrization:

    GET /users?%3Dphone_number=1234567

Since request_param supports matching specific values with equal sign syntax, this doesn't work when route registered with explicit filter args names in request_param.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -232,3 +232,5 @@ Contributors
 - Amit Mane, 2014/01/23
 
 - Fenton Travers, 2014/05/06
+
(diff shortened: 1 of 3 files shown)


sontek:
@timka Did you see @mmerickel's request?

timka:
Sorry for the delay

mmerickel:
Hey @timka, I'm sorry I didn't mention this earlier. Can you please add another commit with your name in CONTRIBUTORS.txt?

timka:
No prob

PR #1378, opened 2014-07-14
Exceptions with Accept-Type application/json will return JSON type error.

nl5887:
* had to implement the server_quality because of failing tests. text/html is preferred. (test test__default_app_iter_with_comment_html)
* because of the server_quality I had to check if the match was in the range of html_types or json_types (using the tuple)
* the json is returning immediately, because no use of "templated" arguments and adding comments is invalid to json spec
* the check for accept at html_types is because of compatibility (with tests), when no accept exists, plain will be default.
* moved the import json to above the pyramid imports
* some fixes becaus of python3 compatibility

### DIFF
--- a/pyramid/httpexceptions.py
+++ b/pyramid/httpexceptions.py
@@ -126,10 +126,14 @@ from string import Template
 from zope.interface import implementer
 
 from webob import html_escape as _html_escape
+from webob.acceptparse import MIMEAccept
+
+import json
 
 from pyramid.compat import (
     class_types,
     text_type,
+    string_types,
     binary_type,
     text_,
     )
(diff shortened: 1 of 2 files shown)


pyramid/httpexceptions.py
```diff
@@ -203,6 +207,9 @@ class HTTPException(Response, Exception):
     ## Set this to True for responses that should have no request body
     empty_body = False
 
+    html_types = [('text/html', 1)] # add more here?
+    json_types = [('application/json',1)] # add more here?
+
```
sontek: @nl5887 Have you had time to look at what the issue was?

mmerickel:
this PR will be replaced to mirror the changes made to webob's exceptions https://github.com/Pylons/pyramid/issues/2406

Issue #1357, opened 2014-06-03
pcreate creates project named pyramid


sirex:
I just tried Pyramid using `pcreate`. `pcreate --help` says:

> Usage: pcreate [options] output_directory

So I decided to use `.` as my `output_directory`, and `pcreate` assumed, that my project name is same as name of `.` directory, which happens to be `pyramid`.

It would be great if `pcreate` would be smart enough to avoid creating project with name `pyramid`. Or at lease there should be option, to specify different project name.

How to reproduce:

    $ virtualenv pyramid
    $ cd pyramid
    $ . bin/activate
    $ pip install pyramid
    $ pcreate -s starter .
    $ python setup.py develop
    Installed /tmp/pyramid/lib/python2.7/site-packages/pyramid_mako-1.0.2-py2.7.egg
    error: Installed distribution pyramid 0.0 conflicts with requirement pyramid>=1.2dev

I think this scenario is quite common for all, who just trying Pyramid and for some of these newcomers, this can be show stopper.

mmerickel:
Fixed by https://github.com/Pylons/pyramid/pull/1837

PR #1392, opened 2014-08-06
pcreate output_dir: base dir name 'pyramid' is forbidden

grzr:
Solution for issue https://github.com/Pylons/pyramid/issues/1357 where pcreate creates project named pyramid.

New behavior:
```
$ virtualenv pyramid
$ cd pyramid
$ . bin/activate
$ pip install pyramid
$ pcreate -s starter .
You must change output_dir. Base dir name 'pyramid' is forbidden.
```

### DIFF
--- a/pyramid/scaffolds/__init__.py
+++ b/pyramid/scaffolds/__init__.py
@@ -18,10 +18,6 @@ class PyramidTemplate(Template):
         misnamings (such as naming a package "site" or naming a package
         logger "root".
         """
-        if vars['package'] == 'site':
-            raise ValueError('Sorry, you may not name your package "site". '
-                             'The package name "site" has a special meaning in '
-                             'Python.  Please name it anything except "site".')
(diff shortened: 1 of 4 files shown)


mmerickel:
@grzr ping

mcdonc:
I think we should try to do an ``__import__`` on the name that is supplied instead of maintaining a manual blacklist, FWIW.

mcdonc:
I also think that when a conflict is detected via ``__import__`` that we should not unconditionally prevent the user from creating a project with that name; instead we should ask them whether they really want to use that name, and provide a flag (maybe ``-f``/``--force``) that turns off the asking behavior.

mmerickel:
After several pings it seems @grzr isn't going to maintain this patch so I'm going to close it. If someone wants to open another version addressing the comments, that'd be great.

mmerickel:
Replaced this PR with https://github.com/Pylons/pyramid/pull/1837

PR #1399, opened 2014-08-13
security docs enhancements

mmerickel:
This PR replaces #1397.

### DIFF
--- a/pyramid/scripts/proutes.py
+++ b/pyramid/scripts/proutes.py
@@ -4,7 +4,6 @@ import textwrap
 
 from pyramid.paster import bootstrap
 from pyramid.scripts.common import parse_vars
-from pyramid.config.views import MultiView
 
 
 PAD = 3
@@ -58,6 +57,8 @@ class PRoutesCommand(object):
         from pyramid.interfaces import IRouteRequest
         from pyramid.interfaces import IViewClassifier
         from pyramid.interfaces import IView
+        from pyramid.interfaces import IMultiView
+
         from zope.interface import Interface
         config_uri = self.args[0]
 
@@ -72,6 +73,7 @@ class PRoutesCommand(object):
             max_view = len('View')
 
             routes = mapper.get_routes()
+
             if not routes:
                 return 0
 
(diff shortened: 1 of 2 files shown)


sontek:
@mmerickel Would it make more sense to introduce `authenticed_principal` / `unauthenticated_principal` and deprecate the old user id ones instead?  To me it does seem quirky to implemented a token based authentication policy and label everything userid.

mmerickel:
I'd be happy to name it a `token` but I think naming it a `principal` is an anti-pattern. The default policies treat the token as a principal but that is a security vulnerability if you are not careful and I think it should be more explicit which principals people are allowing into their authorization system.

tseaver:
I don't get that:  [principal](http://en.wikipedia.org/wiki/Principal_%28computer_security%29) is the precise term for what the authn/authz machinery deals with.

mmerickel:
> They need to be identified and authenticated before they can be assigned rights and privileges over resources in the network. A principal typically has an associated identifier (such as a security identifier) that allows it to be referenced for identification or assignment of properties and permissions.

I'd say authz, not authn.

tseaver:
Principals are *first* "authenticated"/"identified";  only then can their authorization be tested.  We didn't make this jargon up:  it widely used in software security analysis.  If presenting a valid token as part of a request gets you access to otherwise protected resources, then a token *is* a principal:  that is what principal means.

mmerickel:
In my mind, it'd be `authenticated_principals` instead of `authenticated_token` (note the plural) because to me translating a single token into multiple principals (the relationship between `authenticated_userid` and `effective_principals` currently) is natural. I also like how `userid` is pretty simply to explain as an arbitrary `token` so we don't have to rename the API at all to clarify the behavior. Basically everyone using Pyramid is probably using that particular API and changing it just for docs seems not so great.

I genuinely think the distinction is useful in understanding the API. If you feel so strongly about it I won't complain if you want to change it. I've made my feelings pretty clear at this point. But I'd appreciate having some comments from someone else as well.

Issue #1398, opened 2014-08-13
using Accept-Language in the default locale negotiator


tisdall:
Is there a rational behind not using the `Accept-Language` values in a HTTP request as a fallback in the `default_locale_negotiator`?  I understand the preference of using a `_LOCALE_` value over the `Accept-Language` as there seems to be issues with people properly setting that value, but why not use that value when there's no `_LOCALE_` set?  I think a lot of browsers set that browser's language automatically on install based on the OS's settings so it typically defaults to a reasonable value.

I wanted to see if there was some reason for this before submitting any sort of pull request.

ztane:
We did the language list approach exactly like that (available_languages in .ini); however there is a problem with the matching that for example Chinese locales have complicated names, and we needed to match everything that could possibly come from the client with whatever our locale names were and it wasn't eventually pretty at all.

mmerickel:
This definitely belongs in a recipe instead of in core but it'd be really nice if someone could write up an example for the cookbook or even as part of the i18n chapter. For example update  http://docs.pylonsproject.org/projects/pyramid/en/1.6-branch/narr/i18n.html#detecting-available-languages

with a snippet negotiator that used `request.accept_language.best_match()` as you're describing and then fallback to `default_locale_negotiator` if no match was found.

ztane:
I do not have time now, but here is a starter for someone:

```
#: Mapping of language codes send by browsers to supported dialects.
BROWSER_LANGUAGES = {
    'en-ca': 'en_GB',
    'en-gb': 'en_GB',
    'en-us': 'en_US',
    'fi':    'fi_FI',
    'fi-fi': 'fi_FI',
    'en':    'en_GB',
    'zh':    'zh_CN',
    'zh-cn': 'zh_CN',
    'zh-hk': 'zh_HK',
    'zh-tw': 'zh_TW'
}


def locale_negotiator(request):
    """Locale negotiator for pyramid views.

This version differs from Pyramid's :py:func:`default_locale_negotiator
<pyramid.i18n.default_locale_negotiator>` in that it prefers the preferred
locale of the currently authenticated user, and has a fallback to the
preferred languages from browser.
"""
    locale = default_locale_negotiator(request)
    if locale is None and request.accept_language:
        locale = request.accept_language.best_match(BROWSER_LANGUAGES.keys()).lower()

        # default to en_GB as it does not have silly units etc...
        locale = BROWSER_LANGUAGES.get(locale, 'en_GB')
[...]

PR #1402, opened 2014-08-15
to demo i18n improvements

tisdall:
I don't seem to be getting my point across very well in #1398
so I'm throwing together this PR to demonstrate what I'm saying.

The default_locale_negotiator now looks at a
`pyramid.available_languages` value to know what possible languages
values are acceptable.  Also, it will now use that value to match
against `Accept-Language` http request headers if no `_LOCALE_` value
is found.

Backwards compatibility:  If no `pyramid.available_languages` is set
then the `_LOCALE_` value is returned as before or `None`.
(The `Accept-Language` is ignored as there's nothing to match
against)

Obviously this needs some work, docs, and tests...  I thought it'd
be easier to see what I was talking about with a solid example.

The `pyramid.available_languages` is in this form:

```
pyramid.available_languages = 
    en English
    en-US English (United States)
    fr French
    pr-BR Portuguese (Brazil)
```

### DIFF
--- a/pyramid/i18n.py
+++ b/pyramid/i18n.py
@@ -141,7 +141,8 @@ def default_locale_negotiator(request):
             i.split(None, maxsplit=1)
             for i in aslist(
                 request.registry.settings.get(
-                    'pyramid.available_languages'
+                    'pyramid.available_languages',
+                    []
                 ),
                 flatten=False
             )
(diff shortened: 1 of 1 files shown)


mmerickel:
What's the point of supporting the long name of the locale (English) versus just "en"? That's not actually used by `best_match` is it?

tisdall:
I think my purpose here was that I also used the structure to create an html `select` to allow people to dynamically change the language and the full label was for human display.  So, not really needed in this context.

...  Okay, I just looked at the linked issue #1398 and I explained it in there as such.

PR #1417, opened 2014-09-25
Support returning app_iter from renderer.

lrowe:
Sets `response.app_iter = result` for renderer results which are not text, bytes or None.

My motivating case here is that rendering very large json responses can lead to 'memory leaks' where python is unable to release the space required for the large string back to the OS. This causes process size to increase as it allocates first 100MB for a large json response, then later for a slightly larger json response, it allocates say 105MB, then 115MB... (These allocations are visible in /proc/<pid>/smaps.)

My hope is that by using using json serializer that returns an iterable these large string allocations will not be required:

```
class JSONResult(object):
    def __init__(self):
        self.app_iter = []
        self.write = self.app_iter.append

    @classmethod
    def serializer(cls, value, **kw):
        fp = cls()
        json.dump(value, fp, **kw)
        return fp.app_iter


json_renderer = pyramid.renderers.JSON(serializer=JSONResult.serializer)
```

### DIFF
--- a/pyramid/renderers.py
+++ b/pyramid/renderers.py
@@ -456,6 +456,10 @@ class RendererHelper(object):
         if result is not None:
             if isinstance(result, text_type):
                 response.text = result
+            elif isinstance(result, bytes):
+                response.body = result
+            elif hasattr(result, '__iter__'):
+                response.app_iter = result
             else:
                 response.body = result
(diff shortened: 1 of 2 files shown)


mmerickel:
Yay, thank you!

PR #1471, opened 2014-11-22
update tox/travis to check code coverage on py3

### LINKED ISSUE #1470: update tests to check coverage of py3 code instead of setting "no cover"
It's possible to run coverage on multiple targets and combine their results into a final analysis. We should try to do this to properly test coverage on py2 and py3.

http://nedbatchelder.com/code/coverage/cmd.html#cmd-combining
http://stackoverflow.com/questions/19274492/how-calculate-the-global-coverage

Related: https://github.com/Pylons/pyramid/issues/979

mmerickel:
Coverage is combined between py2 and py3 for an aggregate coverage
metric. This means we can stop putting "no cover" gates around py3 code
and ignoring whether it's ever tested.

This is the best way I could find to do this based on several resources [1,2]. I'd be ecstatic if someone has a cleaner way. The only negative I see of this approach is running the tests one final time to dump xunit and coverage.xml files.

Fixes #1470.

[1] http://blog.ionelmc.ro/2014/05/25/python-packaging/
[2] https://github.com/tuskar/python-tuskarclient/blob/master/tox.ini


mmerickel:
I'm open to some alternatives but this latest push 14126ca has the following outputs:

- `coverage-py26.xml`, `nosetests-py26.xml`
- `coverage-py27.xml`, `nosetests-py27.xml`
- `coverage-py32.xml`, `nosetests-py32.xml`
- `coverage.py33.xml`, `nosetests-py33.xml`
- `coverage.py34.xml`, `nosetests-py34.xml`
- `coverage.pypy.xml`, `nosetests-pypy.xml`
- `coverage.pypy3.xml`, `nosetests-pypy3.xml`
- `coverage.xml` (an aggregate of the project that is at 100%)

Main change required to our jenkins box is to read `nosetests-py26.xml` instead of `nosetests.xml` because that one is not generated any longer (it was always just py26 but the name has changed).

tseaver:
LGTM overall.  I think I would prefer to make `py27` the "canonical" PY2 version at this point, but we can tweak that later.

mmerickel:
I've thought about this a bit and the main change I want to make before this is merged is that the tests run dog-damn slow on travis when running coverage. I don't have a good idea how to fix it though. Tentative idea is:

- Turn off tests for most envs.
- Run final (slow) test on just py26,py32,cover with an env variable to turn on coverage.

mmerickel:
@tseaver can you please re-evaluate this based on my latest changes? Minus some quirks on tox syntax I'm pretty happy with this configuration. The outputs are the following:

- nosetests-py26.xml
- nosetests-py27.xml
- nosetests-py32.xml
- nosetests-py33.xml
- nosetests-py34.xml
- nosetests-pypy.xml
- nosetests-pypy3.xml
- coverage-py2.xml
- coverage-py3.xml
- coverage.xml (an aggregate of the project that is at 100%)

mmerickel:
I stole a bit of this from https://github.com/tornadoweb/tornado/blob/0b2ecd3f71619ad077d54a54bbf43b5be2b593b3/tox.ini

mcdonc:
I have no problem with any of this.

mmerickel:
@mcdonc jenkins will have to be fixed to use nosetests-py27.xml

From what I have read at some point you can actually have jenkins use multiple xml files to combine the results, but I cannot verify this.

tseaver:
Jenkins [did break](http://jenkins.pylonsproject.org/job/pyramid/1700/).  I tweaked its config to look for `nosetests-py27.xml`, and [re-ran to ensure it succeeds](http://jenkins.pylonsproject.org/job/pyramid/1701/).

Issue #600, opened 2012-05-10
Urldispatch chapter - assume scan


gjhiggins:
Modify the urldispatch chapter examples to assume a scan rather than add_view.

PR #1483, opened 2014-12-10
Revise URL Dispatch documentation to use config.scan() in Examples 1-3

zmarvel:
This is in response to issue #600.


mmerickel:
Alright, finally got around to looking at this. Thank you for updating these examples!

Issue #1391, opened 2014-07-31
`not_` fails on `accept` predicate


Treora:
Not sure if it's important, but the `not_` operator does not work on the `accept` predicate for a view configuration:
```AttributeError: 'not_' object has no attribute 'lower'```
([here] is the culprit)
I haven't checked other functions or predicates, just encountered this one.

[here]: https://github.com/Pylons/pyramid/blob/3587a53dc28b8f6411816ccd7fd8fdee0d88acb4/pyramid/config/views.py#L1122

mmerickel:
Yep, #1487 explicitly removes `not` support from accept, as it's not labelled as a predicate now.

PR #1487, opened 2014-12-15
Documentation/accept

bertjwregeer:
As a follow up to: #1391 and #1481 this is a documentation change that moves accept from a predicate to a non-predicate.

This then explicitly documents that `not_` will not work with it since it is not technically a predicate.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -135,6 +135,7 @@ Docs
   https://github.com/Pylons/pyramid/issues/1391 for a bug report stating
   ``not_`` was failing on ``accept``. Discussion with @mcdonc led to the
   conclusion that it should not be documented as a predicate.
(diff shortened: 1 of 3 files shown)


CHANGES.txt
```diff
@@ -130,6 +130,12 @@ Deprecations
 Docs
 ----
 
+- Moved the documentation for ``accept`` on ``Configurator.add_view`` to no
+  longer be part of the predicate list. See
+  https://github.com/Pylons/pyramid/issues/1391 for a bug report stating
+  ``not_`` was failing on ``accept``. Discussion with @mcdonc led to the
+  conclusion that it should not be documented as a predicate.
```
mmerickel: Can you reference this PR pretty please.

PR #1498, opened 2014-12-26
Alert the user if they are using an unbound class methods to `add_view`

sontek:
This prevents the following from giving an unrecognizable error:

``` python
from waitress import serve

class View(object):
    def __init__(self, request):
        self.request = request

    def hello_world(self):
        return Response('Hello world!')

def create_app():
    config = Configurator()
    config.add_route('hello_world', '/')
    config.add_view(View.hello_world, route_name='hello_world')

    app = config.make_wsgi_app()
    return app

if __name__ == '__main__':
    app = create_app()
    serve(app)

```

Before this PR you would get the exception:

```
  File "app.py", line 21, in <module>
    app = create_app()
  File "app.py", line 17, in create_app
    app = config.make_wsgi_app()
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/__init__.py", line 956, in make_wsgi_app
    self.commit()
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/__init__.py", line 625, in commit
    self.action_state.execute_actions(introspector=self.introspector)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/__init__.py", line 1084, in execute_actions
    tb)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/__init__.py", line 1076, in execute_actions
    callable(*args, **kw)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/views.py", line 1236, in register
    derived_view = deriver(view)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/views.py", line 159, in __call__
    view)))))))))
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/views.py", line 103, in inner
    wrapper_view = wrapper(self, view)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/views.py", line 171, in mapped_view
    mapped_view = mapper(**self.kw)(view)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/views.py", line 424, in __call__
    view = self.map_nonclass(view)
  File "/home/sontek/venvs/pyramid/src/pyramid/pyramid/config/views.py", line 462, in map_nonclass
    mapped_view.__text__ = object_description(view)
pyramid.exceptions.ConfigurationExecutionError: <type 'exceptions.AttributeError'>: 'instancemethod' object has no attribute '__text__'
  in:
  Line 15 of file app.py:
    config.add_view(View.hello_world, route_name='hello_world')
```


pyramid/compat.py
```diff
@@ -244,3 +244,20 @@ def unquote_bytes_to_wsgi(bytestring):
...
+def is_unbound_method(fn):
+    """
+    This consistently verifies that the callable is bound to a
+    class.
+    """
+    is_bound = is_bound_method(fn)
+
+    if not is_bound and inspect.isroutine(fn):
+        spec = inspect.getargspec(fn)
+        has_self = len(spec.args) > 0 and spec.args[0] == 'self'
```
mmerickel: Assuming they use the convention of "self" as the first function argument and that they never accidentally use it on something else. I'm all in favor of better feedback so long as it doesn't break someone's valid code. This is probably esoteric enough to not matter but I had to bring it up. Thoughts?

bertjwregeer: It's interesting that you brought this up, because when i first read over this code I had the same question, but I let it go because I thought that if someone isn't using `self`, they are writing very unpythonic code...

bertjwregeer:
:+1:

pyramid/compat.py
sontek: Yeah, and they will just continue getting the existing poor behavior if they don't use `self`

bertjwregeer: We can't try and save the user from themselves every time. I think with the fixes you (@sontek) added we have already helped out 95% of the people that might make a mistake, for the rest we will get a ticket or someone asking for help on IRC.

mmerickel:
@sontek can you merge master into this so I can hit the little green button?

sontek:
@mmerickel Its merged with master now

mmerickel:
Not a fan of you slowly pep-8ing the code in these PRs, but thanks anyway!

bertjwregeer:
All of the extra whitespace changes are noise and make it difficult to see the meat in the changeset.

Issue #1256, opened 2014-03-04
No public way to change the response factory


eevee:
`Request` has a `ResponseClass` class attribute, but it's not actually used by `Request.response`.  That uses the registered `IResponseFactory`.  I can change _that_ directly, but it seems like there should be a `config.set_response_factory`?

sontek:
In #1499 we've opted to ignore `ResponseClass` and introduce `config.set_response_factory`

mmerickel:
Fixed in #1499.

PR #1499, opened 2014-12-26
Expose response class

sontek:
This is a small clean-up to utilize the `ResponseClass` that is already available on the Request object.  This allows developers to define their own request factory with a new response class and have it used everywhere.

This is an attempt to help with #1256

As was suggested in the ticket, it might make more sense to just create a `set_response_factory` API and a `response_factory` kwarg of the configurator, but since the ResponseClass is already there it doesn't hurt to utilize it.  I can't think of a time you would want to create a response when you don't have a request which is why I opted to take this route instead.

### DIFF
--- a/pyramid/request.py
+++ b/pyramid/request.py
@@ -216,8 +216,8 @@ class Request(
         right" attributes (e.g. by calling ``request.response.set_cookie()``)
         within a view that uses a renderer.  Mutations to this response object
         will be preserved in the response sent to the client."""
-        response_factory = _get_response_factory(self.registry, self)
-        return response_factory()
+        response_factory = _get_response_factory(self.registry)
+        return response_factory(self)
 
     def is_response(self, ob):
         """ Return ``True`` if the object passed as ``ob`` is a valid
--- a/pyramid/request.py
+++ b/pyramid/request.py
@@ -10,7 +10,6 @@ from pyramid.interfaces import (
     IRequest,
     IResponse,
     ISessionFactory,
-    IResponseFactory,
     )
 
 from pyramid.compat import (
@@ -27,7 +26,10 @@ from pyramid.security import (
     AuthorizationAPIMixin,
     )
 from pyramid.url import URLMethodsMixin
-from pyramid.util import InstancePropertyMixin
+from pyramid.util import (
+    InstancePropertyMixin,
+    _get_response_factory,
+)
 
 class TemplateContext(object):
     pass
@@ -214,9 +216,7 @@ class Request(
         right" attributes (e.g. by calling ``request.response.set_cookie()``)
         within a view that uses a renderer.  Mutations to this response object
         will be preserved in the response sent to the client."""
-        registry = self.registry
-        response_factory = registry.queryUtility(IResponseFactory,
-                                                 default=Response)
+        response_factory = _get_response_factory(self.registry, self)
         return response_factory()
 
     def is_response(self, ob):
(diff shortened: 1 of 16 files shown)


sontek:
@mmerickel `set_request_factory` requires a class that implements the same interface as `Request`.  I'm fine with changing this for `set_response_factory` but they will work differently at that point

mmerickel:
We added `request` to `__json__(self, request)` on model objects. That's more pervasive and significant than adding `request` to the response factory and serves the same purpose of allowing the developer to access their settings. I'm just having a real difficult time coming up with a good reason not to require it for a public API. If we don't add it then the first person who wants it will be required to use the threadlocals. We don't want that.

mmerickel:
If we are super concerned about breaking someone using the current private API then we could add another interface for the new public api but I don't think it's necessary.

sontek:
@mmerickel Only thing I'm worried about at that point is the `ResponseClass` on the `Request` object. Would we just always ignore this class and only use the factory?

sontek:
@mmerickel  32cb805 turns the response factory into a factory that takes a request and drops support for using `ResponseClass` as an entrypoint for configuring your response class

bertjwregeer:
I like the idea of being able to pass the request to the `Response` class, and even if we can't see why it is needed, someone will find some sort of use for it.

Could you also add an entry to `CHANGES.txt`?

mmerickel:
@sontek yes I think we should ignore the fact that `Request.ResponseClass` and `Response.RequestClass` exist. They were inherited from webob and can disappear quietly imo.

mmerickel:
This PR looks great. Final request is to please either a) remove `_get_response_factory` or b) move `_get_response_factory` into the `pyramid.response` module to avoid any future circular imports between `pyramid.response` and `pyramid.util`. I'll happily merge afterward!

sontek:
@mmerickel I prefer having it because we can change the implementation in one place, I'll make the move though

mmerickel:
Tank you SO MUCH

pyramid/request.py
```diff
@@ -214,10 +213,8 @@ def response(self):
         right" attributes (e.g. by calling ``request.response.set_cookie()``)
         within a view that uses a renderer.  Mutations to this response object
         will be preserved in the response sent to the client."""
-        registry = self.registry
-        response_factory = registry.queryUtility(IResponseFactory,
-                                                 default=Response)
-        return response_factory()
+        response_factory = _get_response_factory(self.registry)
+        return response_factory(self)
```
marc1n: This is backward incompatibility (not mentioned in CHANGES.txt for 1.6 ), because response_factory can not be Request class now.

mmerickel: The impl of `_get_response_factory` is identical to this code and defaults to `Response` if the user has not registered their own factory. I see no incompatibility here.

mmerickel: I think maybe what you are referring to is changing the factory to accept the `request` object. Unfortunately `IResponseFactory` was a private API prior to 1.6 so we do not document those changes.

marc1n: Yes, I used in 1.5 IResponseFactory in order to set custom Response class. It seems there was no other way. 

How could I know that is private and how should I set custom Response class in 1.5?

mmerickel: Custom responses classes were not officially supported in 1.5. Pyramid documents all public APIs in the API section of the documentation.

Note the difference from http://docs.pylonsproject.org/projects/pyramid/en/1.5-branch/api/interfaces.html to http://docs.pylonsproject.org/projects/pyramid/en/1.6-branch/api/interfaces.html with regards to `IResponseFactory`.

PR #1524, opened 2015-01-08
Explicitly declare the PickleSerializer()

joulez:
The default currently for SignedCookieSessionFactory is the PickleSerializer. If this was to ever change tests will fail.
Looking forwardly to a JSONSerializer and related tests in a seperate PR but interested in any comments. Thanks.

### DIFF
--- a/pyramid/tests/test_session.py
+++ b/pyramid/tests/test_session.py
@@ -299,7 +299,9 @@ class TestBaseCookieSession(SharedCookieSessionTests, unittest.TestCase):
 class TestSignedCookieSession(SharedCookieSessionTests, unittest.TestCase):
     def _makeOne(self, request, **kw):
         from pyramid.session import SignedCookieSessionFactory
+        from pyramid.session import PickleSerializer
         kw.setdefault('secret', 'secret')
+        kw.setdefault('serializer', PickleSerializer())
         return SignedCookieSessionFactory(**kw)(request)
 
     def _serialize(self, value, salt=b'pyramid.session.', hashalg='sha512'):


tseaver:
I don't really see the point:  the Travis failure is in coverage for the following test in `SignedCookieSesstionFactory.__init__`:

     if serializer is None:
        serializer = PickleSerializer()

Now, if the PR changed the signature of `_makeOne` to take an optional `serializer` (defaulting to `None`) and pass it through to the class, I could at least see how it might be useful in the cases you are contemplating.

joulez:
Pickle serialization is the default method in pyramid and alot of tests are orientated around this (re `_serialize`. I suppose I'm wondering if it would be a good idea to be more explicit about what serialization methods are being used in the tests. 

If at any point in the future the serializer defaults to something else other than `PickleSerializer` a couple of these tests (that call `_serialize`) fail and/or if pyramid wishes to support and include other serializers (ie. `JSONSerializer`).

tseaver:
Sure, some tests would fail if the default changed:  that is an expected thing.  As I said, if you wanted to change `_makeOne` to allow passing a non-default seraiizer, that would be fine.  I just don't see the point in duplicating the hard-wiring of the default from the code itself into a test jig.

joulez:
Thanks for the clarity. I think I approached this from the wrong direction. Please review #1525 and perhaps a discussion on changing the default serializer (and refactoring tests for it) can be had another time. There is concern that pickle can be volatile and insecure. But a least a alternative serializer can be provided perhaps?

bertjwregeer:
@joulez An alternate serializer is available in `WebOb` and may be used if you wish, however we can't change it wholesale due to the contract of `ISession`.

PR #1525, opened 2015-01-09
Add support for a json serializer (JSONSerializer) to the session module...

joulez:
Do you think this would be a good idea to add to the library? The test is pretty much exactly like the DummySerializer except for the fact that the JSONSerializer is more pyramid specific (using pyramid.compat helpers). Thanks.

### DIFF
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -11,6 +11,7 @@ from zope.interface import implementer
 from webob.cookies import SignedSerializer
 
 from pyramid.compat import (
+    json,
     pickle,
     PY3,
     text_,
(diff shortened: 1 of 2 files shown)


mmerickel:
Thanks for this but unfortunately I cannot accept it. The `JSONSerializer` actually already exists in webob. It is not included in pyramid because the `ISession` interface defines a session as something pickleable. JSON is not this lenient and as such is only useable by individuals who know they and their addons won't need this feature.

joulez:
Sorry, didn't notice webob has a `JSONSerializer`. I agree with the current trend to move away from pickled data on both the client and/or server, I didn't think the `ISession` doc string was a mandate rather than just a comment on the current default - that should be deprecated IMO. However it's entirely your call. Thanks again for considering this issue.

bertjwregeer:
The `ISession` interface says that the data to be stored in the session has to be pickle-able. At this time that means there are a lot of Pyramid applications that are built on top of that assumption. Switching to JSON as a serialization would break all of those people.

joulez:
@bertjwregeer - That would not be the suggestion. Have you considered deprecation warnings?

bertjwregeer:
Generally when we deprecate something we want to have a replacement available. In this case no such thing exists. At this time so long as the signing key is not made public the implementation is perfectly safe for use within an application.

joulez:
OK it appears you are steadfast on this. I disagree with the reasons. Minimizing attack vectors is a Good Thing™

mmerickel:
It's not an attack vector until you can show me an exploit. The signing protects the contents of the pickle from tampering. The signed-pickled-cookie has been used for many years in beaker and has been vetted by security experts. If an exploit is available we will immediately change our stance but at this time it is not an attack vector to our knowledge.

PR #1533, opened 2015-01-20
Give pserve the ability top open server in browser

msabramo:
E.g.:

    pserve app.ini --browser

or:

    pserve app.ini -b

Cc: @sontek, @mmerickel, @mcdonc

### DIFF
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -21,9 +21,11 @@ import textwrap
 import threading
 import time
 import traceback
+import webbrowser
 
 from paste.deploy import loadserver
 from paste.deploy import loadapp
+from paste.deploy.loadwsgi import loadcontext, SERVER
 
 from pyramid.compat import PY3
 from pyramid.compat import WIN
@@ -121,6 +123,11 @@ class PServeCommand(object):
         dest='monitor_restart',
         action='store_true',
         help="Auto-restart server if it dies")
+    parser.add_option(
+        '-b', '--browser',
+        dest='browser',
+        action='store_true',
+        help="Open a web browser to server url")
     parser.add_option(
         '--status',
         action='store_true',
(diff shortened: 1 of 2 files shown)


pyramid/scripts/pserve.py
```diff
@@ -334,6 +341,15 @@ def serve():
...
                 self.out('Exiting%s (-v to see traceback)' % msg)
 
+        if self.options.webbrowser_open:
+            def open_browser():
+                context = loadcontext(SERVER, app_spec, name=app_name, relative_to=base,
+                        global_conf=vars)
+                url = 'http://{host}:{port}/'.format(**context.config())
+                time.sleep(1)
+                webbrowser.open(url)
+            threading.Thread(target=open_browser).start()
```
mmerickel: can we mark the thread as detached so that it doesn't block the process from terminating?

msabramo: Done.

mmerickel:
I'm amazed to see that the `run` method has `no cover` on it. We need to fix that in the future. Anyway if you could add a changelog entry I'll merge this.

mmerickel:
I'm going back and forth on the cli option names `-w` and `--webbrowser-open`. Is there any other tool that provides similar options? I'd lean toward like `--open-browser` or `-w` myself.

msabramo:
```
$ gem server --help
...
    -l, --launch[=COMMAND]           launches a browser window
                                     COMMAND defaults to 'start' on Windows
                                     and 'open' on all other platforms
```

msabramo:
```
$ ipython notebook --help
...
--browser=<Unicode> (NotebookApp.browser)
    Default: u''
    Specify what command to use to invoke a web browser when opening the
    notebook. If not specified, the default browser will be determined by the
    `webbrowser` standard library module, which allows setting of the BROWSER
    environment variable to override it.
...
```

msabramo:
Maybe `-b` or `--browser`?

mmerickel:
Yeah I can't think of something that `-b` would clobber right now, so it's fine with me.

msabramo:
Cool, updated.

mmerickel:
Awesome! Thanks!

marc1n:
This change does not respect --server-name option. It only works when server name is 'main' (default).

mmerickel:
@marc1n Sure, if you would like to submit a PR fixing this it'd be accepted.

PR #1541, opened 2015-01-22
Prevent DeprecationWarning from setuptools>=11.3

dairiki:
Recent versions of `setuptools` have deprecated passing an argument to `EntryPoint.load`.  This prevents the deprecation warning for setuptools >= 11.3.  (The warning may still be issued for setuptools>=10.2,<11.3 — I'm not sure about this, and not sure that it is worth fixing in any case.)

See the changelog entries for [setuptools 11.3](https://pythonhosted.org/setuptools/history.html#id8) and [setuptools 10.2](https://pythonhosted.org/setuptools/history.html#id13) for more on this.

This should probably also be applied to the 1.5-branch.  Let me know if you'd like a separate PR for that.

### DIFF
--- a/pyramid/path.py
+++ b/pyramid/path.py
@@ -341,9 +341,10 @@ class DottedNameResolver(Resolver):
         # See https://pythonhosted.org/setuptools/history.html#id8
         ep = pkg_resources.EntryPoint.parse('x=%s' % value)
         if hasattr(ep, 'resolve'):
-            return ep.resolve()  # setuptools>=10.2
+            # setuptools>=10.2
+            return ep.resolve()  # pragma: NO COVER
         else:
-            return ep.load(False)
+            return ep.load(False)  # pragma: NO COVER
 
     def _zope_dottedname_style(self, value, package):
         """ package.module.attr style """
--- a/pyramid/path.py
+++ b/pyramid/path.py
@@ -337,8 +337,13 @@ class DottedNameResolver(Resolver):
                 value = package.__name__
             else:
                 value = package.__name__ + value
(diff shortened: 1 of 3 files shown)


pyramid/path.py
```diff
@@ -337,8 +337,13 @@ def _pkg_resources_style(self, value, package):
                 value = package.__name__
             else:
                 value = package.__name__ + value
-        return pkg_resources.EntryPoint.parse(
-            'x=%s' % value).load(False)
+        # Calling EntryPoint.load with an argument is deprecated.
+        # See https://pythonhosted.org/setuptools/history.html#id8
+        ep = pkg_resources.EntryPoint.parse('x=%s' % value)
+        if hasattr(ep, 'resolve'):
```
tseaver: We need a `# pragma: NO COVER` on both legs of this `if` (straddling sucks).

dairiki: Ugh...  But done in ae16a63.  Let me know if you'd rather I rebase and merge the two commits.

mmerickel:
Hey @dairiki this looks great. Can you please add your name to the CONTRIBUTORS.txt?

dairiki:
I've just signed CONTRIBUTORS.txt and rebased these commits to the current master.

mmerickel:
Thank you!

dairiki:
Thanks @mmerickel !

Should this also be applied to the 1.5-branch?  (Would you like a separate PR for that?)

mmerickel:
@dairiki that'd be great. I swore I already backported it but I can't find it atm.

Issue #1544, opened 2015-01-24
Can't use `context` predicate with `ContextFound` subscriber


msabramo:
It seems like this should be possible. I'd like to stuff like:

```python
@subscriber(ContextFound, context=Role)
def add_context(event):
    print('*** ContextFound: %r ***' % event.request.context)
```

but I get:

```pytb
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/__init__.py", line 637, in commit
    self.action_state.execute_actions(introspector=self.introspector)
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/__init__.py", line 1096, in execute_actions
    tb)
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/__init__.py", line 1088, in execute_actions
    callable(*args, **kw)
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/adapters.py", line 55, in register
    order, preds, phash = predlist.make(self, **predicates)
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/util.py", line 156, in make
    raise ConfigurationError('Unknown predicate values: %r' % (kw,))
pyramid.exceptions.ConfigurationExecutionError: <class 'pyramid.exceptions.ConfigurationError'>: Unknown predicate values: {'context': <class 'kotti_gabriel.resources.Role'>}
  in:
  Line 80 of file /Users/marca/dev/git-repos/pyramid/pyramid/events.py:
    config.add_subscriber(wrapped, iface, **self.predicates)
```

msabramo:
Ah, I just noticed the following at http://docs.pylonsproject.org/docs/pyramid/en/latest/narr/hooks.html#subscriber-predicates:

> There are no default subscriber predicates. You must register one to use one.

mmerickel:
Yeah unfortunately the event system does not really have a standard api for the format of an event so it's difficult to write generic predicates. We could ship default ones but the request is not required to be on an arbitrary event object. So at this point it's left up to you to define any predicates you want to buy into and presumably they will only work on event types you wrote them for.

I think it'd be nice if the API allowed you to register a predicate for a particular type of event instead of allowing it to work with all events but atm that's how it is.

msabramo:
I wonder why there are no default subscriber predicates -- was this a conscious decision to omit them? Perhaps because not all predicates make sense for all events? In that case, maybe they should be implemented but be caveat emptor, or even better, documented.

msabramo:
[...]

PR #1545, opened 2015-01-24
Subscriber predicates can list events supported

msabramo:
A subscriber predicate can specify what events it supports by setting an attribute called `events` -- e.g.:

```python
class ContextSubscriberPredicate(object):
    events = (ContextFound,)
    ...


def includeme(config):
    config.add_subscriber_predicate('context', ContextSubscriberPredicate)
```

If an attempt is made to add a subscriber for an event that they don't support:

```python
@subscriber(BeforeRender, context=Role)
def test_context_found_subscriber(event):
    print('*** ContextFound: %r ***' % event.request.context)
```

then a `ConfigurationError` is raised - e.g.:

    pyramid.exceptions.ConfigurationExecutionError:
    <class 'pyramid.exceptions.ConfigurationError'>: Event
    <class 'pyramid.events.BeforeRender'> not supported by subscriber predicate
    <class 'kotti_gabriel.ContextSubscriberPredicate'>;
    supports: (<class 'pyramid.events.ContextFound'>,)
      in:
      Line 80 of file /Users/marca/dev/git-repos/pyramid/pyramid/events.py:
        config.add_subscriber(wrapped, iface, **self.predicates)

This solves a limitation that @mmerickel mentioned in #1544.

Cc: @mmerickel, @sontek

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -80,6 +80,11 @@ Features
 - Support keyword-only arguments and function annotations in views in
   Python 3. See https://github.com/Pylons/pyramid/pull/1556
 
+- Subscriber predicates can have an ``events`` attribute that will be checked
+  when adding a subscriber. A ``ConfigurationError`` will be raised when
+  attempting to use a subscriber predicate with an event that it doesn't
+  support. See https://github.com/Pylons/pyramid/pull/1545
+
 Bug Fixes
 ---------
 
--- a/docs/narr/hooks.rst
+++ b/docs/narr/hooks.rst
@@ -1555,12 +1555,33 @@ for example, the predicate can be used with subscribers registered for
 :class:`pyramid.events.NewRequest` and :class:`pyramid.events.ContextFound`
 events, but it cannot be used with subscribers registered for
 :class:`pyramid.events.ApplicationCreated` because the latter type of event
-has no ``request`` attribute.  The point being: unlike route and view
-predicates, not every type of subscriber predicate will necessarily be
-applicable for use in every subscriber registration.  It is not the
-responsibility of the predicate author to make every predicate make sense for
-every event type; it is the responsibility of the predicate consumer to use
-predicates that make sense for a particular event type registration.
+has no ``request`` attribute.
 
+Pyramid 1.6 adds the ability to raise an exception when a subscriber predicate
+is used for an event type that it doesn't support. To use this feature, add an
+``events`` attribute to the subscriber predicate with one or more event classes
+that the predicate supports. So the ``RequestPathStartsWith`` subscriber
+predicate shown earlier could be written like this:
 
+.. code-block:: python
+    :linenos:
 
+    class RequestPathStartsWith(object):
+        events = (NewRequest, ContextFound)
+
+        # The rest of the class is the same as before...
+
+    config.add_subscriber_predicate(
+        'request_path_startswith', RequestPathStartsWith)
+
+Now if you try to use the subscriber predicate for an unsupported event type,
+say ``ApplicationCreated``:
+
+.. code-block:: python
+    :linenos:
+
+    @subscriber(ApplicationCreated, request_path_startswith='/path')
+    def test_context_found_subscriber(event):
+        pass
+
+then a ``ConfigurationError`` will be raised.
--- a/docs/narr/hooks.rst
+++ b/docs/narr/hooks.rst
@@ -1557,28 +1557,36 @@ events, but it cannot be used with subscribers registered for
 :class:`pyramid.events.ApplicationCreated` because the latter type of event
 has no ``request`` attribute.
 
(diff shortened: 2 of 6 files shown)


msabramo:
Oops, forgot to update docs. Will try to do that later.

mmerickel:
Sooo... this is interesting.

I feel like if I had attempted this problem myself I would have added an extra argument to `add_subscriber_predicate(..., events=[ContextFound, INewResponse])` that would say "add this predicate only for these event types", instead of coupling the event types to the predicate itself. Did you consider this?

msabramo:
Yeah I did consider that briefly, but then I thought that it made more sense to put it in the class, so the code is all together and it's easier to see why the subscriber predicate only works with the declared events.

msabramo:
Just modified the documentation (`docs/narr/hooks.rst`) and `CHANGES.txt`.

msabramo:
Also in a subscriber predicate vein: PR #1546

msabramo:
I think I have a better idea for this. Basic idea is that instead of predicates listing the events they support, they list what they need the event object to have. For example, many predicates depend on `event.request` and some may depend on `event.request.context`. 

This is more extensible. If new events are added, we don't need to add them to a zillion predicates.

mmerickel:
Sounds like you need an Event->Predicate adapter.

msabramo:
Yeah I have some code locally where I take all the view/route predicates and adapt them into subscriber predicates. Then you get a bunch of useful subscriber predicates. 

I need to instrument the predicates with what they require and then have `add_subscriber` check the applicability.

msabramo:
Just updated the docs.

@mmerickel, @sontek: Please review.

msabramo:
Pyramid 1.6 is going to be pretty cool.

bertjwregeer:
Just going through this, and looking at the way this is implemented I can think of an example where the event data does not have standard attributes. The `BeforeRender` puts the event data into a dictionary, and thus there is no attribute named request. Could this be fixed, probably but it seems like it is asking for trouble especially since there is no requirements for events to contain certain attributes or attributes at all for that matter.

While a good idea, I am not sure that having a developer annotate the classes with a tuple of strings is a good idea either, and it seems to be error prone. We have the power of Zope at our disposal, we should use it instead. For example, when adding the subscriber, we check that the predicate has been registered, and then we check to see if the event we are registering for implements the required sauce for that predicate, otherwise we raise an error.

msabramo:
Are you thinking of something like marker interfaces? e.g.: `IEventWithRequest` ?

I had that idea but didn't pursue it because I'm not an expert at ZCA and this was a bit simpler and easy to get done. But I could take a bit more time to make it more robust.

mmerickel:
You want to adapt an arbitrary Event into something the Predicate expects. The list of strings on the predicate itself doesn't feel a) safe for refactoring or b) extensible by a third party who wants to reuse your predicate with their own event types or c) even worse, person A trying to use person B's event with person C's predicate. It'd be nice if person A could adapt B to C without having to edit either of them.

```
class ContextSubscriberPredicate(object):
    def __call__(self, event):
        # expect event.request to work

class BeforeRenderToContextSubscriberEvent(object):
    def __init__(self, event):
        self.request = event['request']

def includeme(config):
    config.add_subscriber_predicate('context', ContextSubscriberPredicate, adapters={IBeforeRender: BeforeRenderToContextSubscriberEvent})

    # or

    config.add_subscriber_predicate_adapter('context', IBeforeRender, BeforeRenderToContextSubscriberEvent)

@subscriber_predicate(IBeforeRender, context=Foo)
def dosomething(event):
    print event.request
```

I apologize for criticism with minimal solutions, I'm trying to come up with a way that feels better than a list of strings but it does get a little verbose when trying to let person A tie together something from B and something from C.

msabramo:
Just fixed the merge conflict. I'm going to look later at @bertjwregeer and @mmerickel's comments and try to make it more Zopey, more maintainable, etc.

mmerickel:
I'm not sure what to do with this PR but so far it just doesn't seem to be working out. This whole feature-detection thing between arbitrary events and predicates is definitely tricky and I'm afraid to over-engineer it when so few people seem to need it.

I personally think the correct API for this is `config.add_subscriber_predicate(name, factory, type_or_iface=Interface)`. You can then register the predicate for each event it should be available for individually. For example:

```python
config.add_subscriber_predicate('context', ContextPredicate, pyramid.interfaces.IBeforeRender)
config.add_subscriber_predicate('context', ContextPredicate, pyramid.interfaces.IContextFound)
```

Any changes necessary to handle `BeforeRender` events versus `ContextFound` events can be done by registering a slightly different factory. We could then consider registering some default handlers for the builtin events if an API like this exists.

Anyway, this work seems to have stagnated so I'm going to close the PR for now. I'll be happy to reopen it if you want to pursue an approach like this.

PR #1546, opened 2015-01-24
Add a subscriber_predicate decorator

msabramo:
The declarative version of the imperative way: `add_subscriber_predicate`

E.g.:

```python
    @subscriber_predicate('context')
    class ContextSubscriberPredicate(object):
            self.val = val

        def text(self):
            return 'context = %s' % (self.val,)

        phash = text

        def __call__(self, event):
            return isinstance(event.request.context, self.val)
```

Cc: @mmerickel, @sontek

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -77,6 +77,9 @@ Features
   output by showing the module instead of just ``__repr__``.
   See https://github.com/Pylons/pyramid/pull/1488
 
+- Added ``pyramid.events.subscriber_predicate`` decorator to register a class
+  as a subscriber predicate. See https://github.com/Pylons/pyramid/pull/1546
+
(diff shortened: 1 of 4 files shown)


mmerickel:
I guess we'd want `route_predicate_config` and `view_predicate_config` as well if we're gonna do this. I think the `_config` convention was started with `view_config`, `notfound_view_config`, etc so we may wanna keep going with it. It seems a little silly to me to make these decorators since it only takes the one parameter which is a string that may conflict, but if people think it's useful we go forward with it.

mmerickel:
Are there any plans to pursue creating decorators for the various predicate categories? Considering I asked about it a year ago with no response, I'm going to close this until more interest is expressed in the feature.

PR #1548, opened 2015-01-25
[Backport from 1.6] Give pserve the ability to open server in browser (pserve --browser)

msabramo:
E.g.:

    pserve app.ini --browser

or:

    pserve app.ini -b

### DIFF
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -345,7 +345,7 @@ class PServeCommand(object):
             def open_browser():
                 context = loadcontext(SERVER, app_spec, name=app_name, relative_to=base,
                         global_conf=vars)
-                url = 'http://{host}:{port}/'.format(**context.config())
+                url = 'http://127.0.0.1:{port}/'.format(**context.config())
                 time.sleep(1)
                 webbrowser.open(url)
             t = threading.Thread(target=open_browser)
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -21,9 +21,11 @@ import textwrap
 import threading
 import time
 import traceback
+import webbrowser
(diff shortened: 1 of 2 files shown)


pyramid/scripts/pserve.py
```diff
@@ -334,6 +341,17 @@ def serve():
                     msg = ''
                 self.out('Exiting%s (-v to see traceback)' % msg)
 
+        if self.options.browser:
+            def open_browser():
+                context = loadcontext(SERVER, app_spec, name=app_name, relative_to=base,
+                        global_conf=vars)
+                url = 'http://{host}:{port}/'.format(**context.config())
```
mmerickel: shouldn't this always be localhost? the "host" option is specific to an individual wsgi server and is usually the interface to listen on, not the ip address to access. Technically the port is also specific to the wsgi server but probably acceptable.

msabramo: Oh I don't know maybe? Host seems to always be 127.0.0.1 or 0.0.0.0 and those both work. 

What problem are you worried about?

mmerickel: http://0.0.0.0 works? Are you sure? I'm not so sure.

msabramo: I think you're right. 

http://superuser.com/questions/536156/how-do-i-get-0-0-0-0-to-resolve-to-localhost-when-browsing-a-url-that-contains-0

I think it works on my Mac but it seems this is not universal. 

I'll change it.

mmerickel:
Sweet, sorry I didn't see that you had pushed a fix. Github doesn't email me when you push.

msabramo:
No worries. Usually I post a comment when I update a PR but I guess I forgot this time.

PR #1550, opened 2015-01-30
Use 'functools.update_wrapper' to preserve attributes from wrapped.

tseaver:


tseaver:
Preserves `__name__` and `__module__`, in addition to `__doc__`.

Note that I updated the `noinst` test to take an actual `wrapped` function (the point of that test was not to test decorating `None`, which is pointless, but to exercise the case where `self` is None (accessed via the class, rather than an instance).

Issue #1552, opened 2015-02-02
View selection breaks when using multiple base classes


wichert:
rest-toolkit uses ABCs to add default behaviour to resources. For example if a resource class adds the ``ViewableResource`` abc to its base classes the default GET view from rest-toolkit will be used for that resource. Under the hood this is implemented by adding views for the ABC to the route.  The user can then also add a view for the exact resource class to override the default view.

Unfortunately this system breaks if a resource uses multiple ABCs  as well as a non-abc base class. This is a test case that demonstrates this:

```python
import abc
from rest_toolkit.compat import add_metaclass  # This is a copy of six.add_metaclass
from webtest import TestApp
from pyramid.config import Configurator


@add_metaclass(abc.ABCMeta)
class Int1(object):
    pass


class OtherBase(object):
    pass


@add_metaclass(abc.ABCMeta)
class Int2(object):
    pass


class Resource(OtherBase, Int1, Int2):
    def __init__(self, request):
        pass


def unknown(context, request):
    return 'unknown method'


def view(context, request):
    return 'hello'


def test_multiple_abcs():
    config = Configurator()
    config.add_route('root', '/', factory=Resource)
    config.add_view(unknown, route_name='root', renderer='string')
    config.add_view(view, route_name='root', context=Int1, request_method='GET', renderer='string')
    config.add_view(view, route_name='root', context=Int2, request_method='POST', renderer='string')
    app = TestApp(config.make_wsgi_app())
    assert 'hello' in app.get('/')
    assert 'hello' in app.post('/')
    assert 'unknown' in app.delete('/')
```

The failure is:

```
tests/test_abc.py:40: in test_multiple_abcs
    assert 'hello' in app.post('/')
E   assert 'hello' in <200 OK text/plain body=b'unknown method'>
E    +  where <200 OK text/plain body=b'unknown method'> = <bound method TestApp.post of <webtest.app.TestApp object at 0x111c6e898>>('/')
E    +    where <bound method TestApp.post of <webtest.app.TestApp object at 0x111c6e898>> = <webtest.app.TestApp object at 0x111c6e898>.post
```

If you remove the `OtherBase` base class from `Resource` the test does not fail.

wichert:
The use of ABCs turns out to be a red herring: I get the exact same error if I use normal base classes.

wichert:
Simplest possible test case:

```python
from webtest import TestApp
from pyramid.config import Configurator

class OtherBase(object): pass
[...]

PR #1553, opened 2015-02-04
Rework route to use views with most predicates first.

sontek:
I haven't written the test yet, but this solves #1552

### DIFF
--- a/pyramid/router.py
+++ b/pyramid/router.py
@@ -165,9 +165,8 @@ class Router(object):
                 # look for other views that meet the predicate
                 # criteria
                 pred_matches = []
-                missed_count = -1
 
-                for iface in context_iface.__sro__[1:]:
+                for index, iface in enumerate(context_iface.__sro__[1:]):
                     previous_view_callable = view_callable
                     view_callable = adapters.lookup(
                         (IViewClassifier, request.request_iface, iface),
@@ -184,18 +183,17 @@ class Router(object):
                             if pred(context, request):
                                 count += 1
 
-                        pred_matches.append((view_callable, count))
+                        pred_matches.append((view_callable, count, index))
                     else:
-                        pred_matches.append((view_callable, missed_count))
-                        missed_count -= 1
+                        pred_matches.append((view_callable, 0, index))
 
                 sorted_matches = sorted(
                     pred_matches,
-                    key=lambda x: x[1],
+                    key=lambda x: (x[1], x[2]),
                     reverse=True,
                 )
 
-                for view_callable, count in sorted_matches:
+                for view_callable, count, index in sorted_matches:
                     if view_callable is not None:
                         try:
                             response = view_callable(context, request)
--- a/pyramid/router.py
+++ b/pyramid/router.py
@@ -171,8 +171,10 @@ class Router(object):
                     view_callable = adapters.lookup(
                         (IViewClassifier, request.request_iface, iface),
                         IView, name=view_name, default=None)
+
                     # intermediate bases may lookup same view_callable
-                    if view_callable is previous_view_callable:
+                    if view_callable is previous_view_callable or \
+                       view_callable is None:
                         continue
 
                     # if we have predicates, lets see how many match and try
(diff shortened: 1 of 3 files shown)


sontek:
@mmerickel @tseaver What do you guys think about this?

sontek:
@wichert This should fix your problem from #1552

wichert:
I can confirm that this change fixes #1552.

pyramid/router.py
```diff
@@ -164,14 +164,38 @@ def handle_request(self, request):
...
+                       view_callable is None:
                         continue
+
+                    # if we have predicates, lets see how many match and try
+                    # most specific first.
+                    if hasattr(view_callable, '__predicates__'):
+                        count = 0
+                        for pred in view_callable.__predicates__:
+                            if pred(context, request):
+                                count += 1
```
tseaver: Why would we include `view_callable` at all if `pred(context, request)` returns False?

wichert: You can probably grab all view_callables, sort them by number of predicates and then look for the first one that matches?

mmerickel: The number of predicates is sort of irrelevant - each predicate has a weight and then views are sorted and tested based on the sum total of all predicate weights on the view. This is per-context, however, which is causing the problem. There have been murmurings in the past about making context a true predicate instead of the thing it is now, which would solve this problem indirectly.

sontek: @tseaver I just kept it in because the code was already handling predicatemismatch down below but we can totally remove it, I'll push a separate commit

sontek:
@tseaver See the commit here for preferring the best match instead of looping over and catching predicatemismatch:

https://github.com/sontek/pyramid/commit/d05aff124594415621f7f46d6c115bb6db8751ee

mmerickel:
I don't have a good idea yet but this feels wrong. Can someone help me list out our options? Or even whether fallback should even be different from normal view lookup? Why not just use this idea for view lookup in all cases?

- This modifies fallback lookup to completely ignore the hierarchy of specificity and only use the predicates to make the decision.
- This fallback means we always can use a specific context to override this predicate-length lookup. However for that to work the context must be exact, not just more-specific in the hierarchy.
- The way I mentally approached the original PR adding the fallback lookup was as inheritance levels. We were only thinking about single-inheritance so it was obvious, but I think the idea is the same here. Can we preserve that? For example, when we have A <- (B <- D, C) can we try the most-specific A, then level 2 siblings B+C then D? Python below if this syntax doesn't work.

```python
class D: pass
class B(D): pass
class C: pass
class A(B, C): pass
```

mmerickel:
Also, we are focused on classes here, but how does this change if we are looking at an interface hierarchy?

bertjwregeer:
Talking this through with @mmerickel on irc, as well as @sontek.

One of the issues I see is that using the most specific (by having the most predicates) would mean this would break:

    class A: pass
    class B(A): pass
    class C(B): pass

Now, on `class A` we have a function that has a lot of predicates (for example `submit_test` with 10 predicates). Now on `class B` we override this with a predicate list (for example `submit_test_http_var_matches` with 5 predicates). This new `submit_test_http_var_matches` has a more specific predicate on some HTTP header variable, which would still pass on `A.submit_test()` but `B.submit_test_http_var_matches()` is technically the more specific match, and is just one up in the hierarchy from  `class C`, and thus I would expect would win the view. However with the new fallback code, this would instead pick `A.submit_test()` since it has more predicates. If I understand the new code correctly.

mmerickel:
I've implemented a different solution that preserves more of what we expect from view lookup. This PR throws away any ordering based on context-specificity which I really feel like people won't actually want.

bertjwregeer:
With the fix @mmerickel provided in #1557 I am a -1 on this PR.

mmerickel:
As far as #1557 goes I'll not be merging that until I get a blessing from @tseaver or @mcdonc as it's hitting a hot route in pyramid's router (re: performance) as well as being a somewhat sensitive topic on how the lookup occurs.

sontek:
Closing this in favor of #1557

PR #1557, opened 2015-02-05
provide a fix for view lookup fallbacks

mmerickel:
Alternate fix for #1552 versus #1553.

This PR removes the usage of adapters.lookup in favor of an explicit search through the SRO for matching views. It is likely less efficient right now because it is circumventing the optimized `adapters.lookup` call in `zope.interface`. The problem with the original fallback code is that we would check each interface/class in a context's SRO against all views, regardless of their original iface.

In the example below (note the annotations on the asserts), the context is a `Resource` object, containing an SRO of `[Resource, OtherBase, Int1, Int2, object, Interface]`. Note that we actually have multiple views registered for this context 
: `hello` on `Int1`, `hello` on `Int2` and `unknown` on `Interface`. However, the `adapters.lookup` call in the router is only going to return the first one it finds. In this case (based on the SRO) `Int1`'s `hello`. This fails due to the predicate, so now we enter fallback logic. Fallback logic currently uses `OtherBase` in another `lookup` call and finds the `unknown` view registered for `Interface`.

`OtherBase` is checked for views when we really don't want it to have any in this instance. However `Interface` is in `OtherBase`'s hierarchy, and thus the `unknown` view is found. We wouldn't expect this view to be found until the end of the SRO when we check `Interface`.

This PR solves that behavior by checking explicitly all views in the SRO one by one. The new implementation of `find_views` is what I'd expect to see if `lookup` was able to return all matching views, but I'm not aware of an API in `zope.interface` to return this.

```python
from webtest import TestApp

class OtherBase(object): pass
class Int1(object): pass
class Int2(object): pass

class Resource(OtherBase, Int1, Int2):
    def __init__(self, request):
        pass

def unknown(context, request):
    return 'unknown'

def view(context, request):
    return 'hello'

def test_views_for_multiple_base_classes():
    config = Configurator()
    config.add_route('root', '/', factory=Resource)
    config.add_view(unknown, route_name='root', renderer='string')
    config.add_view(view, route_name='root', renderer='string',
            context=Int1, request_method='GET')
    config.add_view(view, route_name='root', renderer='string',
            context=Int2, request_method='POST')
    app = TestApp(config.make_wsgi_app())
    assert 'hello' in app.get('/') # --> Pass
    assert 'hello' in app.post('/') # --> Fail due to bug mentioned above
    assert 'unknown' in app.delete('/') # --> Pass
```


bertjwregeer:
:+1: This looks good.

wichert:
@mcdonc @tseaver Can you review this?

mcdonc:
Frozen in indecision on this.  I think I need to set up some sort of real-world benchmark to see how much this slows down real-world view execution.

davisagli:
FWIW, it looks like the adapter registry's `subscriptions` method is almost what we want, but not quite. It returns all matching components instead of just the first one like `lookup`. But, views would have to be registered in a different way (using `subscribe` instead of `register`) and the API doesn't handle registering and looking up subscribers with names even though the internal data structure supports it. :(

mmerickel:
`subscriptions` is inherently unordered as well, isn't it?

davisagli:
http://docs.zope.org/zope.interface/adapter.html says "Subscribers defined for less general required interfaces are returned before subscribers for more general interfaces."

mmerickel:
We need to do some benchmarking but just intuitively I think there may be a benefit from merging `IView`, `IMultiView` and `ISecuredView` into a common class for lookup. We would keep the original 3 for configuration purposes but since they are mutually exclusive we could re-register the final views for some `ILookupView` iface so that we only needed 1 lookup instead of 3 per view. This is the inner loop so it should have a decent benefit, but again without profiling I'm not sure where the bottleneck actually is.

davisagli:
The registry lookup APIs memoize their result until the registry changes and the cache lookup is implemented in C, so it seems very likely that we'll lose compared to that by implementing our own lookup loop in Python. But yeah, we need benchmarking, not hypotheses.

mmerickel:
Interesting observation about subscribers. As I said in my previous comment we should be able to add a lookup-optimized version of the view hierarchy separately from the config-optimized version. If we did this then we could use whatever mechanism is useful including possibly subscribers. We'd need a way to differentiate view subscribers from someone who just registered a subscriber for `Interface` though. Would that require a separate registry? At the end of the day we care about performing lookup on the following data `(request_iface, context_iface, name)`.

mcdonc:
Thanks @davisagli for comparing the lookup we want to subscriber lookup.  I may end up adding an API to zope.interface as a result.

mcdonc:
Initial trolling (adding a pdb to router.py right before ``_find_views``) makes me think we need a deeper fix, because execution of retrieving ``/`` via HTTP (which is a "hot route") causes the execution of "adapters.registered" to be called at least four times.  While I don't know how much actual clock time this adds to response generation, I suspect we can do better, possibly by adding an API in z.i.  Here's pdb output for fun:


```
> /home/chrism/projects/pyramid/pyramid/router.py(143)handle_request()
-> views_iter = _find_views(
(Pdb) s
> /home/chrism/projects/pyramid/pyramid/router.py(144)handle_request()
-> registry,
(Pdb) s
> /home/chrism/projects/pyramid/pyramid/router.py(145)handle_request()
-> request.request_iface,
(Pdb) s
> /home/chrism/projects/pyramid/pyramid/router.py(146)handle_request()
-> context_iface,
(Pdb) s
> /home/chrism/projects/pyramid/pyramid/router.py(147)handle_request()
-> view_name,
(Pdb) s
> /home/chrism/projects/pyramid/pyramid/router.py(150)handle_request()
-> view_callable = next(views_iter, None)
(Pdb) s
--Call--
> /home/chrism/projects/pyramid/pyramid/view.py(420)_find_views()
-> def _find_views(registry, request_iface, context_iface, view_name):
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(421)_find_views()
-> registered = registry.adapters.registered
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(422)_find_views()
-> view_types = (IView, ISecuredView, IMultiView)
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(423)_find_views()
-> for req_type, ctx_type in itertools.product(
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(424)_find_views()
-> request_iface.__sro__, context_iface.__sro__
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(426)_find_views()
-> source_ifaces = (IViewClassifier, req_type, ctx_type)
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(427)_find_views()
-> for view_type in view_types:
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(428)_find_views()
-> view_callable = registered(
(Pdb) l
423  	    for req_type, ctx_type in itertools.product(
424  	        request_iface.__sro__, context_iface.__sro__
425  	    ):
426  	        source_ifaces = (IViewClassifier, req_type, ctx_type)
427  	        for view_type in view_types:
428  ->	            view_callable = registered(
429  	                source_ifaces,
430  	                view_type,
431  	                name=view_name,
432  	            )
433  	            if view_callable is not None:
(Pdb) p req_type
<InterfaceClass pyramid.request.home_IRequest>
(Pdb) p ctx_type
<implementedBy pyramid.traversal.DefaultRootFactory>
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(429)_find_views()
-> source_ifaces,
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(430)_find_views()
-> view_type,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(431)_find_views()
-> name=view_name,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(433)_find_views()
-> if view_callable is not None:
(Pdb) p view_callable
None
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(427)_find_views()
-> for view_type in view_types:
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(428)_find_views()
-> view_callable = registered(
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(429)_find_views()
-> source_ifaces,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(430)_find_views()
-> view_type,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(431)_find_views()
-> name=view_name,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(433)_find_views()
-> if view_callable is not None:
(Pdb) p view_callable
None
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(427)_find_views()
-> for view_type in view_types:
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(428)_find_views()
-> view_callable = registered(
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(429)_find_views()
-> source_ifaces,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(430)_find_views()
-> view_type,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(431)_find_views()
-> name=view_name,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(433)_find_views()
-> if view_callable is not None:
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(427)_find_views()
-> for view_type in view_types:
(Pdb) p view_callable
None
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(424)_find_views()
-> request_iface.__sro__, context_iface.__sro__
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(426)_find_views()
-> source_ifaces = (IViewClassifier, req_type, ctx_type)
(Pdb) p req_type
<InterfaceClass pyramid.request.home_IRequest>
(Pdb) p ctx_type
<InterfaceClass zope.interface.Interface>
(Pdb) n
> /home/chrism/projects/pyramid/pyramid/view.py(427)_find_views()
-> for view_type in view_types:
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(428)_find_views()
-> view_callable = registered(
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(429)_find_views()
-> source_ifaces,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(430)_find_views()
-> view_type,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(431)_find_views()
-> name=view_name,
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(433)_find_views()
-> if view_callable is not None:
(Pdb) 
> /home/chrism/projects/pyramid/pyramid/view.py(434)_find_views()
-> yield view_callable
(Pdb) p view_callable
<function my_view at 0x2845500>
```

mcdonc:
I suspect we also need to change the exception view handling tween to use _find_views (or whatever equivalent we come up with):

https://github.com/Pylons/pyramid/blob/fix.view-lookup/pyramid/tweens.py#L43

mmerickel:
Yes, I totally forgot to update the exception view tween.

davisagli:
I made a derivative of this branch which caches the view lookup: https://github.com/Pylons/pyramid/compare/fix.view-lookup...davisagli:fix.view-lookup

I've been using this test to benchmark it:

```
import time
import unittest

def bench(n):
    def make_benchmark(func):
        def benchmark(*args, **kw):
            t = time.time()
            i = 0
            while i < n:
                func(*args, **kw)
                i += 1
            elapsed = time.time() - t
            print('Ran {} times; {} microseconds per call.'.format(n, elapsed / n * 1000000))
        return benchmark
    return make_benchmark

class TestViewBenchmarks(unittest.TestCase):

    def _makeRequest(self):
        from pyramid.request import Request
        environ = {
            'wsgi.url_scheme':'http',
            'wsgi.version':(1,0),
            'SERVER_NAME':'example.com',
            'SERVER_PORT':'6543',
            'PATH_INFO':'/',
            'SCRIPT_NAME':'',
            'REQUEST_METHOD':'GET',
            }
        return Request(environ=environ)

    def test_dispatch(self):

        config = Configurator()
        def view(context, request):
            return {}
        config.add_view(view, renderer='json')
        app = config.make_wsgi_app()
        request = self._makeRequest()
        request.registry = app.registry

        @bench(1000)
        def dispatch():
            app.handle_request(request)

        dispatch()
```

Results on master: Ran 1000 times; 80.8877944946289 microseconds per call.
Results with this PR: Ran 1000 times; 117.54012107849121 microseconds per call.
Results with the cache: Ran 1000 times; 78.55081558227539 microseconds per call.

mcdonc:
pviews and proutes also use adapters.lookup and will need fixing

bertjwregeer:
Isn't this information already stored somewhere in zope.interface? Does it not make more sense to make a fix there instead?

mcdonc:
@bertjwregeer i tried to make use of z.i's various internal caches, but to do so, I would have had to change ``z.i.adapters.AdapterRegistry``  to  send some sort of event when the adapter registry was changed, and z.i doesn't actually have an event system, so it was a deeper sort of change than I wanted. 

What I ended up doing was taking  the advice of @davisagli  and just caching the result of finding all potential views within ``pyramid.view._find_views``.  The cache is invalidated when any call to ``pyramid.config.Configurator.add_view`` is called.  I think this is relatively sane.  I did not change how view lookup itself was done,  so any add-ons which sort of reach under the hood  and add a view without help from add_view (and there are some ,  I wrote at least one) will continue to  work, but they will fail to invalidate the cache.  That's mostly OK because the cache is only populated when requests start to enter the system,  and, at least for the add-ons I wrote, they don't add views after that starts to happen.

I fixedthe exception view tween, all the scripts, and anywhere else I saw ``config.registry.adapters.lookup`` used.

I  think this is ready for merge, and if I hear no complaints today I'll do it tomorrow.

bertjwregeer:
That makes sense, I was just thinking it may have been a small change to z.i's lookup functionality to return the full list end to end, rather than a single entry taking into account the hierarchy, but if that is not the case then this solution is definitely the best way forward.

mmerickel:
This looks excellent @mcdonc. +1

pyramid/view.py
```diff
@@ -414,3 +417,93 @@ def callback(context, name, ob):
...
+                    '__call_permissive__',
+                    view_callable
+                    )
+
+            # if this view is secured, it will raise a Forbidden
+            # appropriately if the executing user does not have the proper
+            # permission
+            response = view_callable(context, request)
+            return response
+        except PredicateMismatch as _pme:
```
mmerickel: consider some way of returning all predicate mismatches, but as a separate issue from this pr

bertjwregeer:
\o/

PR #1559, opened 2015-02-06
scaffolds: Add sphinx config

msabramo:
Advantage: It's good to encourage people to document their projects.
Disadvantage: They could just run `sphinx-quickstart`.
Advantage: They may not **know** to run `sphinx-quickstart` so maybe
it's good to put the files there to make them curious.

In the future we may want to add https://pypi.python.org/pypi/pyramid_autodoc from @sontek to the default Sphinx configs...


blaflamme:
I don't know if we should provide this in the scaffold, but if we add this we should make sphinx a dependency to make it work, or at least provide a docs dependency as extra_require.

pyramid/scaffolds/alchemy/docs/conf.py
```diff
@@ -0,0 +1,271 @@
...
+
+# The encoding of source files.
+#source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'{{project}}'
+copyright = u'2015, {{author}}'
```
stevepiercy: Could you change this to use a variable instead of a fixed year?

pyramid/scaffolds/starter/docs/conf.py
```diff
@@ -0,0 +1,271 @@
...
+
+# The encoding of source files.
+#source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'{{project}}'
+copyright = u'2015, {{author}}'
```
stevepiercy: Variable > fixed date.

pyramid/scaffolds/zodb/docs/conf.py
```diff
@@ -0,0 +1,271 @@
...
+
+# The encoding of source files.
+#source_encoding = 'utf-8-sig'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'{{project}}'
+copyright = u'2015, {{author}}'
```
stevepiercy: One more. Variable > fixed year.

stevepiercy:
I like this idea. It practices what we preach: "If it ain't documented, it's broken."

mmerickel:
I'm against adding this extra stuff to the scaffolds, but I'm willing to be in the minority. Just wanted to voice my opinion.

bertjwregeer:
I am a -1 on this. While I agree that documentation is a big part, we shouldn't be telling people what tools to use even if Sphinx is the only one I can think of off the top of my head.

stevepiercy:
I thought it over some more. When I lead tutorials for those who are just trying out Pyramid through a scaffold, no one has ever asked how to document their project even when I prompt that documentation is very important. Also as an extension to @bertjwregeer's comment, the inclusion of a tool implies we have an opinion to use it and support it.

I'll fall back to an undecided or 0 on this idea as it is. Perhaps if it were optional...?

mmerickel:
I think this doesn't belong in the scaffold, we're trying to keep this thing simple for people to get started with. Pyramid has some support for using multiple scaffolds in a project though so this may have a place externally.

PR #1560, opened 2015-02-06
Add tox.ini files to scaffolds

msabramo:
Cc: @sontek, @mmerickel, @goodwillcoding, @wichert

### DIFF
--- a/pyramid/scaffolds/alchemy/test-requirements.txt_tmpl
+++ b/pyramid/scaffolds/alchemy/test-requirements.txt_tmpl
@@ -1,0 +1,3 @@
+mock
+pytest
+pytest-cov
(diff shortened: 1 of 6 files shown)


mmerickel:
I don't think anyone will want to test their webapp on 6 versions of python. I'm -1 on this myself. This is also using pip which we don't advocate yet across the pylons projects.

bertjwregeer:
I'm a -1 on this. People have their own way of doing testing. While I think tox is a fantastic tool, adding it to the scaffold by default will only add one more thing we will need to explain to people as to why it exists.

sontek:
We are already in a pretty bad habit of recommending tools, we just seem to choose the least popular option when we do it. All 3 of our scaffolds ship with chameleon by default ;)

bertjwregeer:
@sontek Touché.

msabramo:
Ok that's 2 -1s and a good point about it being one more thing to explain. I'll close.

wichert:
-1

@mmerickel is absolutely right here: it doesn't make sense to test a web application with multiple Python versions, and I doubt anyone does that. It is very useful with for frameworks and libraries (and even there you can do the same thing very well without using to - personally I've never used it for example), but that is not what our scaffolds create.

In addition: while I personally love py.test, I don't think we want to make that a standard dependency.

PR #1565, opened 2015-02-07
Fix: default content type issues in renderers

### LINKED ISSUE #1344: It should be possible to set Content-Type to text/html when using renderers
Currently, this code will be served to the browser as text/plain thus the HTML are not rendered by the browser:

```

def hello_world(request):
    request.response.content_type = "text/html"
    return "<p>Hello World</p>"

config = Configurator()
config.add_route('hello', '/')
config.add_view(hello_world, route_name='hello', renderer='string')
app = config.make_wsgi_app()
make_server('', 8000, app).serve_forever()
```

A little bit of investigative work shows that the issue is here:

```
File: /pyramid/renderers.py
155 def string_renderer_factory(info):
156     def _render(value, system):
157         if not isinstance(value, string_types):
158             value = str(value)
159         request = system.get('request')
160         if request is not None:
161             response = request.response
162             ct = response.content_type
163             if ct == response.default_content_type:
164                 response.content_type = 'text/plain'
165         return value
166     return _render
```

Here response.default_content_type == 'text/html' and the string renderer replaces the specified content_type with its default of text/plain. I think this is unintuitive/unexpected behavior, instead when request.response.content_type is explicitly set to 'text/html', the renderer should be not change it. 

I didn't test the json renderer, but since there is a similar code over there, I'd assume it has the same bug as well.

bertjwregeer:
Alright, I wanted to provide this as a possible way to fix #1344.

This uses the fact that `content_type` is a property in WebOb's `Response` and it is not used when setting the `content_type` to `default_content_type` in `Response.__init__()`, since that just shoves it directly into the headerlist. So this will break if anyone changes WebOb, but WebOb tends to move slower than molasses, this fix should hold for at least a couple of years!

Although I do think that this is a very dirty hack, it does currently fix the issue. I really hope there is a different fix that is much cleaner!

### DIFF
--- a/pyramid/renderers.py
+++ b/pyramid/renderers.py
@@ -162,7 +162,7 @@ def string_renderer_factory(info):
         if request is not None:
             response = request.response
             ct = response.content_type
-            if ct == response.default_content_type:
+            if response.implicit_content_type:
                 response.content_type = 'text/plain'
         return value
     return _render
@@ -259,8 +259,7 @@ class JSON(object):
             request = system.get('request')
             if request is not None:
                 response = request.response
-                ct = response.content_type
-                if ct == response.default_content_type:
+                if response.implicit_content_type:
                     response.content_type = 'application/json'
             default = self._make_default(request)
             return self.serializer(value, default=default, **self.kw)
@@ -366,7 +365,7 @@ class JSONP(JSON):
                     ct = 'application/javascript'
                     body = '%s(%s);' % (callback, val)
                 response = request.response
-                if response.content_type == response.default_content_type:
+                if response.implicit_content_type:
                     response.content_type = ct
             return body
         return _render
--- a/pyramid/response.py
+++ b/pyramid/response.py
@@ -27,7 +27,26 @@ _BLOCK_SIZE = 4096 * 64 # 256K
 
 @implementer(IResponse)
 class Response(_Response):
-    pass
+    # This hack will break if WebOb's __init__ ever starts setting the
+    # content_type using the getter/setter's rather than just shoving it
+    # directly into the header list.
+    #
+    # This hack is required because we have no other way to know if the current
+    # content-type has been set by the developer, or if it is the default of
+    # text/html. We need to know this, so that if it is the default we can
+    # mutate it to the correct one in our renderers, unless it has been
+    # modified. See https://github.com/Pylons/pyramid/issues/1344
+    implicit_content_type = True
+
+    def _content_type__set(self, value):
+        super(Response, self)._content_type__set(value)
+        self.implicit_content_type = False
+
+    content_type = property(_Response._content_type__get, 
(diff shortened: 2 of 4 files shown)


mmerickel:
Yeah I considered something like this. There's at least a few ways to set the content type that are not accounted for here.

- `Response(content_type=...)`
- `Response(headerlist=[('Content-Type', ...)])`
- `response.headerlist.append(('Content-Type', ...))`
- `response.headers['Content-Type'] = ...`
- `response.environ['HTTP_CONTENT_TYPE'] = ...`

With this in mind I'm just not sure if an approach that attempts to detect a write will work. It'd need to be a check within a read.

bertjwregeer:
That's exactly what I was afraid of. However just so you are aware, `content_type` only pulls data out of `headerlist`, so even if you were to set it in the environ, it wouldn't really do anything (other than probably downstream on the WSGI server).

Fixing the first is pretty simple, by overriding `__init__` for Response within Pyramid. 

The others are more tricky :-(

mmerickel:
I didn't realize it wasn't using the environ immediately, thanks for pointing that out.

bertjwregeer:
The downside with a read needing to know is that somehow `Response()` has to record somewhere that it is no longer `default_content_type`. I have a bad feeling we are trying to shoe-horn a solution that won't work well because WebOb immediately takes whatever you set `content_type` to and shoves it into the `headerlist`. At that point it is a bare `str` and we can't really do anything with it.

The solution I've provided above will work for 95% of the users of Pyramid, those that are going out of their way to modify the headers directly are doing so with explicit knowledge that they are doing it. At some point we have to take the training wheels off...

bertjwregeer:
I have an idea... what if instead of setting the default content type up-front in the Response, we keep it empty until we are ready to send the headers to the WSGI servers in the response.

Now we can easily see if it has been set or not, and then do the appropriate thing. According to the RFC for HTTP/1.1 a server __SHOULD__ (I really wish that was a MUST) send a `Content-Type`, so the use case of the user removing all content types while technically legal should be avoided anyway.

----

I'd also be a strong proponent to setting the default content-type to application/octet-stream, and then making sure all of our renderers correctly set the content-type on the response. According to the RFC for HTTP/1.1 clients are supposed to default to application/octet-stream if the server doesn't send an appropriate content-type response.

mmerickel:
I believe it's "should" because certain responses like "204 no content" don't have a content-type/body.

bertjwregeer:
That's an HTTP reply without an entity-body, that one SHOULD not include a content-type since there is no content.

If there is an entity-body, right now a reply SHOULD contain the content-type, when it could have been a MUST.

mmerickel:
We've decided that an appropriate solution cannot happen until we make some changes to webob. Tracking that issue at https://github.com/Pylons/webob/issues/204.

Issue #1537, opened 2015-01-21
response factory's request argument should be documented as being allowed to be None


mmerickel:
The response factory api was introduced and is unreleased in #1499.

It should be documented that the `request` may be `None` to support usage with the rendering subsystem that works without a request object. This would allow `render_to_response` from #1536 to use the response factory to create a response object to return.

PR #1567, opened 2015-02-07
Document the factory requires a positional argument

bertjwregeer:
Start on https://github.com/Pylons/pyramid/issues/1537

Since it was never documented other than in the glossary that it takes a single positional argument, we don't need to add much to document that it may be None.


mmerickel:
Ok, I'll take it from here, thank you. I noticed `IResponseFactory` wasn't in the public api yet and had a really weird docstring.

PR #1574, opened 2015-02-10
Prevent timing attacks when checking CSRF token

dstufft:


mmerickel:
I think you need to check if the token is `None` prior to passing it into `strings_differ` right now.

dstufft:
Ah right, that's what I get for using the web editor.

dstufft:
There we go, fixed that.

mmerickel:
Thank you!

PR #1575, opened 2015-02-11
clone a new RendererHelper per request

### LINKED ISSUE #1268: Caching of RendererHelper in view_config()
*incomplete*: Wanted to get the thoughts and findings that @mmerickel were going through down in the bug tracker.

As of right now `pyramid_mako` has broken the `pyramid.reload_templates` setting, digging into it a little deeper the issue can be traced back to the renderer factories return being cached in `RendererHelper`: https://github.com/Pylons/pyramid/blob/master/pyramid/renderers.py#L400

When used with `render()`, this isn't really an issue since a new `RenderHelper` is created for each call to it.

However using `view_config()` the `RendererHelper` gets cached, so after the first request the return from the renderer factory is reified and the factory is never called again, and the `RendererHelper` persists.

---

To support `reload_templates` the returned function that deals with rendering is actually doing all of the work of verifying the template on disk has changed or not, this basically moves any logic for asset resolving out of a one time thing in the factory back into something that has to be done every time the renderer is called and there is no good way to cache the information so that asset resolving doesn't have to be done each and every call to render.

---

Is there a better way to solve this, to support `reload_templates` without having to go through the trouble the renderers currently go through? Is there a way to save the result of an asset lookup, or an effective way to cache it?

mmerickel:
fixes #1268

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -128,6 +128,13 @@ Bug Fixes
 - Prevent "parameters to load are deprecated" ``DeprecationWarning``
   from setuptools>=11.3. See https://github.com/Pylons/pyramid/pull/1541
 
+- Avoiding sharing the ``IRenderer`` objects across threads when attached to
+  a view using the `renderer=` argument. These renderers were instantiated
+  at time of first render and shared between requests, causing potentially
+  subtle effects like `pyramid.reload_templates = true` failing to work
+  in `pyramid_mako`. See https://github.com/Pylons/pyramid/pull/1575
+  and https://github.com/Pylons/pyramid/issues/1268
+
 Deprecations
 ------------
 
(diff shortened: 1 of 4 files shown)


bertjwregeer:
LGTM.

mmerickel:
@mcdonc needs to sign-off since he said "don't remove the reify" without providing explanation in #1270

mcdonc:
LGTM, no real speed concern, and I agree that caching the view renderer between calls is an optimization without much purpose.  That said, I think we should add a clone() method to the IRendererInfo interface to make it clear that it is non-optional (even though this is pretty inside baseball).

mmerickel:
Yeah if we're worried about people using it we could easily pass a lambda/factory through the machinery but since clone was already there with a comment about its use in views, and wasn't actually called anywhere, it seemed like a good thing to use. I will add it to the interface though.

PR #1576, opened 2015-02-11
Don't create sdist with tox (latest setuptools doesn't like mixing)

sontek:


tseaver:
I really don't like this change:  one main advantage of using `tox` has been that it ensured our sdists stayed installble.  What is it real motivator?  When I revert the change locally, `tox --recreate` passes everyting cleanly (w/ setuptools 12.2).

sontek:
@tseaver Yeah, this was necessary at the time because the latest setuptools broke everything if you ran tox twice.  I believe they have fixed this now but I need to test.

Issue #689, opened 2012-09-13
Terminal input issues after restart during PDB session


Sapphire64:
Hello again :) I have issues on two of my systems - both amd64 Debian Wheezy with KDE 4.8.4. Both reproducible in python 2.7 and in python 3.2.

Description:
When I'm adding `import pdb; pdb.set_trace()` into my Pyramid's project code I have possibility to debug it with interactive interpreter. This is very cool, but 2nd or 3rd restart (by ^C or automatically) breaks this cool feature - I can't see nothing what I'm typing in terminal - just blank input line:
"(Pdb) {always blank here} ". 
Any input actually works (but somewhat weird), but it does not echo this input in terminal. So I always have to close this terminal window, open new and start `../bin/pserve development.ini` again. 

To describe more this issue: when I'm typing `su` in this broken terminal it gives me only a second to type password and then it automatically press 'enter' so system responds with "Wrong password".

This issue easily reproducible on both of my systems, in py2 and py3 pyramid projects, in Konsole, Yakuake and XTerm terminals.

I'm not sure what can it be, maybe not exactly Pyramid bug, but maybe bug of Pyramid's webserver or pdb module itself. Looks like this is not an issue of terminals - even if Konsole & Yakuake are almost the same inside, oldschool XTerm quite differs from them. Maybe I should try to reproduce this in system's tty terminal.

PR #1577, opened 2015-02-13
Add workaround to make sure echo is enabled after reload

davisagli:
This is a workaround for #689, by making sure the terminal's echo is enabled after reloading on systems that support that.


mmerickel:
awesome thank you

PR #1578, opened 2015-02-13
Fix a couple minor documentation issues

davisagli:
Do these require changelog entries?


stevepiercy:
Is it related to this change log entry?
https://github.com/Pylons/pyramid/blob/master/CHANGES.txt#L12-L13

davisagli:
Nope

stevepiercy:
OK. This LGTM, but I would like one of the more senior contributors to review and merge. Ping @mmerickel

PR #1581, opened 2015-02-17
add InstancePropertyHelper and apply_request_extensions

mmerickel:


bertjwregeer:
LGTM

mmerickel:
That little edit to IRequestFactory was because IRequestFactory is a public api and IRequest is apparently not. So I didn't want the private api referenced from the public docstring.

bertjwregeer:
I had not realised that `IRequest` is not a public API. Probably because at this point it is excessively large ;-)

mmerickel:
Well I actually thought it was. But it's private and just a marker interface actually.

pyramid/request.py
```diff
@@ -307,3 +312,22 @@ def call_app_with_subpath_as_path_info(request, app):
...
     return new_request.get_response(app)
+
+def apply_request_extensions(request, extensions=None):
+    """Apply request extensions (methods and properties) to an instance of
+    :class:`pyramid.interfaces.IRequest`. This method is dependent on the
+    ``request`` containing a properly initialized registry.
+
+    After invoking this method, the ``request`` should have the methods
+    and properties that were defined using
+    :meth:`pyramid.config.Configurator.add_request_method`.
```
sontek: Should we document how you would use `extensions`?

pyramid/router.py
```diff
@@ -213,7 +214,7 @@ def invoke_subrequest(self, request, use_tweens=False):
             try:
                 extensions = self.request_extensions
                 if extensions is not None:
-                    request._set_extensions(extensions)
+                    apply_request_extensions(request, extensions=extensions)
```
sontek: `self.request_extensions` is just the interface that `apply_request_extensions` queries anyways, do we need to pass them in?

mmerickel: This is just an optimization to have the extensions queried from the registry only once at app-setup-time. This code is in pyramid's hot route, so I wanted to keep it as close to the original version as possible.

pyramid/request.py
mmerickel: I explicitly left `extensions` off of the public api documentation. It's there just for pyramid's hot route in the router.

mmerickel: We could name it `_extensions` if it would make you feel better.

PR #1583, opened 2015-02-17
remove the token from the ICacheBuster api

### LINKED ISSUE #1582: ICacheBuster interface leaks some implementation details
The ``ICacheBuster`` interface has a method, ``token(pathspec)`` which is used to generate a token for the given ``pathspec``. As far as I can tell the only place this method is used is for calling code to generate a token and then to immediate pass it into ``ICacheBuster.pregenerate``.

I feel like baking this ``token()`` method into the ``ICacheBuster`` interface is leaking the implementation details of the built in cache busting methods into the interface. Conceptually what you need in the cache busting interface is a way to convert and URL back and forth between the non-cachebusting and the cachebusting form. However this API assumes that Pyramid is the tool that is generating the token or the cache buster URLs. This doesn't always hold true though, for instance you might have a static asset pipeline outside of Pyramid that generates a manifest file and then your ``ICacheBuster`` simply needs to look up the path in that manifest file, but since the ``token()`` method is mandatory, you're forced to write a no-op method when you don't need this method.

Suggested on IRC was, I think, a better way of expressing this:

* Drop the ``ICacheBuster.token(pathspec)`` method.
* Modify ``ICacheBuster.pregenerate(token, subpath, kw)`` to ``ICacheBuster.pregenerate(pathspec, subpath, kw)``.
* Modify ``PathSegmentMd5CacheBuster`` so that they can accept a token generating function (e.g. ``PathSegmentMd5CacheBuster(tokenizer=Md5AssetTokenGenerator())`` and they'll simply call that inside of their ``pregenerate`` function.

Since this API hasn't been released yet it may be a good idea to review it before there are backwards compatibility considerations to take into account.

mmerickel:
This exposes the QueryStringCacheBuster and PathSegmentCacheBuster
public APIs alongside the md5-variants. These should be more cleanly
subclassed by people wishing to extend their implementations.

Fixes #1582.

### DIFF
--- a/pyramid/static.py
+++ b/pyramid/static.py
@@ -220,8 +220,7 @@ class PathSegmentMd5CacheBuster(PathSegmentCacheBuster,
     .. versionadded:: 1.6
     """
     def __init__(self):
-        PathSegmentCacheBuster.__init__(self)
-        Md5AssetTokenGenerator.__init__(self)
+        super(PathSegmentMd5CacheBuster, self).__init__()
 
 class QueryStringCacheBuster(object):
     """
@@ -262,8 +261,7 @@ class QueryStringMd5CacheBuster(QueryStringCacheBuster,
     .. versionadded:: 1.6
     """
     def __init__(self, param='x'):
-        QueryStringCacheBuster.__init__(self, param=param)
-        Md5AssetTokenGenerator.__init__(self)
+        super(QueryStringMd5CacheBuster, self).__init__(param=param)
 
 class QueryStringConstantCacheBuster(QueryStringCacheBuster):
     """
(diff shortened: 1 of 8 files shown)


pyramid/static.py
```diff
@@ -192,38 +192,55 @@ def token(self, pathspec):
...
+    """
+    An implementation of :class:`~pyramid.interfaces.ICacheBuster` which
+    inserts an md5 checksum token for cache busting in the path portion of an
+    asset URL.  Generated md5 checksums are cached in order to speed up
+    subsequent calls.
+
+    .. versionadded:: 1.6
+    """
+    def __init__(self):
+        PathSegmentCacheBuster.__init__(self)
```
dstufft: Any reason you're calling these explicitly instead of just doing:

```python
def __init__(self, *args, **kwargs):
    super(PathSegmentMd5CacheBuster, self).__init__(*args, **kwargs)
```

Should be less fragile that way as long as you make sure you pull out any kwargs/args that the class whose defining the ``__init__`` adds.

dstufft:
Added a comment about the particular implementation, but otherwise this looks good to me.

pyramid/static.py
mmerickel: Yeah alright. Fixed.

dstufft:
LGTM

PR #1586, opened 2015-02-19
Revert #1576 "Don't create sdist with tox"

tseaver:
" (latest setuptools doesn't like mixing)"

This reverts commit c534e00ed06ef17506cf5f74553310ec15653834.

I can see no valid reason for that change, and it loses us the assurance that our sdists can be correctly installed.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,4 @@
 [tox]
-skipsdist = True
 envlist = 
     py26,py27,py32,py33,py34,pypy,pypy3,cover
 
--- a/tox.ini
+++ b/tox.ini
@@ -4,14 +4,14 @@ envlist =
 
 [testenv]
 commands = 
-    python setup.py -q dev
+    pip -q install pyramid[testing]
     python setup.py -q test -q
 
 [testenv:cover]
 basepython =
     python2.6
 commands = 
-    python setup.py -q dev
+    pip install pyramid[testing]
     nosetests --with-xunit --with-xcoverage
 deps =
     nosexcover


tseaver:
Undoes #1576.  @mmerickel, @sontek PTAL.

sontek:
@tseaver This was necessary because of a bug in a specific version of setuptools (which happened to be the latest at the time this was merged), they've since fixed that bug

sontek:
@tseaver just tested and the bug still exists int he latest setuptools:

```
Installing collected packages: pyramid
  Found existing installation: pyramid 1.6.dev0
====================
LINK POINTER: /home/sontek/venvs/pyramid/src/pyramid
DIST: /home/sontek/venvs/pyramid/src/pyramid/.tox/py27/lib/python2.7/site-packages
    Exception:
      File "/home/sontek/venvs/pyramid/src/pyramid/.tox/py27/local/lib/python2.7/site-packages/pip/basecommand.py", line 232, in main
        status = self.run(options, args)
      File "/home/sontek/venvs/pyramid/src/pyramid/.tox/py27/local/lib/python2.7/site-packages/pip/commands/install.py", line 347, in run
        root=options.root_path,
      File "/home/sontek/venvs/pyramid/src/pyramid/.tox/py27/local/lib/python2.7/site-packages/pip/req/req_set.py", line 543, in install
        requirement.uninstall(auto_confirm=True)
      File "/home/sontek/venvs/pyramid/src/pyramid/.tox/py27/local/lib/python2.7/site-packages/pip/req/req_install.py", line 575, in uninstall
        '(at %s)' % (link_pointer, self.name, dist.location)
    AssertionError: Egg-link /home/sontek/venvs/pyramid/src/pyramid does not match installed location of pyramid (at /home/sontek/venvs/pyramid/src/pyramid/.tox/py27/lib/python2.7/site-packages)

```

sontek:
@tseaver I tracked the introduction of the bug to this PR:

https://github.com/pypa/pip/commit/c509c5c4aba060a64754c90e1687dd265f58ef77

Which flips the logic on checking if you are in `setup.py develop`  and blows up since the way the pyramid tox is setup it gets `pip installed` via tox and then runs `setup.py develop` inside there.

mmerickel:
We can't just revert this change, it actually fixes an oddity that tox was installing the project twice when running tests.

1) Installed the sdist, which we turned off.
2) Installed again via `python setup.py dev`.

To fix this we need to modify the tox to avoid running `python setup.py dev` and depend on some other way to install the deps.

mmerickel:
Side note, if we do *not* run `setup.py dev` inside of tox then we need to fix the fact that we don't have a MANIFEST because some of our tests will fail due to missing assets. We need to either fix the manifest, or install setuptools-git, or keep doing it the way we are.

tseaver:
I can reproduce the "second run fails" problem, and have a fix coming.

tseaver:
Dropping `setup.py dev` in favor of `pip install pyramid[testing]` didn't trigger any test failures for me, but that could be a freak (Travis is happy, though).

mmerickel:
I'm surprised that worked... Where are the test fixtures coming from if tox is running "sdist"? We don't have "usedevelop = True" so shouldn't they be missing without our manifest?

mmerickel:
After some sleuthing I discovered why this works, and I'm not sure how I feel about it.

It's effectively doing the following.

```
virtualenv env
env/bin/python setup.py sdist --formats=zip
env/bin/python pip install dist/pyramid-1.6.dev0.zip
env/bin/python setup.py test
```

That final `setup.py test` is from the CWD git checkout. So it's really kind of ignoring the sdist isn't it?

mmerickel:
So I did `PYTHONVERBOSE=1 tox -e py26`.

I can confirm `python setup.py test` is executing the tests against the git checkout (CWD), not the sdist.
Second, it appears as if `pip install pyramid[testing]` is pulling down from PyPI instead of doing something like `pip install -e .[testing]` which would pull from the git checkout (CWD).

I think the only way to actually test the sdist itself is to set `skipdist=True` and then make an sdist ourselves, copy it to the folder, unzip it, and run python setup.py test on it, all within the copy's CWD.

mmerickel:
Consider `skipsdist=True` with

```
commands=
    python setup.py sdist
    tar xzvf dist/pyramid-1.6.dev0.tar.gz
    cd pyramid-1.6.dev0
    python setup.py test
```

Anyway @tseaver I'm just trying to build on your original concern "it loses us the assurance that our sdists can be correctly installed". What is your goal here?

mmerickel:
I got something like this to work with

```
[tox]
envlist = 
    py26,py27,py32,py33,py34,pypy,pypy3,cover

[testenv]
commands = 
    {envbindir}/pip install pyramid[testing]
    {envbindir}/nosetests --xunit-file=nosetests-{envname}.xml {envsitepackagesdir}/pyramid
```

in combination with

```
setup_requires = []

# if a git repo then add setuptools-git as a setup_requires
if os.path.isdir(os.path.join(here, '.git')):
    setup_requires += ['setuptools_git']
```

So it runs the tests shipped with the sdist instead of the ones in the cwd.

bertjwregeer:
@mmerickel I remember this being brought up as a possible fix earlier, however it was shot down pretty heavily by everyone involved as a "hack" that we shouldn't do.

mmerickel:
I've incorporated these changes into #1471. Can we get some comments there?

dstufft:
Note, that the other way to actually test what's in the sdist and not use the current directory is to make it so you can't ``import pyramid`` from the current directory. In other words, do ``mkdir src && git mv pyramid src/`` (and adjust the ``setup.py``).

mmerickel:
Closing in favor of #1471.

Issue #1580, opened 2015-02-16
PackageAssetSource package should be a string...


sbrunner:
I'm using pyramid revision 620bdefbff81792fd047f273a234264ce3d0a764 with fa.jquery 0.9.5, and on this version I have this error:
```
Module weberror.evalexception:431 in respond         view
>>  app_iter = self.application(environ, detect_start_response)
Module fanstatic.publisher:219 in __call__         view
>>  return self.app(environ, start_response)
Module fanstatic.injector:54 in __call__         view
>>  response = request.get_response(self.app)
Module webob.request:1320 in send         view
>>  application, catch_exc_info=False)
Module webob.request:1284 in call_application         view
>>  app_iter = application(self.environ, start_response)
Module pyramid.router:242 in __call__         view
>>  response = self.invoke_subrequest(request, use_tweens=True)
Module pyramid.router:217 in invoke_subrequest         view
>>  response = handle_request(request)
Module pyramid_debugtoolbar.toolbar:166 in toolbar_tween         view
>>  return handler(request)
Module pyramid.tweens:21 in excview_tween         view
>>  response = handler(request)
Module pyramid_tm:95 in tm_tween         view
>>  reraise(*exc_info)
Module pyramid_tm:76 in tm_tween         view
>>  response = handler(request):0
Module pyramid.router:163 in handle_request         view
>>  response = view_callable(context, request)
Module pyramid.config.views:326 in attr_view         view
>>  return view(context, request)
Module pyramid.config.views:302 in predicate_wrapper         view
>>  return view(context, request)
Module pyramid.config.views:242 in _secured_view         view
>>  return view(context, request)
Module pyramid.config.views:352 in rendered_view         view
>>  result = view(context, request)
Module pyramid.config.views:488 in _class_view         view
>>  response = getattr(inst, attr)()
Module pyramid_formalchemy.views:119 in models         view
>>  return self.render(models=results)
Module pyramid_formalchemy.views:170 in render         view
>>  main = get_renderer('pyramid_formalchemy:templates/admin/master.pt').implementation(),
Module pyramid.renderers:153 in get_renderer         view
>>  return helper.renderer
Module pyramid.decorator:37 in __get__         view
>>  val = self.wrapped(inst)
Module pyramid.renderers:404 in renderer         view
>>  return factory(self)
Module pyramid_chameleon.zpt:10 in renderer_factory         view
>>  return renderer.template_renderer_factory(info, ZPTTemplateRenderer)
[...]

PR #1585, opened 2015-02-18
send a package name instead of package instance to pkg_resources

mmerickel:
Fixes #1580.

PR #1587, opened 2015-02-19
Fix asset override with package

msabramo:
[`AssetsConfiguratorMixin.override_asset`](https://github.com/Pylons/pyramid/blob/master/pyramid/config/assets.py#L308) does:

```python
__import__(override_package)
to_package = sys.modules[override_package]
override_source = PackageAssetSource(to_package, override_prefix)
```

so it's assuming that the `package` argument to [`PackageAssetSource.__init__`](https://github.com/Pylons/pyramid/blob/master/pyramid/config/assets.py#L215) takes a module object.

But then `PackageAssetSource` had a bunch of methods that did stuff like:

- `pkg_resources.resource_exists(self.package, path)`
- `pkg_resources.resource_filename(self.package, path)`
- `pkg_resources.resource_stream(self.package, path)`

and all these `pkg_resources` functions need their `package_or_requirement`
argument to be a **string**; not a module - see
https://pythonhosted.org/setuptools/pkg_resources.html#basic-resource-access, which says:

> the `package_or_requirement argument` may be either a Python package/module
> name (e.g. `foo.bar`) or a `Requirement` instance.

This causes errors when overriding assets -- e.g.: I am using Kotti and
Kotti has this code (https://github.com/Kotti/Kotti/blob/master/kotti/__init__.py#L251):

```python
    for override in [a.strip()
                     for a in settings['kotti.asset_overrides'].split()
                     if a.strip()]:
        config.override_asset(to_override='kotti', override_with=override)
```

A Kotti add-on called kotti_navigation does this
(https://github.com/Kotti/kotti_navigation/blob/master/kotti_navigation/__init__.py#L12):

```python
settings['kotti.asset_overrides'] += ' kotti_navigation:kotti-overrides/'
```

The above code is all legit as far as I can tell and it works fine in pyramid
1.5.2, but it fails with pyramid master with the following:

```pytb
  File "/Users/marca/python/virtualenvs/kotti_inventorysvc/lib/python2.7/site-packages/pkg_resources.py", line 959, in resource_filename
    self, resource_name
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/assets.py", line 31, in get_resource_filename
    filename = overrides.get_filename(resource_name)
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/assets.py", line 125, in get_filename
    result = source.get_filename(path)
  File "/Users/marca/dev/git-repos/pyramid/pyramid/config/assets.py", line 224, in get_filename
    if pkg_resources.resource_exists(self.package, path):
  File "/Users/marca/python/virtualenvs/kotti_inventorysvc/lib/python2.7/site-packages/pkg_resources.py", line 948, in resource_exists
    return get_provider(package_or_requirement).has_resource(resource_name)
  File "/Users/marca/python/virtualenvs/kotti_inventorysvc/lib/python2.7/site-packages/pkg_resources.py", line 225, in get_provider
    __import__(moduleOrReq)
TypeError: __import__() argument 1 must be string, not module
```

This was a little tricky to resolve because the `override_asset` function wants
to pass a module object to `PackageAssetSource.__init__`, but there are a
number of tests in `pyramid/tests/test_config/test_assets.py` that assume that
it takes a string. So I ended up making it legal to pass either one, so that I
don't have to change as much calling code.

See https://github.com/Kotti/kotti_navigation/issues/13

Cc: @mmerickel, @j23d, @disko

### DIFF
--- a/pyramid/config/assets.py
+++ b/pyramid/config/assets.py
@@ -214,6 +214,10 @@ class PackageAssetSource(object):
     """
     def __init__(self, package, prefix):
         self.package = package
+        if hasattr(package, '__name__'):
+            self.pkg_name = package.__name__
+        else:
+            self.pkg_name = package
         self.prefix = prefix
 
     def get_path(self, resource_name):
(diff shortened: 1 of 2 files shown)


msabramo:
I believe that this fixes https://github.com/Pylons/pyramid/issues/1580, which was pointed out to me by @sontek after I submitted the PR.

msabramo:
Similar to but also different from: https://github.com/Pylons/pyramid/pull/1585

msabramo:
I need to add a test.

msabramo:
OK, I enhanced the tests in 5ace659. With that commit and without e51295b, there are 5 errors with `TypeError: __import__() argument 1 must be string, not module`. With e51295b, all the tests pass.

mmerickel:
I like this fix better. We want to keep importing the `to_package` to detect errors at config time. Thanks for the fix!

sbrunner:
Many thanks @msabramo for your pull request :-)

PR #1589, opened 2015-02-20
Enhance test_assets to expose #1580

msabramo:
This enhances existing tests so that they detect the issue in #1580.

See #1580

Cc: @mmerickel, @sbrunner, @j23d, @disko


mmerickel:
Fixed via #1587.

PR #1593, opened 2015-02-24
1.5 backport: only reset terminal echo flag if stdin is a tty

davisagli:
Backport of https://github.com/Pylons/pyramid/pull/1592


bertjwregeer:
LGTM.

@mmerickel This is causing broken startup for a lot of people.

Issue #1604, opened 2015-03-11
Problem generating HTML documentation (master branch)


ericpar:
Hello,

I've created a fresh virtual environment (```mkvirtualenv```) which uses Python 2.7.
I installed the requirements found in ```rtd.txt```.
I cloned the present repo.
I switched in the ```<pyramid-repo>/docs``` directory and issued ```make html```.

I got this error after a few tens of seconds of processing:

```
...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 132 source files that are out of date
updating environment: 132 added, 0 changed, 0 removed
reading sources... [100%] whatsnew-1.5                                          

Warning, treated as error:
/home/eric/sandbox/pyramid/docs/api/interfaces.rst:92: WARNING: autodoc: failed to import interface u'ICacheBuster' from module u'pyramid.interfaces'; the following exception was raised:
Traceback (most recent call last):
  File "/home/eric/Envs/pyramid/lib/python2.7/site-packages/Sphinx-1.3-py2.7.egg/sphinx/ext/autodoc.py", line 392, in import_object
    obj = self.get_attr(obj, part)
  File "/home/eric/Envs/pyramid/lib/python2.7/site-packages/Sphinx-1.3-py2.7.egg/sphinx/ext/autodoc.py", line 288, in get_attr
    return safe_getattr(obj, name, *defargs)
  File "/home/eric/Envs/pyramid/lib/python2.7/site-packages/Sphinx-1.3-py2.7.egg/sphinx/util/inspect.py", line 115, in safe_getattr
    raise AttributeError(name)
AttributeError: ICacheBuster

make: *** [html] Error 1
```

Is there a missing dependency ?

mcdonc:
I think you need the SPHINXBUILD envvar when running make if you have Pyramid installed in a venv:

```
cd pyramid/docs
make html SPHINXBUILD=/home/eric/Envs/pyramid/bin/sphinx-build
```

mcdonc:
You probably need to do "/home/eric/Envs/pyramid/bin/python setup.py develop" within the pyramid checkout too, if that isnt obvious.

mmerickel:
As a fun alternative, the `tox.ini` now builds the documentation, so you can just run `tox -e py2-docs` if you want.

mcdonc:
tox -e py2-docs is a better idea yup.. then cd to .tox/py2-docs/docs/html for the result

mcdonc:
[...]

PR #1605, opened 2015-03-11
Added tox in requirements for pip. Added entry in README.rst on how to g...

ericpar:
Entry in README.rst on how to generate the documentation locally.
Added missing requirement in ```rtd.txt```.

See the issue #1604.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,6 +20,26 @@ Support and Documentation
 See the `Pylons Project website <http://pylonsproject.org/>`_ to view
 documentation, report bugs, and obtain support.
 
+
+Local documentation
+~~~~~~~~~~~~~~~~~~~
+
+Documentation can be generated locally with this command:
+
+.. code-block:: bash
+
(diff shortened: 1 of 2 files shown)


mcdonc:
Thanks for this.  However:

1.  rtd.txt is a file used for Read The Docs and RTD doesn't use tox to build the docs, so the dependency is not  warranted in that file.

2. The info  about how to  build the  docs is already in HACKING.txt.  It shows a third way, which  AFAICT also works.  The fact that you can  use tox to build the docs  doesn't really  mean it should become the canonical way,  IMO.

So I'm going  to  reject this  PR  for now,  apologies.

ericpar:
Hi,

No worries. I did not look in the hacking file as I thought it was meant
for those who wanted to contribute to the project. I will have a look at
the method described in there.

And I didn't know "rtd.txt" was meant for Read The Docs...

Regards,

Eric

PR #1606, opened 2015-03-13
example was missing trailing parathesis

russellballestrini:


stevepiercy:
Thanks for catching this. I combined the selected ranges into a single range in commit b0218c8 to master.

PR #1607, opened 2015-03-13
fix comment in forbidden_view_config

ledmonster:
The comment was confusing. I think '403 Forbidden' is more appropriate in this situation.


tseaver:
Thank you!

PR #1610, opened 2015-03-15
Allow passing in a custom IResponse implementor for append_slash

### LINKED ISSUE #1609: Allow specifying the type of redirect for append_slash=True
It would be great to allow specifying the type of redirect to be used when using the ``append_slash`` parameter to ``notfound_view_config``. Maybe the API could be ``append_slash=True`` gives you the current behavior, and ``append_slash=HTTPMovedPermanently`` for instance could be used to use ``HTTPMovedPermanently`` to do the redirect?

dstufft:
Fixes #1609


mcdonc:
Ty @dstufft !

PR #1618, opened 2015-03-25
Add more badges

armab:
l6x9xhc 1

Could be more useful for those who checking the project from time to time in GitHub.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,12 +1,21 @@
 Pyramid
 =======
 
+.. image:: https://img.shields.io/pypi/v/pyramid.svg?style=flat
+        :target: https://pypi.python.org/pypi/pyramid
+        :alt: Last release
+
 .. image:: https://travis-ci.org/Pylons/pyramid.png?branch=master
         :target: https://travis-ci.org/Pylons/pyramid
+        :alt: Build status
 
 .. image:: https://readthedocs.org/projects/pyramid/badge/?version=master
         :target: http://docs.pylonsproject.org/projects/pyramid/en/master/
         :alt: Documentation Status
+
+.. image:: https://img.shields.io/pypi/dm/pyramid.svg?style=flat
+        :target: https://pypi.python.org/pypi/pyramid#downloads
+        :alt: Popularity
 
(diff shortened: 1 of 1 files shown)


stevepiercy:
Thanks for this. I need to check with some of the core contributors which versions and branches should have badges, if any. The terminology of latest, master, and other non-numeric release and branch names can be confusing.

stevepiercy:
@armab after a [brief discussion about badges in #pyramid](https://botbot.me/freenode/pyramid/2015-03-26/?msg=35074798&page=1), I'm closing this PR in favor of another to add a badge and link to docs latest. Thanks for bringing up the issue.

armab:
@stevepiercy thanks for fast reaction.

Does it makes sense to fill another PR and add code coverage badge via https://coveralls.io/ ?

Like: 
code coverage for master
-----

All that provides really good brief overview on latest project state not only for those who choosing framework for next project, but also for possible contributors

stevepiercy:
I would say no, because the service requires the owner of the Pylons account to authorize usage, and even though there is no fee to FOSS projects, it still has a bad smell. Avoid entanglements.

armab:
But the owner already granted access for Travis-CI.

How about http://scrutinizer-ci.com/ (it has code coverage feature too) or it's kind of avoiding using 3rd party services when possible?

stevepiercy:
For those who care about code coverage:
http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/introduction.html?highlight=coverage#testing
https://github.com/Pylons/pyramid/blob/master/HACKING.txt#L65-L67
Example: https://travis-ci.org/Pylons/pyramid/jobs/55902353

We do not need yet another service.

armab:
My point is mostly about popularization.

My thinking is: If you care about coverage, why not to show that in easy way. There are very few frameworks with so good coverage. if you support different python versions - why not to show that easily, the same is about popularity or latest release version.
From user's (already framework consumer or possible framework consumer) perspective it has sense.

But I understood your point, thanks for the explanation and your contribution to make `pyramid` awesome.

Issue #1597, opened 2015-03-04
Purpose: Alembic in scaffold Alchemy


marioidival:
Nowadays various frameworks is standard with some kind of database migration tool together. In alchemy scaffold, use SQLAlchemy and we have a migration framework for it when downloaded.

I know that the Pyramid is agnostic, but I believe we should think about migrations that scaffold.

robertlagrant:
+1

marioidival:
https://github.com/Pylons/pyramid/pull/2024

bertjwregeer:
See pull requests #2024 and #1625. I think this would be a great addition to #2024.

mmerickel:
related https://github.com/Pylons/pyramid-cookiecutter-alchemy/pull/7

stevepiercy:
https://github.com/Pylons/pyramid-cookiecutter-alchemy/pull/7 was merged.

I'll be doing some work on the docs wherever the alchemy cookiecutter is used. I'll create a new branch off of `master` and push that up for ease of collaboration.

stevepiercy:
I've started work in a PR #3307 on a new branch `alembic-cookiecutter-alchemy` in Pyramid.

mmerickel:
Fixed via #3307 and https://github.com/Pylons/pyramid-cookiecutter-alchemy/pull/7. Thanks for the hard work everyone involved!

marioidival:
<3

PR #1625, opened 2015-04-13
Added super basic Alembic configuration to the alchemy paster template

nek4life:
Created a super basic alembic configuration for the alchemy paster template for issue https://github.com/Pylons/pyramid/issues/1597. I'd like to get feedback on the approach and whether or not this is something to be pursued. If so I'll go ahead and create tests and documentation to go along with the patch. I will likely need help writing tests for the template output as it doesn't seem as straight forward as writing a regular unit test.

### DIFF
--- a/pyramid/scaffolds/alchemy/+package+/alembic/README
+++ b/pyramid/scaffolds/alchemy/+package+/alembic/README
@@ -1,0 +1,1 @@
+Generic single-database configuration.
--- a/pyramid/scaffolds/alchemy/+package+/alembic/env.py_tmpl
+++ b/pyramid/scaffolds/alchemy/+package+/alembic/env.py_tmpl
@@ -1,0 +1,73 @@
+from __future__ import with_statement
+from logging.config import fileConfig
+
+from sqlalchemy import engine_from_config, pool
+
+from alembic import context
+
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+fileConfig(config.config_file_name)
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+from {{package}}.models import Base
+target_metadata = Base.metadata
+
+# other values from the config, defined by the needs of env.py,
+# can be acquired:
+# my_important_option = config.get_main_option("my_important_option")
+# ... etc.
+
+
+def run_migrations_offline():
+    """Run migrations in 'offline' mode.
+
+    This configures the context with just a URL
+    and not an Engine, though an Engine is acceptable
+    here as well.  By skipping the Engine creation
+    we don't even need a DBAPI to be available.
+
+    Calls to context.execute() here emit the given string to the
+    script output.
+
+    """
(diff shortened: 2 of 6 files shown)


blaflamme:
@mmerickel BTW should we agree to update the pattern used for SQLA to something similar to this?

https://gist.github.com/blaflamme/0b25c8fed8db0b87b9b6

nek4life:
Back home now... 

So I guess there are still a few things to decide. 

* Should the migrations be located inside or outside of the package.
* Should the folder name be alembic or migrations.
* Should the sqlalchemy be located in the [DEFAULT] section or remain under the [app:main] section in which case the alembic env.py will need to be modified to support this use case.

And there is the matter of testing. I'm currently unsure how to create the necessary tests for the resulting output of the paster templates. I'll try to dig into the code a bit to see what kind of tests are currently written for the paster templates.

pauleveritt:
You could switch to Wichert’s pyramid_sqlalchemy.

—Paul

nek4life:
@pauleveritt I could be mistaken but I didn't see any Alembic support with the pyramid_sqlalchemy package. I also think there was some discussion about setting up the paster template with a non global (or non scoped I think in sqlalchemy lingua.)

BTW is there separate ticket for that?

pauleveritt:
Sorry about that, I was responding to Blaise’s ticket about SQLA setup in Pyramid. You’re right that the topic should be in a different ticket.

—Paul

blaflamme:
@pauleveritt the discussion we had with @mmerickel and others was about avoiding to use sqla's scoped_session for the better

marioidival:
I was hoping someone comment on that issue in which I created, but thank you for making this PR @nek4life

ergo:
What's the consensus on the issue right now?
I think all actual features of new scaffold are implemented so I could also incorporate alembic to it.

marioidival:
@ergo It's nice :+1:

lmctv:
@ergo :+1: from me, even though documentation is completely missing.

mmerickel:
The alchemy scaffolding changes are already massive. I'm pretty afraid to merge this until that dust has settled. In general I'm not particularly excited about merging an alembic setup into pyramid's core scaffold at all but it seems like the idea has a fair amount of support.

Work that needs to be done here to have a hope of merging:

- [ ] Update the `src` folders in the `wiki2` tutorial as they are derived from the scaffold.
- [ ] Update the `wiki2` tutorial to contain an explaination of how to use and modify the setup to perform migrations.

Chicken-and-the-egg here: these changes will likely be easier to do after waiting until the other scaffold changes are complete in #2024.

ergo:
I have also my own solution to alembic - one that integrates with initialize_db script - so no work would be required on tutorials apart changing the location initial model migration.

I'm more concerned about the fact that alembic will not work with sqlite at all - if memory serves me right.

mmerickel:
Alembic has grown support for SQLite over the years but it does have a few quirks as can be expected. 

> so no work would be required on tutorials apart changing the location initial model migration.

I'm saying someone **MUST** update the tutorial to explain the features. :-)

hyperknot:
I might have a look at this project and try to integrate inklesspen's project and the new 1.7 compatibility changes.

mmerickel:
This work should be done on the cookiecutter. Anyone is welcome to contribute alembic support there.

PR #1627, opened 2015-04-14
Add some validation for the JSONP callback

bertjwregeer:
The callback variable could be used to arbitrarily inject javascript
into the response object. This validates that the callback doesn't begin
with a number and is standard US ASCII characters, because trying to
make sure the JavaScript function name is actually valid would require
parsing JavaScript itself...

### DIFF
--- a/pyramid/renderers.py
+++ b/pyramid/renderers.py
@@ -1,6 +1,7 @@
 import contextlib
 import json
 import os
+import re
 
 from zope.interface import (
     implementer,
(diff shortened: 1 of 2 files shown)


bertjwregeer:
Do note that this isn't a HUGE security vulnerability because to allow this to be abused a user has to be able to insert callback variables into the remote page that happens to be doing a JSONP request to a Pyramid application. If you can already insert data into the page to change the callback variable that is sent to Pyramid, you already have access to the site in the first place, and you really don't need to modify the callback variable.

Either way it's a nice to have that we at least sanity check it.

bertjwregeer:
@mcdonc You pulled this into 1.4 and 1.5, but not master, nor 1.6, do you want me to cherry-pick it onto 1.6 too?

mmerickel:
I've applied this to 1.6-branch.

PR #1629, opened 2015-04-14
fixed documentation; wrong line in docs

ztane:
The code in note box in [wiki2 tutorial, Authorization](http://docs.pylonsproject.org/projects/pyramid/en/latest/tutorials/wiki2/authorization.html#login-logout), is from wrong line; should be line 33:

    config.add_route('view_page', '/{pagename}')

but it is from line 32 (which is the `logout` registration from before the box).


tseaver:
Thank you!

PR #1630, opened 2015-04-15
add reference to latest 1.5.5 version

tisdall:
missing a link in the What's New


mmerickel:
Thank you... would you mind updating the RELEASING.txt with a note about needing to update these references, if it isn't already there?

PR #1633, opened 2015-04-15
Restore request.response after render_to_response if renderer raises exception

dairiki:
This fixes two issues with `pyramid.renderers.temporary_response` described in comments on #1563, namely:
- `Request.response` should be restored even the rendered raises an exception.
- If `request.response` is initially set to `None`, it should be restored to `None` (rather than deleted.)

### DIFF
--- a/pyramid/renderers.py
+++ b/pyramid/renderers.py
@@ -142,25 +142,22 @@ def render_to_response(renderer_name,
 
     return result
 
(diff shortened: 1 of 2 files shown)


mmerickel:
Awesome, thank you.

mmerickel:
Backported to 1.6-branch as well.

Issue #406, opened 2012-01-18
route_prefix doesn't allow the prefix pattern to match without a trailing slash


mmerickel:
The issue here is an ambiguity in handling `route_prefix`. In general, things work correctly:

``` python
def routes(config):
    config.add_route('add_user', '/add')

config.include(routes, route_prefix='/users')

# resulting pattern: '/users/add'
```

The ambiguity arises when the included function attempts to add a route using the pattern `'/'` or `''` (an empty string). Pyramid elects to treat the prefix as a container.

``` python
def routes(config):
    config.add_route('users', '/')

config.include(routes, route_prefix='/users')

# resulting pattern: '/users/'
```

Remember that in Pyramid's `config.add_route` there is no difference between prepending a `'/'` or leaving no prefix. Thus, the two route definitions below result in the same url being matched and generated.

``` python
config.add_route('r1', '/route')
config.add_route('r2', 'route')
```

There are 4 possible use cases to account for when looking at the issue.

**Case 1**

``` python
def routes(config):
    config.add_route('route', '')

config.include(routes, route_prefix='/prefix')
```

**Case 2**

``` python
def routes(config):
    config.add_route('route', '/')

config.include(routes, route_prefix='/prefix')
```

**Case 3**

``` python
def routes(config):
    config.add_route('route', '')

config.include(routes, route_prefix='/prefix/')
```

**Case 4**

``` python
def routes(config):
    config.add_route('route', '/')

config.include(routes, route_prefix='/prefix/')
```

**Workaround**

Pyramid currently treats the route prefix as a container, thus the resulting route will always be appended with a `'/'`. This may not be ideal, but it at least allows for a workaround where the user can add their own route (at the level of the include):

``` python
def routes(config):
    config.add_route('add', '/add')

config.include(routes, route_prefix='/users')
config.add_route('users', '/users')
```

If there is a clear way to handle each use-case, and document it as such, then it may be possible to change this behavior but to me it's hard to expect a user to "do the right thing" with case 2 due to the way pyramid currently handles `add_route` patterns with and without a prefixed slash.

sontek:
@mmerickel When you going to fix this so that #406 will just work? =)

bertjwregeer:
I have a feeling that this new bug report is related: https://github.com/Pylons/pyramid/issues/2143

jvanasco:
Regarding the 2.0 (#2362) possibility:

[...]

PR #1639, opened 2015-04-17
Added support for a prefixed base url that does not contain a trailing slash

nek4life:
This was simply a behavior I wanted so I went through to see if I could accomplish hacking on pyramid to make it so. Pyramid's current behavior when supplying an empty string is to match the root of the domain I think this behavior should be carried through to routes with a route prefix. At first I wanted to match both an empty string and '/' but that would break a lot of existing applications that depend on the current behavior.

Right before submitting this patch I looked and found a long discussion for issue #406 which in hindsight I should have read prior to doing this work. I will go back and read through the discussion to see if what I changed makes sense based on the content there.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -230,3 +230,5 @@ Contributors
 - Antti Haapala, 2013/11/15
 
 - Amit Mane, 2014/01/23
+
+- Charlie Choiniere, 2017/04/03
--- a/docs/narr/urldispatch.rst
+++ b/docs/narr/urldispatch.rst
@@ -964,6 +964,24 @@ then only match if the URL path is ``/users/show``, and when the
 :meth:`pyramid.request.Request.route_url` function is called with the route
 name ``show_users``, it will generate a URL with that same path.
 
+To create a prefixed route that matches requests to the ``route_prefix``
+without a trailing slash set the route ``pattern`` to an empty string.
+
+.. code-block:: python
+   :linenos:
+
+   from pyramid.config import Configurator
+
+   def users_include(config):
+       config.add_route('show_users', '')
+
+   def main(global_config, **settings):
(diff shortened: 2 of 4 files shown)


mmerickel:
This is in the discussion for Pyramid 2.0 and I personally think this is a good approach despite it being somewhat bw incompatible. That being said you haven't signed the CONTRIBUTORS.txt agreement yet so I'm closing this for now. If you want to push such a commit we can keep it open.

nek4life:
Hey @mmerickel I'd definitely like to see this functionality for Pyramid 2.0. Are you suggesting merging this prior to that despite it being somewhat bw incompatible or waiting until 2.0? The only reason I ask is I tried to make this semi backwards compatible while getting the new functionality, but I'm wondering if it's for Pyramid 2.0 then would you consider going full backwards incompatible and this merge only being a half measure?

In either case. If this is to be merged I've updated my branch to include my signature to the contributors agreement. I can open a new pull request, just let me know.

Thanks for following up.

mmerickel:
@nek4life If you push to your branch it should update this PR automatically unless Github has gotten confused.

The only incompatibility that I see is in what `''` means for `config.add_route`. I was avoiding revisiting this issue myself until 2.0 probably but noticed that the PR couldn't be used as is without your signature while I was pruning the backlog.

mmerickel:
Ok it found the commit after I reopened the PR.

nek4life:
I agree in it's current form setting an empty string isn't immediately obvious. However, it was in my list of things to try since when I added '/' it appended a slash to the end of the url. So FWIW my first instinct was to replace the slash with an empty string expecting for it to be removed when I did so. At least this way there is an option to remove it from url when the route is prefixed by a group.

Thanks for opening the request back up.

mmerickel:
replaced by #3415

Issue #1554, opened 2015-02-04
pserve --daemon fails with "Bad file descriptor on the use of os.urandom()"


zart:
From IRC discussion:

    <oyse> I have a problem with running pserve --daemon. It fails with Bad file
           descriptor on the use of os.urandom()
    <oyse> Is this a known problem?
    <oyse> I am running Pyramid 1.5.2 on Python 3.4.0
    <oyse> From what I found on internet it could be related to closing of
           filedescriptor, but I would assume that would have been taken care of a
           long time ago if that was the problem
     * Zart ran python2 and python3, and looked at /proc/<pid>/fd/
    <Zart> 2.7 used socket, 3.4 opens /dev/urandom. since pserve closes every fd
           after double fork() this breaks os.urandom() on py3k..
    <Zart> https://docs.python.org/3/whatsnew/3.4.html#significant-optimizations
           seems like it's 3.4 specific optimizations
    <Zart> https://github.com/Pylons/pyramid/blob/master/pyramid/scripts/pserve.py#L412

So it seems that lazy-init optimizations in 3.4 might be a culprit here.

zart:
This might have been fixed in 3.4.1 or later: http://bugs.python.org/issue21207

bertjwregeer:
Downside to trying to be too smart ...

mcdonc:
TBH I'd rather remove the ``--daemon`` feature than add a hack for versions of Python earlier then 3.4.1.  So many good process managers exist; ``pserve`` should not try to be one.

hpfmn:
so what would you recommend to daemonize pserve?

tseaver:
http://supervisord.org/

mmerickel:
@hpfmn yes supervisord works well or every computer comes with its own init system for running services such as sysvinit, upstart or systemd.

bertjwregeer:
the --daemon feature has been deprecated.

Issue #1216, opened 2014-01-02
pserve --stop-daemon does not work with --monitor-restart


arobinson:
If pserve is started with --monitor-restart, the command is restarted when pserve --stop-daemon is run and the pid file's process ID is not updated. The only way to seem to stop it is using kill on the process ID instead of relying on pserve.

piotr-dobrogost:
Just a note that you also use --reload which according to issue #912 is not supported when using --daemon.

bertjwregeer:
@mmerickel and I have fixed the use of `--monitor-restart` with `--daemon`. Don't think it will work with `--reload`.

Either way, this feature is being deprecated.

Issue #912, opened 2013-03-14
pserve --reload with --daemon


mmerickel:
I'm not sure if this is supposed to work, but tizoc in #pyramid claimed it may be a problem. If this is supposed to work we should verify that it does. If it is not supposed to work (which I think would be fine) we should add an error message when trying to execute `pserve` with both options enabled.

piotr-dobrogost:
From what I see --reload does not work with --daemon in Pyramid 1.4.5
What's preventing @mmerickel's fix from getting merged?

mcdonc:
My $.02: I'd rather get rid of ``--daemon`` and ``--stop-daemon`` than trying to make ``pserve`` a full-blown process manager or add complexity to make daemon-stopping work with ``--monitor-restart``.  Very good process managers already exist and pserve doesn't really need to do their jobs.

That said, if @mmerickel wants to merge the patch, I'd be fine with it.

mmerickel:
I'm a big fan of removing `--daemon` but it may be controversial.

Side note: my fix is something I would not consider production ready due to the infinite loop caused by saving a file with broken code while `--monitor-restart` is active.

bertjwregeer:
I'd be all for removing `--daemon`. I never understood why we try to be everything with `pserve`. Then again, I run on top of `uWSGI`.

PR #1641, opened 2015-04-20
deprecate pserve --daemon

mmerickel:
This should fix the following issues: #912, #1216, and #1554.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -202,6 +202,17 @@ Bug Fixes
 Deprecations
 ------------
 
+- The ``pserve`` command's daemonization features have been deprecated. This
+  includes the ``[start,stop,restart,status]`` subcommands as well as the
+  ``--daemon``, ``--stop-server``, ``--pid-file``, and ``--status`` flags.
+
+  Please use a real process manager in the future instead of relying on the
+  ``pserve`` to daemonize itself. Many options exist including your Operating
+  System's services such as systemd or upstart, as well as the excellent
+  and simple supervisord.
+
+  See https://github.com/Pylons/pyramid/pull/1641
+
 - Renamed the ``principal`` argument to ``pyramid.security.remember()`` to
   ``userid`` in order to clarify its intended purpose.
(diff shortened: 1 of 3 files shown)


mmerickel:
@bertjwregeer or @mcdonc can you weigh in on how I should go about deprecating this? I think the PR currently annotates the proper spots in the code for each possible subcommand, but I'm not sure dumping the warnings to stdout is best?

Anyway some feedback and I can fix the tests and merge it.

mcdonc:
I'd probably use warnings.warn instead of dumping to stdout.  One way to do this if each switch ends up calling a method is to use ```zope.deprecation.deprecate``` as a decorator of that method:

```
@deprecate('This is a deprecated method...')
def themethod(self, foo):
    ....
```

mmerickel:
I didn't go with `warnings.warn` because the CLI already outputs some stuff and after python 2.6 no one ever sees the warnings. The only way to turn on Python's deprecation warnings is to not use pserve, and use the module syntax.

```
$ env/bin/python -Wall -m pyramid.scripts.pserve development.ini --status
/Users/michael/work/oss/pyramid/pyramid/scripts/pserve.py:622: DeprecationWarning: The daemon options have been deprecated in Pyramid 1.6. They will be removed
in a future release per Pyramid's deprecation policy. Please consider using
a real process manager for handling your processes like supervisord or systemd.

The following commands are deprecated:
    [start,stop,restart,status] --daemon, --stop-server, --status, --pid-file

  ''', DeprecationWarning, stacklevel=1)
No PID file pyramid.pid
```

Also this is a deprecation of the CLI flags, not a misuse of a python function, so the stacklevel doesn't make as much sense either.

mmerickel:
To be clear, python's `DeprecationWarning` support really is terrible now that they are turned off by default and I wish we used something else like `UserWarning`.

mcdonc:
Yeah, I suppose since the command already dumps to stdout in this context it's fine to use print statements.

PR #1652, opened 2015-04-30
Update logging.rst

areski:
The point 5 from this tutorial http://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/logging.html mention that we should see a debug message. This appears before any needed changes in `development.ini`, thus this patch intend to change the order and add an extra point to put forward changes in the ini file.


mmerickel:
Thank you!

PR #1666, opened 2015-05-11
Update authorization.rst

areski:
fix the url for hello route '/howdy'

### DIFF
--- a/docs/quick_tutorial/authorization.rst
+++ b/docs/quick_tutorial/authorization.rst
@@ -93,7 +93,7 @@ In summary: ``hello`` wants ``edit`` permission, ``Root`` says
 Of course, this only applies on ``Root``. Some other part of the site
 (a.k.a. *context*) might have a different ACL.
 
-If you are not logged in and visit ``/hello``, you need to get
+If you are not logged in and visit ``/howdy``, you need to get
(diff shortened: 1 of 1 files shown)


tseaver:
[`howdy` is not the view name](https://github.com/Pylons/pyramid/blob/e564ff8aa47aed237d7c148056260fb89d781004/docs/quick_tutorial/authorization/tutorial/views.py#L26).

areski:
As the name is starting by a '/' I wasn't sure we refer to the name of the route but rather the url, '/howdy' is the correct url
https://github.com/Pylons/pyramid/blob/e564ff8aa47aed237d7c148056260fb89d781004/docs/quick_tutorial/authorization/tutorial/__init__.py#L22

tseaver:
My bad.

tseaver:
(Or maybe Paul's bad for using a different view / route name than the actual URL name).

areski:
nobody fault to my opinion, but it would have been cleaner to use hello everywhere

pauleveritt:
It’s my style to have the URLs and routes and views have some differences, just to show there isn’t magic in naming.

—Paul

tseaver:
My comment was actually in jest -- having them differ is reasonable.

PR #1670, opened 2015-05-13
add ptpython support for pshell

dakra:
This pull request adds support for ptpython and ptipython in pshell.
Unfortunately ptpython.repl.embed is not capable of displaying a banner when
the shell is created. I opened an issue for it on https://github.com/jonathanslenders/ptpython/issues/41
ptipython does display a banner though.
ptipython is not in the fall-thru chain of shells since it depends on ipython and if ipython is not installed (first fall thru) then it doesn't make sense to try ptipython of course.

### DIFF
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -273,20 +273,21 @@ exposed, and the request is configured to generate urls from the host
 
 .. _ipython_or_bpython:
 
-IPython or bpython
-~~~~~~~~~~~~~~~~~~
+IPython, bpython or ptpython
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you have `IPython <http://en.wikipedia.org/wiki/IPython>`_ and/or
-`bpython <http://bpython-interpreter.org/>`_ in
+If you have `IPython <http://en.wikipedia.org/wiki/IPython>`_,
+`bpython <http://bpython-interpreter.org/>`_ and/or
+`ptpython <https://github.com/jonathanslenders/ptpython>` in
 the interpreter you use to invoke the ``pshell`` command, ``pshell`` will 
 autodiscover and use the first one found, in this order:
-IPython, bpython, standard Python interpreter. However you could 
+IPython, bpython, ptpython, standard Python interpreter. However you could
 specifically invoke one of your choice with the ``-p choice`` or 
 ``--python-shell choice`` option.
 
 .. code-block:: text
 
-   $ $VENV/bin/pshell -p ipython | bpython | python development.ini#MyProject
+   $ $VENV/bin/pshell -p ipython | bpython | ptpython | ptipython | python development.ini#MyProject
 
 .. index::
    pair: routes; printing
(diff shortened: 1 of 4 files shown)


dakra:
Now I updated the ptpython shell to print the banner too.

mmerickel:
I'll be honest, I'm not excited about adding more interpreters to pshell. I'm starting to wonder if we should make this pluggable with extension packages like `pyramid_ptpython`, etc and remove the core support.

dakra:
Ok, if this pull request is not accepted then I'll just make a `pyramid_ptpython` package that has a `pservept`(?) command which is basically pshell.py from my pull request or what's the best way to do it?

PS: The patch is just ~20 lines of actual code.

Just tell me if it has a chance to get accepted or not and I'll then make a separate package if necessary.
Thanks :)

mmerickel:
I think we can make this pluggable very easily by just iterating through a setuptools entrypoint (`pyramid.pshell_interpreter`?) until one works (overridable by specifying a particular one using `-p` like normal). Then we could break out each interpreter extension into its own package and let people install the one they want.

mmerickel:
@mcdonc Can you weigh in on this? I don't mind merging this PR but it's hitting a point where we're adding integrations that we aren't really interested in maintaining.

dakra:
To add another idea. Why not make it completely shell independent and just give the user some bootstrap function that he can run in his own already started shell?

For example I use mostly pycharm and to get `pshell` as my python console in pycharm I created this simple gist https://gist.github.com/dakra/4730736 (basically just slightly modified pshell.py) and have this run after pycharm started my console.

So we could could make a function for that and then the user can start ipython qt/notebook, ptpython, pycharm or whatever and just enter something like
```python
from pyramid import pshell
pshell.bootstrap('development.ini')
```
And then he's all set.

No need at all to support any shells inside the pyramid package.

mmerickel:
@dakra I'm not completely sure how we'd accomplish what you're describing. I think you're talking about passing a dict into `bootstrap` which would get mutated. This dict could be `locals`. Thus something like:

```python
from pyramid.paster import bootstrap
bootstrap('development.ini', env=locals())
```

dakra:
Ah, right.. I forgot that we're calling a function (even from another module) so it's not like in my pycharm startup script.

So it would have to look like your code example.

sontek:
@mmerickel @dakra What do you think about this:

https://github.com/Pylons/pyramid/pull/1891

That will allow you to register new shells via entry points and we can start maintaining each individual shell as separate packages instead of being in core pyramid.

dakra:
@sontek imo looks fine.
Not sure if it's better to just remove all external shells then.
Ipython is kinda standard but if you have a new pyramid environment it wouldn't even be
more "work" for the enduser `pip install pyramid_ipython` (which of course then depends on ipython) instead of just `pip install ipython`.

Advantage is that the shells can update independently from a pyramid release, disadvantage is that you have one more package to install.

sontek:
@dakra Yeah, I figure eventually we'll create pyramid_ipython, pyramid_bpython, etc.   But I feel like this at least gets us in the right direction without losing any backwards compatibility.

mmerickel:
We have added the ability in #1891 to specify extra shells using entry points. I'm hoping that this is good enough to implement pyramid_ptpython but it would be very helpful to us if you could test that out and provide some feedback before it is released in case any changes should be made!

PR #1725, opened 2015-05-22
include tests; grammar; correct title tag

stevepiercy:


stevepiercy:
@pauleveritt in this PR and #1723, #1724, it appears that the references to `tests.py` was omitted, so I added it. However, I'm not sure whether everything that should (or can) be tested is covered. Would you please have a look?

PR #1735, opened 2015-05-23
- add jQuery to the template

stevepiercy:
- punctuation, grammar


stevepiercy:
@pauleveritt this PR and #1733, #1734 add jQuery to the template in `forms/tutorial/wikipage_addedit.pt`. The TinyMCE widget did not render properly without it and other javascripts returned errors stating jQuery did not exist.

However I'm not sure whether this is the preferred way of doing so. Would you please review this and let me know whether I should adapt anything?

PR #1814, opened 2015-05-30
Added note on check_csrf view predicate

kissgyorgy:
Also it is an add_view parameter, not add_route.


stevepiercy:
@kissgyorgy Would you please backtick the expression and wrap to 79 columns?
```
   so ``check_csrf=True`` behavior
```

kissgyorgy:
Did it and rebased.

stevepiercy:
Thank you!

PR #1826, opened 2015-06-02
Add AlchemyDox Template

danielpronych:
Add an additional alchemy template that contains docstrings as out-of-the-box documentation within a new alchemy template, with the name alchemydox, without modifying the original alchemy template (though based on this template). This entry_point, alchemydox, has been added to setup.py and is accessible for creating a new application based on this template. Though, not necessary, the
application documentation, if created with Doxygen, is best created with the doxypy.py script as an INPUT_FILTER. Inputted documentation is still useful even if not using Doxygen.

Example documentation pages can also be provided, such as in the gh-pages branch (if this considered helpful) to show that this works correctly. The related doxyfile or doxypy, is not included at this time; however, it can also be included if helpful. The doxypy script is available on GitHub. As stated in the previous paragraph, there is no requirement to actually use Doxygen at any time with an application using this scaffold; however, this does lay the groundwork to make it easier to use Doxygen later (if a developer/development team so choose). Other notes with primary changes:

* All files in this end in _tmpl since the package name documentation is defaulted into all python scripts and the package names are detailed in docstrings at application creation.
* When the setup.py file has descriptions in generated documentation.

Thanks for your consideration of these changes.


mcdonc:
Thanks for the pull request!

However, I'd suggest packaging this scaffold in its own distribution and adding it to PyPI, even if you only want to reuse it for your own project and if other folks are interested in it, great too.   But Iit's unlikely we'll ever add more scaffolds to Pyramid itself.  Pyramid's scaffolding story is meant to be extensible without needing to change Pyramid itself.  You can add a ``[pyramid.scaffold]`` entry point to any distribution you create and upload to PyPI.

danielpronych:
Thanks for your response, as it is considered unlikely that additional scaffolds be added to the core, and since only minor modifications (essentially docstrings and headers for all basic Python scripts, have technically be made to the `alchemy` template with a different scaffold name) have been made. Only a few minor changes needing to be made apply these changes specifically to the `alchemy` scaffold rather than create a new scaffold. Please advise if this would be feasible for inclusion if these changes were applied to the primary template instead.  Looking forward to your response!

mcdonc:
Sorry, no.  Rationales are as follows:

- The PR seems focused on creating Doxygen-compatible docstrings within
  code in the scaffold.  Folks who use the scaffold are people who are probably
  very new to Pyramid and whom may even be new to Python.  Seeing the 
  markup is going to make them wonder what it's for, and (for better or worse)
  may make them think it's "Pyramid magic", and we want to keep things
  as simple as possible here at that level.

- The bar for adding code to a scaffold that ships with Pyramid is high because
  the tutorials are based on scaffolding output, and the binding is very tight.  For 
  an example of how tight the bond is, see https://raw.githubusercontent.com/Pylons/pyramid/master/docs/tutorials/wiki2/definingmodels.rst
  and notice that we mention actual line numbers in the tutorial.  Even minor
  changes to scaffolds have a ripple effect of needing to change many lines
  across many files in tutorials.  Of course, this is not a good thing, but we've
  been unable to come up with a way to avoid it without making the tutorial
  experience suffer.

- It's already dead-easy to create a distribution that houses a scaffold.  Given the
  above two points and the fact it's so easy, and the goal is to create something
  reusable for *yourself*, you should do that.

PR #1827, opened 2015-06-02
Trivial highlighting errors in the docs

hegazy:
I just noticed it while reading the docs.

### DIFF
--- a/docs/tutorials/wiki2/authorization.rst
+++ b/docs/tutorials/wiki2/authorization.rst
@@ -289,17 +289,17 @@ follows:
 
 .. literalinclude:: src/authorization/tutorial/views.py
    :lines: 57-58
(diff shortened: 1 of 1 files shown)


tseaver:
Thanks for the patch!  I'm afraid that highlighting only half the `@view_config()` decorator invocations will make the example pretty confusing.  What issue were you trying to solve?

hegazy:
The issue is that those lines were not consistent with all the examples in the wiki that is highlighting only the piece of code that got changed no matter what.

stevepiercy:
Sorry, there is an additional trailing comma on the previous line that needed to be added in the code, and therefore highlighted. It was too easy to miss, as this PR just demonstrated. ;)

Closing.

PR #1837, opened 2015-06-10
handle conflicting project names gracefully, fixes: #1357

jenner:
- Check for importable packages/modules with the same
  name as the package to be created
- Added --force-conflicting-name option to enforce the
  use of a conflicting package
- Added appropriate tests

### DIFF
--- a/pyramid/scripts/pcreate.py
+++ b/pyramid/scripts/pcreate.py
@@ -9,6 +9,8 @@ import pkg_resources
 import re
 import sys
 
+user_input = input if sys.version_info[0] == 3 else raw_input
+
 _bad_chars_re = re.compile('[^a-zA-Z0-9_]')
 
 def main(argv=sys.argv, quiet=False):
@@ -92,8 +94,6 @@ class PCreateCommand(object):
 
     @property
     def project_vars(self):
-        options = self.options
-        args = self.args
         output_dir = self.output_path
         project_name = os.path.basename(os.path.split(output_dir)[1])
         pkg_name = _bad_chars_re.sub(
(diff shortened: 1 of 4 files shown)


pyramid/scripts/pcreate.py
```diff
@@ -9,11 +9,20 @@
 import re
 import sys
 
+if sys.version_info[0] == 3:
+    user_input = input # pragma: no cover
+else:
+    user_input = raw_input # NOQA
```
mmerickel: https://github.com/Pylons/pyramid/blob/master/pyramid/compat.py#L212

pyramid/scripts/pcreate.py
```diff
@@ -154,5 +169,48 @@ def out(self, msg): # pragma: no cover
...
+            self.out('')
+            self.show_scaffolds()
+            return False
+        if not self.args:
+            self.out('You must provide a project name')
+            return False
+        available = [x.name for x in self.scaffolds]
+        diff = set(self.options.scaffold_name).difference(available)
+        if diff:
+            self.out('Unavailable scaffolds: %s' % ", ".join(list(diff)))
```
mmerickel: possibly `sorted(diff)` for better output

pyramid/scripts/pcreate.py
```diff
@@ -154,5 +169,48 @@ def out(self, msg): # pragma: no cover
...
+            return False
+
+        pkg_name = self.project_vars['package']
+
+        if pkg_name == 'site' and not self.options.force_bad_name:
+            self.out('The package name "site" has a special meaning in '
+                     'Python. Are you sure you want to use it as your '
+                     'project\'s name?')
+            return self.confirm_bad_name('Really use "{0}"?: '.format(pkg_name))
```
mmerickel: Can we just roll the `site` test into the `__import__` below? It's an importable package just like anything else we're testing.

pyramid/scripts/pcreate.py
```diff
@@ -154,5 +169,48 @@ def out(self, msg): # pragma: no cover
...
+        try:
+            __import__(pkg_name, globals(), locals(), [], 0) # use absolute imports
+        except ImportError as error:
+            pkg_exists = False
+        if not pkg_exists:
+            return True
+
+        if self.options.force_bad_name:
+            return True
+        self.out('Package "{0}" already exists, are you sure you want '
```
mmerickel: I think "A package named "{0}" already exists..." is better than saying "Package foo already exists" as the latter implies that it's the package you want, not some other package.

pyramid/scripts/pcreate.py
jenner: Yeah, but site isn't a package, so it deserves a special treatment IMO, otherwise rookie devs will try to find a "site" package in their current env and wonder what the general error message ('Package "site" already exists...') is supposed to mean

PR #1843, opened 2015-06-12
Add new event BeforeAplicationCreated

uralbash:
This event allows you to add the settings in any order. For example, I configured my include application:

```python
# myapp.__init__.py

    config.include("pyramid_pages")
    ...
    settings[CONFIG_MODELS] =\
            {
                '': WebPage,
                'pages': WebPage,
                'news': NewsResource,
                'gallery': GalleryResource,
            }
```

The following code raises an exception because these  ``settings[CONFIG_MODELS]`` settings are no:

```python
# pyramid_pages.__init__.py

def includeme()
    config = event.config
    # Models from setting factory
    settings = config.registry.settings
    models = settings[CONFIG_MODELS]
    for resource in models.values():
        if hasattr(resource, '__table__') and not hasattr(resource, 'model'):
            continue

        config.add_view(resource.view,
                        attr=resource.attr,
                        route_name=PREFIX_PAGE,
                        renderer=resource.template,
                        context=resource,
                        permission=PREFIX_PAGE)
```

Therefore, it would be very convenient to add this before making an WSGI application.

```python
@subscriber(BeforeApplicationCreated)
def resource_routes(event):
    config = event.config
    # Models from setting factory
    settings = config.registry.settings
    models = settings[CONFIG_MODELS]
    for resource in models.values():
        if hasattr(resource, '__table__') and not hasattr(resource, 'model'):
            continue

        config.add_view(resource.view,
                        attr=resource.attr,
                        route_name=PREFIX_PAGE,
                        renderer=resource.template,
                        context=resource,
                        permission=PREFIX_PAGE)
```

### DIFF
--- a/pyramid/config/__init__.py
+++ b/pyramid/config/__init__.py
@@ -30,7 +30,7 @@ from pyramid.compat import (
     zip_longest,
     )
 
-from pyramid.events import ApplicationCreated
+from pyramid.events import ApplicationCreated, BeforeApplicationCreated
 
 from pyramid.exceptions import (
     ConfigurationConflictError,
@@ -975,11 +975,14 @@ class Configurator(
 
     def make_wsgi_app(self):
         """ Commits any pending configuration statements, sends a
-        :class:`pyramid.events.ApplicationCreated` event to all listeners,
-        adds this configuration's registry to
-        :attr:`pyramid.config.global_registries`, and returns a
-        :app:`Pyramid` WSGI application representing the committed
+        :class:`pyramid.events.BeforeApplicationCreated` event to all
+        listeners, adds this configuration's registry to
+        :attr:`pyramid.config.global_registries`, then sends a
+        :class:`pyramid.events.ApplicationCreated` event to all listeners, and
+        returns a :app:`Pyramid` WSGI application representing the committed
         configuration state."""
+        self.commit()
+        self.registry.notify(BeforeApplicationCreated(self))
         self.commit()
         app = Router(self.registry)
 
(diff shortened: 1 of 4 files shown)


mmerickel:
Pyramid's configurator is actually already built to support this un-ordered, delayed configuration, that you are trying to accomplish, in a much more robust way.

It has an actions api that defers things until commit. Your even completely circumvents the commit cycle which is no good.

If you want to fix something like `pyramid_pages` to allow you to change settings after including the addon, but before a commit (which is what `config.make_wsgi_app()` does as well), then the `includeme` or any other directive that does work should be doing its work in a deferred callable via `config.action`. This gives a lot of benefits[1] including the fact that settings can be mutated prior to the commit.

[1] http://docs.pylonsproject.org/projects/pyramid/en/1.5-branch/narr/advconfig.html

uralbash:
Thanks, 

``config.action`` that's what I need. However, this code works only with version of pyramid >= 1.6a1

```python

def hello(request):
    return Response('Hello world!')

def register(*args):
    config = args[0]
    settings = config.get_settings()
    route_name = settings['route_name']
    config.add_route(route_name, '/hello2')
    config.add_view(hello, route_name=route_name)
    print('Register')
    config.commit()

def foo(config):
    config.action(None, register, args=(config, ))

if __name__ == '__main__':
    config = Configurator()
    config.add_route('hello_world', '/')
    config.add_view(hello, route_name='hello_world')

    config.include(foo)
    config.get_settings()['route_name'] = 'foo'

    print("Before make wsgi app")
    app = config.make_wsgi_app()
    print("After make wsgi app")
    server.serve_forever()
```

mmerickel:
You're right, the configurator has gone through some significant improvements for addons in 1.6.

PR #1844, opened 2015-06-13
metadata: add proper naming convention to metadata - same as suggeste…

ergo:
metadata: add proper naming convention to metadata - same as suggested by alembic/sqlalchemy documentation

This pull request implements what Mike Bayer suggests:
http://alembic.readthedocs.org/en/latest/naming.html


mmerickel:
Thanks @ergo!

PR #1860, opened 2015-07-22
Added application 'name' parameter to bootstrap method

chipkellam:
Added a name parameter to bootstrap function that allows specifying the exact application you want.  Useful when multiple apps are configured, eg. when using Paste#urlmap


mmerickel:
Sorry I just replied to your mailing list post but this is redundant with already-supported functionality in the `config_uri` format.

chipkellam:
I see that line in the documentation now, I totally missed it.  Good to know!

PR #1861, opened 2015-07-23
Use correct cache argument in cache busting doco

davidjb:
Minor correction; it's ``cache_max_age`` rather than ``max_cache_age``.


stevepiercy:
Thank you!

PR #1866, opened 2015-08-03
base64 encode cookie values

bertjwregeer:
This way they are valid according to the RFC, and newer versions of WebOb (>=1.5.0) won't complain about cookie values containing invalid bytes.

Testing completed this way:

    python -W error -W ignore::PendingDeprecationWarning -W ignore::DeprecationWarning setup.py test

Using Python 3.4.3.

### DIFF
--- a/pyramid/tests/test_session.py
+++ b/pyramid/tests/test_session.py
@@ -1,3 +1,4 @@
+import base64
 import json
 import unittest
 from pyramid import testing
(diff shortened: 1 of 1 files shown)


bertjwregeer:
Arghhh, of course Python 2.x doesn't want to play ball. Will figure this out later.

bertjwregeer:
I am at a loss as to why Python 2.x doesn't want to work.

mmerickel:
Fails on PyPy as well.

mmerickel:
Fun sleuthing but the error is because of how base64/binascii changed in py2 [1] vs py3 [2]. Basically py2 almost always raises a `TypeError` whereas py3 more intelligently raises a `ValueError` or subclass `binascii.Error` in the appropriate scenarios. So `pyramid.tests.test_session.DummySerializer` was broken.

[1] https://hg.python.org/cpython/file/v2.7.10/Lib/base64.py#l76
[2] https://hg.python.org/cpython/file/v3.2.1/Lib/base64.py#l68

PR #1883, opened 2015-08-27
Preserve capitalization of Custom Variables

invisibleroads:
Currently, ConfigParser applies lowercase to keys when parsing a configuration file like ``development.ini``. For easier cut and paste testing in ``pshell``, it would be great if pshell preserves case exactly as specified in the configuration file.

    [pshell]
    User = xyz.models.User
    m = xyz.models
    db = xyz.models.db

http://stackoverflow.com/questions/19359556/configparser-reads-capital-keys-and-make-them-lower-case

### DIFF
--- a/pyramid/scripts/pshell.py
+++ b/pyramid/scripts/pshell.py
@@ -61,6 +61,7 @@ class PShellCommand(object):
 
     def pshell_file_config(self, filename):
         config = self.ConfigParser()
+        config.optionxform = str
         config.read(filename)
         try:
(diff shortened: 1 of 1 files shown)


tseaver:
Thanks for the patch!  Could you please add an entry to `CHANGES.rst` describing the change?

mmerickel:
It'd be helpful if you could add or tweak one of the tests to exercise this as well.

mmerickel:
Closing in favor of #2012.

PR #1891, opened 2015-09-06
Use entry points for pshell

sontek:
This also drops support for versions of ipython from 2010 and 2011, I figure we can deprecate the usage and if someone still needs it they can create a `pyramid_ipython_2010` package that easily does it.

This is in response to #1670, with this we can of `pyramid_ptpython` package that provides the functionality.

### DIFF
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -245,7 +245,7 @@ exposed, and the request is configured to generate urls from the host
 .. code-block:: text
 
     $ $VENV/bin/pshell starter/development.ini
-    Python 2.6.5 (r265:79063, Apr 29 2010, 00:31:32) 
+    Python 2.6.5 (r265:79063, Apr 29 2010, 00:31:32)
     [GCC 4.4.3] on linux2
     Type "help" for more information.
 
@@ -278,10 +278,10 @@ IPython or bpython
 
 If you have `IPython <http://en.wikipedia.org/wiki/IPython>`_ and/or
 `bpython <http://bpython-interpreter.org/>`_ in
-the interpreter you use to invoke the ``pshell`` command, ``pshell`` will 
+the interpreter you use to invoke the ``pshell`` command, ``pshell`` will
 autodiscover and use the first one found, in this order:
-IPython, bpython, standard Python interpreter. However you could 
-specifically invoke one of your choice with the ``-p choice`` or 
+IPython, bpython, standard Python interpreter. However you could
+specifically invoke one of your choice with the ``-p choice`` or
 ``--python-shell choice`` option.
 
 .. code-block:: text
@@ -308,19 +308,15 @@ and then your shell factory should return a function that accepts two arguments,
 .. code-block:: python
 
     def ptpython_shell_factory():
-        try:
-            from ptpython.repl import embed
-            def PTPShell(banner, **kwargs):
-                print(banner)
-                return embed(**kwargs)
-        except ImportError:
-            return None
+        from ptpython.repl import embed
+        def PTPShell(banner, **kwargs):
+            print(banner)
+            return embed(**kwargs)
 
         def shell(env, help):
             PTPShell(banner=help, locals=env)
 
         return shell
-
 
 .. index::
    pair: routes; printing
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -288,6 +288,40 @@ specifically invoke one of your choice with the ``-p choice`` or
 
    $ $VENV/bin/pshell -p ipython | bpython | python development.ini#MyProject
 
+Alternative Shells
+~~~~~~~~~~~~~~~~~~
+If you want to use a shell that isn't supported out of the box you can introduce
+a new shell by registering an entrypoint in your setup.py:
+
+.. code-block:: python
+
+    setup(
+        entry_points = """\
+            [pyramid.pshell]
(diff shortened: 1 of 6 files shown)


sontek:
@mmerickel please review

mmerickel:
This looks awesome. One minor request: can you please move the `make_XXX` methods into module-level functions instead of classmethods?

sontek:
I was originally going to do that but it would require significant changes to the tests and without mock would be harder to test

docs/narr/commandline.rst
```diff
@@ -288,6 +288,40 @@ specifically invoke one of your choice with the ``-p choice`` or
...
+
+.. code-block:: python
+
+    def ptpython_shell_factory():
+        try:
+            from ptpython.repl import embed
+            def PTPShell(banner, **kwargs):
+                print(banner)
+                return embed(**kwargs)
+        except ImportError:
```
mmerickel: Is the conditional import required for someone making a package binding themselves? I guess it doesn't really matter but thought I'd bring it up.

sontek: The conditional is more of a "best practice" so if they try to use ptpython and don't have it installed it'll fall back to the standard shell but its not required.   Do you think it complicates the example?

mmerickel: I'm leaning toward taking it out but let's involve @bertjwregeer as a tie breaker and leave it at whatever he says.

bertjwregeer: If we are going to go this route, I personally believe that if the shell you select on the command line doesn't exist or is broken it should be explicitly made clear to the user that it isn't possible to start pshell.

i.e. I am against fallback to a default shell. If they are explicitly requesting a particular shell we shouldn't return one that is not the one they requested.

mmerickel: I think that's a good point about how to handle fallbacks. We should probably make that change such that if the shell is explicitly requested then we raise an error instead of doing the fallback. However that doesn't directly relate to this particular piece of code. This is just a docs example on how someone should write a binding.

bertjwregeer: I don't think the conditional import is required in sample code; I much prefer a stack trace to silently failing (and falling back).

mmerickel: @sontek can you fix this up and I'll merge it?

pyramid/scripts/pshell.py
```diff
@@ -159,18 +162,33 @@ def run(self, shell=None):
             closer()
 
     def make_shell(self):
+        shells = {}
+
+        priority_order = ['ipython', 'bpython']
```
mmerickel: I'm not sure there's any real reason why we need to keep these at a higher priority than other entry points.

pyramid/scripts/pshell.py
```diff
@@ -159,18 +162,33 @@ def run(self, shell=None):
...
-            shell = self.make_ipython_shell()
+                if shell is not None:
+                    break
+        else:
+            factory = shells.get(user_shell)
 
-        elif user_shell == 'bpython':
-            shell = self.make_bpython_shell()
+            if factory is not None:
+                shell = factory()
```
mmerickel: I think just add `else: raise ValueError('could not find a shell named ...')` or something to that effect to avoid fallbacks when the user explicitly requested a specific shell.

sontek: Do you think raising an exception here is better than printing out a nice error message?

sontek:
@mmerickel @bertjwregeer I think I fixed everything you suggested.

pyramid/scripts/pshell.py
sontek: Only reason I have priority order here is for backwards compatibility will the fall through.

pyramid/scripts/pshell.py
mmerickel: Sure, an error message would be nicer.

sontek: Yeah, I have it do `self.out(...)` now

mmerickel:
Good stuff. I'll backport it to 1.6.

stevepiercy:
@sontek @mmerickel I'm doing a docs cleanup of commandline.rst for 1.5 and later. Is this feature applicable for 1.6 and later only? If so, I'll add some versionadded or versionchanged Sphinx directives as needed.

dakra:
I just uploaded a simple ``pyramid_ptpython`` ( https://github.com/dakra/pyramid_ptpython ) package for it.
A few things I noticed:
1. https://github.com/Pylons/pyramid/blame/master/docs/narr/commandline.rst#L301 should be something like ``myshell=myshell_package:myshell_factory`` (with `:`)
2. ``pshell --help`` shows fix ``ipython | bpython | python`` as available shells.
3. Would be nice if I can somehow configure a default shell in the ini file instead of always specifying ``-p ptpython``

mmerickel:
@stevepiercy 
This is 1.6+.

@dakra 
1. Yup! I will fix this.
2. Yup, we should update the message to show all of the available shells.
3. So, in theory your plugin *should* get auto-selected. I'd be in favor of a patch that preferred any non-builtin entry points. At that point it's only an issue when you have 2 plugins installed at which point we could probably expose a `preferred_shells = ptpython, ipython` in the `[pshell]` setting. Let me know what your testing reveals about this ordering issue.

mmerickel:
@dakra can you please look at #2012 and move your concerns over to that PR please?

PR #1892, opened 2015-09-12
fix typo in docs

uralbash:


stevepiercy:
Neither the current version nor your PR are correct. It should be "class's". I know it looks funky, but that is what I found in most English style guides. I'm going to close this and submit a correction. Thank you for bringing this up.

PR #1900, opened 2015-10-01
wiki2 documentation precisely matches the scaffolding output, py.test environment prefix

nharringtonwasatch:
As the user progresses through the wiki2 tutorial, the models.py file differs from the provided documentation at "Content Models with models.py" in basiclayout.html. This can be distracting as the narrative tutorial and previous steps precisely match the expected output.

Updated the referenced models.py file to have matching text to that generated by the sqlalchemy scaffold, inform the user about the subsequent deletions in the next step. These deletions should be covered by the line:

```The highlighted lines are the ones that need to be changed, as well as removing lines that reference Index.```

from the first section in "Defining the Domain Model"

### DIFF
--- a/HACKING.txt
+++ b/HACKING.txt
@@ -189,7 +189,7 @@ Running Tests
   Run the tests like so:
 
    $ $VENV/bin/easy_install pytest
-   $ py.test --strict pyramid/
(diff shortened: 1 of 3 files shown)


stevepiercy:
Nice catches. Thank you!

PR #1907, opened 2015-10-01
Mention Circus as a pserve --daemon replacement. Refs #1554, #1641

mnaberez:
[Circus](https://github.com/circus-tent/circus) is a newer Python-based process manager similar to Supervisor. It is worth mentioning as a ``pserve --daemon`` replacement.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -215,8 +215,8 @@ Deprecations
 
   Please use a real process manager in the future instead of relying on the
   ``pserve`` to daemonize itself. Many options exist including your Operating
-  System's services such as systemd or upstart, as well as the excellent
-  and simple supervisord.
+  System's services such as Systemd or Upstart, as well as Python-based
+  solutions like Circus and Supervisor.
 
   See https://github.com/Pylons/pyramid/pull/1641
 
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -622,7 +622,7 @@ class PServeCommand(object):
         self.out('''\
 The daemon options have been deprecated in Pyramid 1.6. They will be removed
 in a future release per Pyramid's deprecation policy. Please consider using
-a real process manager for handling your processes like supervisord or systemd.
+a real process manager for your processes like Systemd, Circus, or Supervisor.
 
 The following commands are deprecated:
     [start,stop,restart,status] --daemon, --stop-server, --status, --pid-file


bertjwregeer:
:-1: The only reason supervisord is listed is because it was originally built by @mcdonc. I don't think it is a good idea to keep adding to that list, what about nosh, deamontools, s6, and others? Should all of those be mentioned?

systemd and upstart are from Red Hat and Ubuntu respectively, two of the largest Linux distributions, and supervisor was mentioned as an alternative (heavily used too).

mnaberez:
> The only reason supervisord is listed is because it was originally built by @mcdonc. I don't think it is a good idea to keep adding to that list, what about nosh, deamontools, s6, and others? Should all of those be mentioned?

I didn't realize that's the only reason.  I figured it was listed because it's Python-based and popular with Python users.  Circus is newer but also seems popular with Python users.  I don't think every tool should be listed, but I think suggesting the two major Python-based solutions would probably be helpful since Pylons users probably like Python.

mmerickel:
I'm more likely to remove the suggested alternatives than to add more process managers to the list. If you feel like circus is being slighted would you prefer me to do that?

mnaberez:
> I'm more likely to remove the suggested alternatives than to add more process managers to the list. If you feel like circus is being slighted would you prefer me to do that?

I think the suggestions in the message are a good idea if `pserve --daemon` is deprecated.  Without any suggestion, users are kind of left with a dead end when they see the deprecation ("ok, I shouldn't use `--daemon`, now what?").  At least it gives them some keywords as a starting point.

The listed OS-level solutions, Upstart and Systemd, are both widely used and it makes sense to mention both.  The Python-based solutions Supervisor and Circus have some overlap but they're not quite the same type of tool (not designed to be "pid 1", written in Python).  They're both popular with Python users in particular and each has their own pros and cons.  I think it would be good to mention both of them.

bertjwregeer:
Define popular, until today I had never heard of Circus, and I am fairly well engrained within the Python community. Supervisor on the other hand is really popular even outside of the Python users due to it's heavy use in Docker at one point in time (not sure if that is still the case).

cguardia:
Not saying that should be the main criterion, but supervisor has almost 400,000 downloads this month on pypi, and circus around 11,000. Don't know if circus is usually installed in some other way, but it doesn't seem as they are in the same order of magnitude, popularity wise.

mnaberez:
> Define popular, until today I had never heard of Circus, and I am fairly well engrained within the Python community.

Circus has enough PyPI downloads and GitHub stars where I think saying it is popular shouldn't be too controversial.  However, I don't think I compared its popularity to Supervisor, and that wasn't my intention.  What I tried to say above is that they are I think there are two types of these tools, the "pid 1" solutions (Upstart, Systemd) and the Python-based solutions (Supervisor, Circus).  Since the Python-based solutions are especially interesting to Python users, and there are definitely differences between them, I think it would be helpful to list the two.

sontek:
I think giving people suggestions is better than removing them completely and although circus may not be as popular, I have heard of it and know companies like Mozilla are using it with great success.  

Since @mnaberez is the maintainer of supervisor and hes suggesting we include circus as an alternative I'm thinking hes probably a good authority on what we should suggest here =)

mmerickel:
Alright, thanks @mnaberez!

PR #1928, opened 2015-10-07
Small fixes to the docs

bertjwregeer:
@stevepiercy these are all for you =)

Some simple fixes, I know that these two also apply to at least the 1.5-branch, not sure if they also need to get back ported to the 1.4-branch.


stevepiercy:
Yeah, I pooched this one with wrapping to 80 columns instead of 79, too. Thanks for the report and PR. I'll do some mop up tonight.

PR #1929, opened 2015-10-07
populate pshell's env_help with docstring of values

jessedhillon:
This changes `pshell` to check the docstring of values assigned to `env`, and if present, use it to as the value for `env_help` for that value instead of the `repr`. For example:

**Before**
```shell
Environment:
  app          The WSGI application.
  pp           <function pprint at 0x7f1b37960050>
  registry     Active Pyramid registry.
  request      Active request object.
  root         Root of the default resource tree.
  root_factory Default root factory used to create `root`.
  transction   <module 'transaction' from '/home/jesse/.virtualenvs/boutique/local/lib/python2.7/site-packages/transaction/__init__.pyc'>
```

**After**
```shell
Environment:
  app          The WSGI application.
  pp           Pretty-print a Python object to a stream [default is sys.stdout].
  registry     Active Pyramid registry.
  request      Active request object.
  root         Root of the default resource tree.
  root_factory Default root factory used to create `root`.
  transaction  Exported transaction functions.  $Id$
```

_SooooOOO nice!_

:haircut: :dancers: :+1:

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -250,3 +250,5 @@ Contributors
 - Karen Dalton, 2015/06/01
 
 - Igor Stroh, 2015/06/10
+
+- Jesse Dhillon, 2015/10/07
--- a/pyramid/scripts/pshell.py
+++ b/pyramid/scripts/pshell.py
@@ -122,7 +122,10 @@ class PShellCommand(object):
             # remove any objects from default help that were overidden
             for k, v in env.items():
                 if k not in orig_env or env[k] != orig_env[k]:
-                    env_help[k] = v
+                    if getattr(v, '__doc__', False):
+                        env_help[k] = v.__doc__.replace("\n", " ")
+                    else:
+                        env_help[k] = v
 
         # load the pshell section of the ini file
         env.update(self.loaded_objects)
--- a/pyramid/tests/test_scripts/test_pshell.py
+++ b/pyramid/tests/test_scripts/test_pshell.py
@@ -289,6 +289,7 @@ class TestPShellCommand(unittest.TestCase):
         def setup(env):
             env['a'] = 1
             env['root'] = 'root override'
+            env['none'] = None
         self.config_factory.items = [('setup', setup)]
         shell = dummy.DummyShell()
         command.run(shell)
@@ -302,6 +303,7 @@ class TestPShellCommand(unittest.TestCase):
             'request':self.bootstrap.request,
             'root_factory':self.bootstrap.root_factory,
             'a':1,
+            'none': None,
         })
         self.assertTrue(self.bootstrap.closer.called)
         self.assertTrue(shell.help)


bertjwregeer:
Please fix coverage in tests, and repush to this PR: https://travis-ci.org/Pylons/pyramid/jobs/84021793

jessedhillon:
I'm not familiar with this CI system, can you give any suggestions about what's the issue and how to repro?

stevepiercy:
@jessedhillon It's not really the CI, it's the lack of coverage for this feature. To avoid this, always run tests locally before committing. Instructions are in [HACKING.txt](https://github.com/Pylons/pyramid/blob/master/HACKING.txt).

For this specific error, click [Details](https://travis-ci.org/Pylons/pyramid/builds/84021783) then click on the [failing job](https://travis-ci.org/Pylons/pyramid/jobs/84021793). Here's the [pertinent line](https://travis-ci.org/Pylons/pyramid/jobs/84021793#L919).

Note to self, I think I need to finally get around to creating that [contributing guidelines file](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) for Pyramid.

bertjwregeer:
Sure, Travis CI just runs all of our tox environments, you can do so locally too:

1. Download and install tox (easy_install tox or pip install tox)
2. At the top-level directory run (to get the coverage report):
    - tox -e py2-cover,py3-cover,coverage

You'll need to make sure that you have both Python 2 (as a binary named python2.7) and Python 3 (as python3.4) installed.

This will run the tests under python 2, collecting information about what lines are covered/not covered, then as python 3 and find out what lines are covered/not covered by tests, and then the last one will merge the two results so that any python 2 or python 3 only code paths don't count against the total lines covered.

jessedhillon:
I see what's going on now, thanks. It's complaining that there's no test that touches the case where the object has a falsy docstring. I'll add that test and resubmit. Aside from the test, do you have any comments on the feature itself or the way it's been implemented?

bertjwregeer:
Nope, looks good to me :-).

bertjwregeer:
Sorry, I should have been better about including what was wrong, rather than just giving you a link to Travis. That's my bad!

jessedhillon:
No worries, I saw the build was failing but didn't really understand the output of the test runner because I was looking for a failing test. I've added a test (did a forced push, didn't want to create a new commit for two lines) that adds `None` to the `env`. `None.__doc__` is `None` also, so that hits the falsy case now.

jessedhillon:
Pyramid is the only sane framework AFAIAC, so I'm happy to be able to make a contribution and look forward to making more.

bertjwregeer:
Last but not least to speed up getting this accepted, add a commit that signs `CONTRIBUTORS.txt` please.

Looks good :+1:

jessedhillon:
Signed

bertjwregeer:
@mmerickel This is good to go :-)

mmerickel:
Great, I'll merge this when I have a second to backport it to 1.6-branch and update the changelog.

PR #1972, opened 2015-10-12
backport #1891 - Use entry points for pshell

mmerickel:


CHANGES.txt
```diff
@@ -164,6 +164,9 @@ Features
   that as the response class instead of the default ``HTTPFound``.  See
   https://github.com/Pylons/pyramid/pull/1610
 
+- Additional shells for ``pshell`` can now be registered as entrypoints. See
+  https://github.com/Pylons/pyramid/pull/1891
+
```
stevepiercy: Should this be moved to "unreleased" instead of added to "1.6a1 (2015-04-15)"?

mmerickel: It's already moved in a later commit. Thanks for noticing the error though.

PR #1975, opened 2015-10-12
add py35 tests

sontek:


bertjwregeer:
For this to work you will need to explicitly declare your Python version in a matrix since Python 3.5 is not yet installed as python3.5 on travis:

For example see:

https://github.com/Pylons/pyramid/blob/cc9126ed2b215fe6c59ca9eb74e6e0a55b9fc76e/.travis.yml

bertjwregeer:
That URL makes it seem like I pushed to the Pyramid repo, but I assure you I definitely did not: https://github.com/bertjwregeer/pyramid/blob/fix/python3.5/.travis.yml

Issue #1973, opened 2015-10-12
Python 3.5 test failures on master


bertjwregeer:
======================================================================
    FAIL: test_commit_conflict_with_two_includes (pyramid.tests.test_config.test_init.ConfiguratorTests)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/Users/xistence/Projects/pyramid/pyramid/tests/test_config/test_init.py", line 1112, in test_commit_conflict_with_two_includes
        config.commit()
    pyramid.exceptions.ConfigurationConflictError: Conflicting configuration actions
      For: ('view', None, '', None, 'd41d8cd98f00b204e9800998ecf8427e')
        Line 526 of file /Users/xistence/Projects/pyramid/pyramid/util.py:
            return wrapped(self, *arg, **defaults)
        Line 526 of file /Users/xistence/Projects/pyramid/pyramid/util.py:
            return wrapped(self, *arg, **defaults)
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/Users/xistence/Projects/pyramid/pyramid/tests/test_config/test_init.py", line 1116, in test_commit_conflict_with_two_includes
        self.assertEqual(c1, 'includeme1')
    AssertionError: 'wrapper' != 'includeme1'
    - wrapper
    + includeme1
    
    
    ======================================================================
    FAIL: test_conflict_set_forbidden_view (pyramid.tests.test_config.test_init.ConfiguratorTests)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/Users/xistence/Projects/pyramid/pyramid/tests/test_config/test_init.py", line 1172, in test_conflict_set_forbidden_view
        config.commit()
    pyramid.exceptions.ConfigurationConflictError: Conflicting configuration actions
      For: ('view', <class 'pyramid.httpexceptions.HTTPForbidden'>, '', None, 'd41d8cd98f00b204e9800998ecf8427e')
        Line 526 of file /Users/xistence/Projects/pyramid/pyramid/util.py:
            return wrapped(self, *arg, **defaults)
        Line 526 of file /Users/xistence/Projects/pyramid/pyramid/util.py:
            return wrapped(self, *arg, **defaults)
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/Users/xistence/Projects/pyramid/pyramid/tests/test_config/test_init.py", line 1175, in test_conflict_set_forbidden_view
[...]

PR #1976, opened 2015-10-12
Fix python 3.5 support

bertjwregeer:
This fixes Python 3.5 support for Pyramid on the master branch and closes https://github.com/Pylons/pyramid/issues/1973.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,28 @@
 language: python
 sudo: false
 
-env:
-  - TOXENV=py26
-  - TOXENV=py27
-  - TOXENV=py32
-  - TOXENV=py33
-  - TOXENV=py34
-  - TOXENV=pypy
-  - TOXENV=pypy3
-  - TOXENV=py2-docs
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,18 +2,28 @@
...
+        - python: 3.5
+          env: TOXENV=py35
+        - python: pypy
+          env: TOXENV=pypy
+        - python: pypy3
+          env: TOXENV=pypy3
+        - python: 3.5
+          env: TOXENV=py2-cover,py3-cover,coverage
+        - python: 3.5
+          env: TOXENV=pep8
```
mmerickel: Why is the matrix necessary?

bertjwregeer: Without it there is no way to test Python 3.5 since it is not installed by default on Travis. So we need to explicitly declare it.

See @sontek's run where he just added the env directly: https://travis-ci.org/Pylons/pyramid/jobs/84862318#L115

I got this from @dstufft on the Pyramid IRC channel.

PR #1985, opened 2015-10-16
WIP -- Feature: configurable view deriver

bertjwregeer:
This is a work in progress PR, I am going to be pushing changes to it over time, the main goal is to let Travis build this branch every so often.

Besides some tests that weren't fully covered, master merged in without any conflicts, so this actually is me rebasing everything on top of master to allow this change set to more easily be reviewed.

Things left to do:

- [x] Merge with master
- [x] Make tests pass fully, including coverage
- [x] Change `view, default, **kw` to `view, value, **kw`. We want to pull the value out of the `kwargs` automatically and set it as the value, if there is no value in the `kwargs` we pass in the registered default.
- [x] Change `**predicates` to `**view_options`, and split out `**predicates` and `**options`. Options is going to hold on to the `kwargs` for the derivation views, and predicates is going to hold on to the `kwargs` for the predicates, as the name implies.
- [x] auto-call `@wraps_view` when registering the view derivation using `add_view_derivation`. We want to keep @wraps_view as private API until someone says they absolutely need it.
- [ ] Documentation for all of the view derivation stuff (going to need help with this, I am terrible at documentation).
- [x] `outer_derivation = attr_wrapped, predicated_view`. These have to exist, without them the whole view machinery does not work. So we don't allow them to be replaced/changed/or anything.
- [x] `inner_derivation = -renderered_view`, we need to register `rendered_view` as a default, but we don't want to dictate that no-one could potentially insert a view derivation after it.
- [x] Update tests to use `view, value, **kw` rather than `view, default, **kw`

----

- [x] Pyramid 1.7 or can we sneak this into 1.6? :smile:

----

The Future Holds:

- [ ] make `config._add_predicate` register the discriminator twice, just in case someone used the discriminator for the `view predicate`, currently it is renamed to `view options`. While a b/w incompatible change one that @mmerickel and I agree doesn't really hurt. We will only do this if absolutely necessary.
- [ ] Add option name to `add_view_derivation`, so that you can define which `kwarg` you want, rather than basing it strictly on your view derivations name.


bertjwregeer:
@mmerickel I have one last thing left to do, and that is learn how the TopologicalSorter() works, maybe I can summon @mcdonc to help with that.

I am not counting documentation, I'll try and get some simple samples put together that we can use to showcase the view derivations, and we can go from there.

@stevepiercy This changes up the view pipeline and gives it more flexibility. So we'll have to update the documentation, especially that graphic that IIRC you built.

stevepiercy:
@bertjwregeer I have plenty of time to work on the docs and diagram. I won't be much help at deriving narrative from code, though, but I can try.

mmerickel:
I can help with the `TopologicalSorter`.

I'm -1 on putting this in 1.6. We can consider getting a 1.7 alpha out fairly quickly however once we are happy with this feature.

bertjwregeer:
@mmerickel I figured out the `TopologicalSorter` and changed `weighs_less_than` and `weighs_more_than` to `over`/`under`.

bertjwregeer:
I'm a terrible person, hope no-one had pulled that branch down... just rebased on top of master.

mmerickel:
Hmm, I think you should reverse the rebase to avoid modifying original commits by the previous authors.

bertjwregeer:
This was the second rebase I've done. I did one when I started the work to do it on top of the latest master, rather than the outdated branch that existed (since merging master into that made the diff a hellish nightmare to review).

I can go back to that branch, merge master in, and then replay all my new commits on top of it if you'd like.

bertjwregeer:
BRB...

PR #2012, opened 2015-10-20
add pshell --list-shells and default_shell ini options

mmerickel:
I'm malleable on the naming of `default_shell`. I considered `preferred_shells` as well.

### DIFF
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -291,7 +291,7 @@ introduce a new shell by registering an entry point in your setup.py:
 
     setup(
         entry_points={
-            'pyramid.pshell': [
+            'pyramid.pshell_runner': [
               'myshell=my_app:ptpython_shell_factory',
             ],
         },
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -269,32 +269,39 @@ request is configured to generate urls from the host
 
 .. _ipython_or_bpython:
 
-IPython or bpython
+Alternative Shells
 ~~~~~~~~~~~~~~~~~~
-
 If you have `IPython <http://en.wikipedia.org/wiki/IPython>`_ and/or `bpython
 <http://bpython-interpreter.org/>`_ in the interpreter you use to invoke the
-``pshell`` command, ``pshell`` will autodiscover and use the first one found,
-in this order: IPython, bpython, standard Python interpreter. However you could
-specifically invoke your choice with the ``-p choice`` or ``--python-shell
-choice`` option.
+``pshell`` command, ``pshell`` will autodiscover and use the first one found.
+However you could specifically invoke your choice with the ``-p choice`` or
+``--python-shell choice`` option.
 
 .. code-block:: text
 
-   $ $VENV/bin/pshell -p ipython | bpython | python development.ini#MyProject
+   $ $VENV/bin/pshell -p ipython development.ini#MyProject
 
-Alternative Shells
-~~~~~~~~~~~~~~~~~~
+You may use the ``--list-shells`` option to see the available shells.
+
+.. code-block:: text
+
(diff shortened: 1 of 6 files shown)


docs/narr/commandline.rst
```diff
@@ -269,32 +269,39 @@ request is configured to generate urls from the host
...
-   $ $VENV/bin/pshell -p ipython | bpython | python development.ini#MyProject
+   $ $VENV/bin/pshell -p ipython development.ini#MyProject
+
+You may use the ``--list-shells`` option to see the available shells.
+
+.. code-block:: text
+
+   $ $VENV/bin/pshell --list-shells
+   Available shells:
+     bpython  [not available]
```
tseaver: I think it would be better just to omit listing any shell whose factory returns `None` in the default listing.

mmerickel: Most factories should not return `None` because their bindings should depend on their reqs. This is mainly for bpython and ipython and I wanted some way to indicate to the user that things are good, they just need to install bpython.

tseaver: Given that we've gone to the trouble to factor this with entry points, wouldn't it be better to split out `pyramid_bpython` and `pyramid_ipython` packages, each with appropriate dependencies, and registering the relevant entrypoint?  Then, anybody who wants one of them just installs that (mostly-empty) package, just like any other shell.

mmerickel: Probably.

mmerickel: Conversely pyramid ships a cherrypy wsgi entry point for no apparent reason.

tseaver:
LGTM

bertjwregeer:
:+1:

mmerickel:
@sontek please provide some thoughts before I merge as it's building on your work.

invisibleroads:
Great, now we can choose the shell for pshell.

pyramid/scripts/pshell.py
```diff
@@ -77,6 +96,8 @@ def pshell_file_config(self, filename):
         for k, v in items:
             if k == 'setup':
                 self.setup = v
+            elif k == 'default_shell':
```
sontek: would `preferred_shells` be a better name for this in the config?

sontek:
@mmerickel I love it =)   I wanted to rip the shells out but didn't think we'd want to break backwards compatibility.   Only thing I saw was the config `default_shell` as a list of preferred shells seemed like a weird name

pyramid/scripts/pshell.py
mmerickel: I started with that and changed to `default_shell` but I can change it back. I don't really care. Naming is hard. Next someone is going to ask to load the option from an environment variable anyway.

mmerickel:
This is probably too much but I realized that none of the shells themselves depend on pyramid so maybe the entrypoint doesn't need to be called something pyramid-specific. For example my apps tend to have their own `shell` command that loads the database etc outside of pyramid and it'd be nice to use these shell implementations there. We could go a step further in the future and pull the whole "shell runner" framework out of pyramid as a dependency.

docs/narr/commandline.rst
```diff
@@ -263,38 +261,40 @@ request is configured to generate urls from the host
...
 
 .. code-block:: python
 
     setup(
-        entry_points = """\
-            [pyramid.pshell]
-            myshell=my_app:ptpython_shell_factory
-        """
+        entry_points={
+            'pyramid.pshell': [
```
dakra: That has to be `pyramid.pshell_runner`. Everything else looks good to me :+1:

mmerickel: Nice catch, thank you.

bertjwregeer:
:shipit:

PR #2013, opened 2015-10-20
remove default cache busters

mmerickel:
This is a first cut at removing the default cache busters.

The next step is to improve the documentation inside the "Customizing the Cache Buster" section of `narr/assets.rst` to include some explanation of how to write your own and why. Initial bullet points are:

- Pyramid does not control all of the urls in your app (for example css). You may want to rewrite these another way (within your asset pipeline) or even via middleware.
- You want to integrate with your asset pipeline.
- Common use case is to use an externally defined `manifest.json` that maps assets to busted slugs:
  
  ```json
  {
      "default.css": "default-11223344.css",
  }
  ```

Are there any other common ways to cache bust that we should touch on other than with a manifest?

One hiccup is that it depends on the asset pipeline what the actual file on disk is named. Thus on the matching side of the cache buster it depends on what you want when you map the external url back to the asset on disk.

Thoughts?

### DIFF
--- a/pyramid/static.py
+++ b/pyramid/static.py
@@ -19,7 +19,10 @@ from pkg_resources import (
 
 from repoze.lru import lru_cache
 
-from pyramid.asset import resolve_asset_spec
+from pyramid.asset import (
+    abspath_from_asset_spec,
+    resolve_asset_spec,
+)
 
 from pyramid.compat import text_
 
@@ -211,7 +214,11 @@ class ManifestCacheBuster(object):
     uses a supplied manifest file to map an asset path to a cache-busted
     version of the path.
 
-    The file is expected to conform to the following simple JSON format:
+    The ``manifest_spec`` can be an absolute path or a :term:`asset spec`
+    pointing to a package-relative file.
+
+    The manifest file is expected to conform to the following simple JSON
+    format:
 
     .. code-block:: json
 
@@ -222,7 +229,7 @@ class ManifestCacheBuster(object):
 
     Specifically, it is a JSON-serialized dictionary where the keys are the
     source asset paths used in calls to
-    :meth:`~pyramid.request.Request.static_url. For example::
+    :meth:`~pyramid.request.Request.static_url`. For example::
 
     .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


tseaver:
[Fails the `pep8` check](https://travis-ci.org/Pylons/pyramid/jobs/86332790).

mmerickel:
@tseaver Thanks. Any other comments or use cases you'd like to see covered? :-)

mmerickel:
@dstufft Would you mind giving me some thoughts before I merge this PR? I added a `ManifestCacheBuster` to pyramid core and I want to know if you'd use it for warehouse.

dstufft:
Yes I would. It looks very similar to the [WarehouseCacheBuster](https://github.com/pypa/warehouse/blob/master/warehouse/utils/static.py) I have already. It looks like it'd be trivial for me to switch to this and it'd let me delete some code from Warehouse.

dstufft:
I guess the main difference between the two is mine accepts an asset specifier for the manifest and yours requires a real file path. I think the asset specifier is better since that's how you mount a static directory too and the ``manifest.json`` typically lives alongside your static files.

pyramid/static.py
```diff
@@ -284,3 +204,83 @@ def __init__(self, token, param='x'):
...
+        self._mtime = None
+        if not reload:
+            self._manifest = self.parse_manifest()
+
+    def parse_manifest(self):
+        """
+        Return a mapping parsed from the ``manifest_path``.
+
+        Subclasses may override this method to use something other than
+        ``json.loads``.
```
dstufft: This method could also be used not only to switch from something other than ``json.loads()`` but also to coerce the underlying manifest into a dictionary in the correct format for ``ManifestCacheBuster``. Might be worth documenting that as well?

mmerickel: Sure I can improve the subclassing note in the docstring.

mmerickel:
Good idea, I can probably also support an asset spec.

mmerickel:
Alright I added asset spec support. Improves on warehouse slightly by supporting relative specs (infer package name from caller).

dstufft:
:ship:

bertjwregeer:
:+1:

chrisrossi:
You guys rock.

mmerickel:
Sorry I force pushed this branch **after** merging it to master... the PR is now fairly messed up. The actual changes are done in https://github.com/mmerickel/pyramid/commit/3a41196208c7fd78cfb177bb5105c6a702436b78

PR #2021, opened 2015-10-21
configurable view deriver

bertjwregeer:
This is a work in progress PR, this branch we want to use to continue pushing changes until this feature is complete and ready to go.

Things left to do:

- [x] Merge master into branch and then continue development
- [x] Make tests pass fully, including coverage
- [x] Change `view, default, **kw` to `view, value, **kw`. We want to pull the value out of the `kwargs` automatically and set it as the value, if there is no value in the `kwargs` we pass in the registered default.
- [x] Change `**predicates` to `**view_options`, and split out `**predicates` and `**options`. Options is going to hold on to the `kwargs` for the derivation views, and predicates is going to hold on to the `kwargs` for the predicates, as the name implies.
- [x] auto-call `@wraps_view` when registering the view derivation using `add_view_derivation`. We want to keep @wraps_view as private API until someone says they absolutely need it.
- [x] `outer_derivation = attr_wrapped, predicated_view`. These have to exist, without them the whole view machinery does not work. So we don't allow them to be replaced/changed/or anything.
- [x] `inner_derivation = -renderered_view`, we need to register `rendered_view` as a default, but we don't want to dictate that no-one could potentially insert a view derivation after it.
- [x] Update tests to use `view, value, **kw` rather than `view, default, **kw`
- [x] Documentation for all of the view derivation stuff
- [x] Update [request processing diagram](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/router.html)

----

The Future Possibly Holds:

- [x] make `config._add_predicate` register the discriminator twice, just in case someone used the discriminator for the `view predicate`, currently it is renamed to `view options`. While a b/w incompatible change one that @mmerickel and I agree doesn't really hurt. We will only do this if absolutely necessary.
- [x] Add option name to `add_view_derivation`, so that you can define which `kwarg` you want, rather than basing it strictly on your view derivations name.


bertjwregeer:
@latteier I just noticed that you have not signed CONTRIBUTORS.txt, could you please do so with a PR on this branch or `master`? Thank you!

latteier:
@bertjwregeer thanks for the reminder. I can't believe that I hadn't done this already. Here's a PR: https://github.com/Pylons/pyramid/pull/2030

bertjwregeer:
Cheers mate :-)

bertjwregeer:
Did some gardening, merged master into this PR, and resolved conflicts.

mmerickel:
1. Currently the API for a view deriver is `(view, value, **options) -> view`. This should be changed to `(view, options) -> view`.
2. `config.add_view_derivation` should not have a view option associated with it by default (many view derivations including almost all of the default ones do not have their own 1-1 derivation-option combinations).
3. We should add a `config.add_view_option` API for conflict detection between options and predicates which developers can use to declare intent to use an option.
4. We need to decide how to handle the `check_csrf` predicate as it is a major reason we're even talking about view derivations.
5. We need to decide what the appropriate behavior is for a predicate and view option conflicting. I'm worried that if you declare a `check_csrf` option, the predicate is still in effect! Is there a way to turn it off?

mmerickel:
As I sat down to work on `config.add_view_option` I'm really struggling with the capabilities of this API.

First, a state of the union for this PR. We've implemented `config.add_view_derivation` with an api of `deriver(view, info) -> view`. Pyramid's view pipeline has been re-written to this idea with default derivations for the renderer, mapper, decorator, auth checker, wrapper and http_cache.

In order to make this work, view derivations have access to most options passed to `config.add_view` in order to decide what to do. For example, the renderer derivation needs the `renderer` option and the wrapper derivation needs the `wrapper` and `name` (view name) options.

We've also made this extensible such that any extra options currently passed to `config.add_view` will be passed to the derivations. Why? So ... we have a single namespace for all options passed to `config.add_view` because they must all fit in as kwargs. Most of these are defined by certain apis. For example:

- `name`, `context` are core functionality. These options must be specified to every invocation of `add_view`.
- `request_method`, `match_param` are predicates. These options are also per-view and have no effect when not specified.
- `renderer`, `mapper`, `permission` are also core functionality. These options can be per-view or global and are defined by their own apis like `config.set_default_permission` or `config.set_view_mapper`.
- derivation options, such as `require_csrf=True` which are new and are options passed to control view derivations.

All together this entire set is folded into a single namespace named `view options` for the view. We are exposing all of these options to the view derivations in order for them to do their job. They can pick and choose what they care about from the set. For example a `require_csrf` view derivation may care about the `require_csrf` option as well as the `request_method` option. If the view is only registered for `GET` / `HEAD` then it stands to reason the derivation could turn itself into a noop at config-time instead of doing that check per-request.

Cool! So what do we need?

Before this feature the only way to extend the view options namespace was via predicates. If you called `config.add_view(..., foo=True)` without calling `config.add_view_predicate('foo', ...)` you would see a `ConfigurationError` with an unrecognized option `foo`. **Right now** this PR breaks that behavior by not erroring. The `foo` option is passed to the view derivation.

So what do we want to do about this? We want a way for the derivation to declare that it uses certain options. First thought: `config.add_view_derivation('csrf_checker', check_csrf, options=('require_csrf', 'request_method'))`. Originally we had proposed `config.add_view_option('require_csrf', default=True)`. I'm moving away from this idea for 2 reasons:

1. The view option is no longer tied to the derivation. Thus if the derivation is overridden the option still remains registered.
2. I no longer see a reason for default values passed to derivations. These can be solved by simply passing the default value directly to the view derivation. For example, `config.add_view_derivation('csrf_checker', check_csrf(True), options=('require_csrf',))`. The `check_csrf` could return a derivation factory that defaults to on unless `config.add_view(..., require_csrf=False)` is passed to override it. For this reason I just don't see a need to framework the default options at this time.

Sorry this is long but I wanted to write down where this cool feature is at and what I'm going to do next: add an `options` argument to `add_view_derivation`.

mmerickel:
I've renamed the api to view derivers instead of derivations. Feel free to bikeshed this but the callable we are registering is `(view, info) -> view` so a "deriver of a new view" made sense to me whereas the end result is a derivation.

mmerickel:
latest changes document the ``IViewDeriver`` interface and add the ability for derivers to expose a list of options. These options extend what can be passed to ``config.add_view`` to allow user-defined options outside of the custom predicate names. An example follows:

```python
def require_csrf_view(view, info):
    if info.options.get('require_csrf', False):
        def protected_view(context, request):
            check_csrf(request, raises=True)
            return view(context, request)
    return view

config.add_view_deriver('require_csrf', require_csrf_view)
config.add_view(..., require_csrf=True)
```

stevepiercy:
Ping @mmerickel @bertjwregeer. Added "view deriver" to the request processing diagram. If it looks good, I'll push the .png, .svg, and .graffle to this branch.

image

bertjwregeer:
There is a view deriver egress too... the view derive wraps the view after all.

stevepiercy:
Revised diagram.

- added "view deriver egress"
- repositioned "view deriver ingress" and "view deriver egress"

We're not certain about the response adapter position.

pyramid_request_processing

mmerickel:
As I said in chat the other day this PR doesn't actually change this diagram at all. Response adapters are still above decorators. If you'd like to rename the section to "view deriver" instead of "view" in the legend, maybe that is the most appropriate action.

stevepiercy:
- reverted all changes
- updated legend
- fixed width of one box

Request Processing Diagram

mmerickel:
LGTM @stevepiercy

mmerickel:
Added a first cut at some documentation tonight and I came across 2 issues that I want to discuss.

1) I switched the signature to `add_view_deriver(deriver, name)` to make `deriver` first. This more closely mirrors `add_tween(tween)`. The difference between these apis is an issue I want to discuss further because it has been bugging me for some time now - mostly regarding tweens. There is currently a restriction on tweens that they be an importable object and their name is **always** the Python dotted name. I think this is done to ensure the tween has a consistent name but it also limits the api in some ways. For example, the tween can't be a product of some other computation. It must come directly from the top-level code that exposes it. It can't be used twice with different options in the pipeline by specifying different names and possibly options. Finally, it can only receive options from the `registry` and `settings`.

At the end of the day these restrictions exist, and view derivers are almost identical in purpose to tweens, just with a different scope. As such I'd like to keep their APIs as similar as possible - whether that means limiting view derivers or fixing tweens... well that's why I'm bringing it up because I'm not sure.

2) In the topological sort order `FIRST` is the `mapped_view` and `LAST` is the `attr_wrapped_view`. This is weird because "over" means closer to the view callable instead of "wrapping". This is different from tweens where "over" is closer to `INGRESS` and further from `MAIN`. I think I'd like to switch the order such that "over" is closer to the `INGRESS` deriver.

mmerickel:
I just took care of point 2.... "over" now means closer to ingress, which is the same terminology used by tweens.

bertjwregeer:
While I agree that the tween API has its limitations, I don't think nor want that to hold up merging this new API in its current form just because the two API's are dissimilar. I do think we should allow `maybe_dotted` for the view derivers, but passing in a class or function or anything along those lines is fine as well. I don't think limiting view derivers is the right thing to do.

We should start another PR to discuss how to fix the tween issue.

stevepiercy:
Docs and docstrings LGTM (subject to successful Travis-CI builds)!

bertjwregeer:
Opened a new ticket for the Tween configuration. I really don't want it holding up these changes because I am already using view derivations using this branch for a semi-production app :P

mmerickel:
Okay my last issue is some UX / book-keeping stuff.

Tweens do a bit of work to ensure that over/under are strings and also provide `pyramid.tweens.INGRESS`, `pyramid.tweens.EXCVIEW` and `pyramid.tweens.MAIN` APIs for easily adding things over/under. `config.add_tween` also does some work to avoid adding things over ingress and under main.

Currently it's basically wrong to ever call `config.add_view_deriver` without specifying BOTH over and under arguments. Without both you have very minimal guarantees on where the deriver will appear. The only case that is handled properly is the default which will place the tween over rendered_view and under decorator_view.

The main thing that worries me is that if you specify `under='decorated_view'` there is no guarantee that the deriver will appear before `rendered_view`. This will be a bug if your deriver is depending on the response being a response object as the api changes after egress of rendered_view.

Anyway I don't really have any action items about this but I want to bring it up as an easy way to shoot yourself in the foot. One option is to always specify over/under if they are `None`, forcing a user to override them. Envision `def add_view_deriver(config, name, over='rendered_view', under='decorated_view')`. This leaves the defaults in a good spot and at least it will always be over rendered_view unless you change it. The main foot-gun here is that you will get a cyclic dependency error if you specify `under='rendered_view'` without overriding `over='rendered_view'` to something - and there is actually no API right now to define that something because we don't have MAIN or INGRESS so we'd have to define them.

I swear this is my last issue.

mmerickel:
I don't really have a problem leaving things as-is but here's what I'm talking about. If a deriver does not define both an over and an under then the deriver is underspecified and could be placed into unexpected spots in the view pipeline. For example if I say `under=authdebug_view` then the deriver could end up being sorted under `rendered_view` which has a different contract for the return value `(context, request) -> any_python_obj` versus `(context, request) -> response`. While the deriver may sort properly in the developers app it may not work in someone else's app depending on the other derivers added.

The feature I've been considering is simply a sanity check step after registering all derivers that would ensure that every deriver eventually maps back to one of the builtins in both `over` and `under` directions. It would be a breaking change down the road to require derivers to be fully specified which is why I was thinking about it now. Note that tweens currently suffer from the problem I'm describing as well.

bertjwregeer:
I vote for making users fully specify their over/under and anything else raises an error. None of this wishy washy type stuff that leads to potential errors down the road.

bertjwregeer:
:+1: I am very happy with this branch!

mmerickel:
I'm also happy with it. The only other thing I can think of changing atm is some checks to force preserving the order of the builtins but I don't think it's a big deal.

bertjwregeer:
Thank you @mmerickel for all your hard work on making this a succes :-)

PR #2030, opened 2015-10-22
Sign contributor list.

latteier:
I forgot to do this. I was reminded by @bertjwregeer


bertjwregeer:
Thank you! We'll need this in #2021.

PR #2044, opened 2015-10-27
pserve: track files with SyntaxError and skip reload

### LINKED ISSUE #681: Development mode server kills itself on syntax errors
Hi there,

I've noticed an annoying feature of the development mode of pyramid, that is, it's server kills itself if it encounters a syntax error.

While that is certainly safe, what I think it should really do is just try again after the next file change (that usually fixes the syntax error....).

domenkozar:
Current implementation tries to compile changed file and keeps a list (would be better to use a set) of syntax error files. Until that list exists, it won't reload.

There is a possible race condition if one would save a file with no syntax error and then quickly another file with a syntax error. Hard to solve that with current implementation.

The downside of this approach is, it prints error to stdout all the time while syntaxerror is present.

Fixes #681

### DIFF
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -14,6 +14,7 @@ import errno
 import logging
 import optparse
 import os
+import py_compile
 import re
 import subprocess
 import sys
@@ -820,7 +821,7 @@ class Monitor(object): # pragma: no cover
         if %errorlevel% == 3 goto repeat
 
     or run a monitoring process in Python (``pserve --reload`` does
-    this).  
+    this).
 
     Use the ``watch_file(filename)`` function to cause a reload/restart for
     other non-Python files (e.g., configuration files).  If you have
@@ -843,6 +844,7 @@ class Monitor(object): # pragma: no cover
         self.poll_interval = poll_interval
         self.extra_files = list(self.global_extra_files)
         self.instances.append(self)
+        self.syntax_error_files = []
         self.file_callbacks = list(self.global_file_callbacks)
 
     def _exit(self):
@@ -887,11 +889,28 @@ class Monitor(object): # pragma: no cover
                 continue
             if filename.endswith('.pyc') and os.path.exists(filename[:-1]):
                 mtime = max(os.stat(filename[:-1]).st_mtime, mtime)
+                pyc = True
+            else:
+                pyc = False
             if filename not in self.module_mtimes:
                 self.module_mtimes[filename] = mtime
             elif self.module_mtimes[filename] < mtime:
-                print("%s changed; reloading..." % filename)
-                return False
+                if pyc:
+                    filename = filename[:-1]
+                # check if a file has syntax errors. If so, track it until it's fixed
+                for filename_promise in self.syntax_error_files + [filename]:
+                    try:
+                        py_compile.compile(filename_promise, doraise=True)
+                    except py_compile.PyCompileError:
+                        print("%s has a SyntaxError; NOT reloading." % filename_promise)
+                        if filename_promise not in self.syntax_error_files:
+                            self.syntax_error_files.append(filename_promise)
+                    else:
+                        if filename_promise in self.syntax_error_files:
+                            self.syntax_error_files.remove(filename_promise)
+                if not self.syntax_error_files:
+                    print("%s changed; reloading..." % filename)
+                    return False
         return True
 
     def watch_file(self, cls, filename):


bertjwregeer:
Ohhhh, I like it.

pyramid/scripts/pserve.py
```diff
@@ -887,11 +889,28 @@ def check_reload(self):
...
-                print("%s changed; reloading..." % filename)
-                return False
+                if pyc:
+                    filename = filename[:-1]
+                # check if a file has syntax errors. If so, track it until it's fixed
+                for filename_promise in self.syntax_error_files + [filename]:
+                    try:
+                        py_compile.compile(filename_promise, doraise=True)
+                    except py_compile.PyCompileError:
+                        print("%s has a SyntaxError; NOT reloading." % filename_promise)
```
jessedhillon: At the head of this file, logging is set up. Shouldn't we use that logger instead of `print`?

jessedhillon: Additionally, if we move this line down into 907, then we will only get this message the first time a `SyntaxError` in a file prevents us from reloading.

mmerickel: > At the head of this file, logging is set up. Shouldn't we use that logger instead of print?

`print` is wrong but so is logging, we should be using the `self.out` assigned to the command object or adding a `self.out_err` of sorts to use stderr.

> Additionally, if we move this line down into 907, then we will only get this message the first time a SyntaxError in a file prevents us from reloading.

I think it's fine to try to print things out only once AS LONG AS there isn't much noise between the errors. It's better to be verbose about the state of the world versus printing an error once and then there's a bunch of stuff on the console from other changes and you forget about that one file. Then you wonder why the thing isn't reloading. For this reason I'm leaning toward showing all the errors all the time.

jessedhillon: > `print` is wrong but so is logging,

Why is logging wrong? Logging is user-configurable and allows you to agnostically write to the user's preferred output. As far as informational versus erroneous conditions, that why log levels exist -- you can use `info()` and `error()`.

> For this reason I'm leaning toward showing all the errors all the time.

I think if you do it this way, you're likely to inure the user to important messages. Additionally, what other output would the server generate if it's stopped due to a `SyntaxError`? More messages from other errors.

The key point is that the software should only be generating output in response to user-concerning events. The fact that a loop completed another cycle is not of concern to the user. If it prints out a list of syntactically-incorrect files every N seconds, that's leaking a fact about the server which I, as a user, don't care about.

Additionally, there may be error messages on screen which I *do* care about and need to track *e.g.* stack traces from previous unhandled exceptions. Why should those get bumped out of history by messages being generated on an interval which I don't control?

Suppose a project has `x.py` and `y.py` in it. As a dev,

- If I introduce an error into `x.py`, I would like to see:
  ```
  x.py has an error
  ```
  That error should be printed out once.

- If I then introduce an error into `y.py`, then I would like to see:
  ```
  x.py has an error
  y.py has an error
  ```
  In this case I am learning that my modification to `y.py` triggered a reload cycle, **which I expected**, and additionally `x.py` remains unfixed. These are both germane to my user-initiated action, and therefore I care about seeing the first message re-printed.

- If I then fix `x.py` and save, this triggers another reload, which again is expected. I therefore would expect to see the current list of erroneous files (only `y.py` now)

In short, bothering the user by default feels like a poorly thought out process. The user is liable to feel that they are bearing the cost of the programmer's failure to keep count. I can't think of single situation where someone would see a wall of 
```
x.py has a syntax error
x.py has a syntax error
x.py has a syntax error
x.py has a syntax error
x.py has a syntax error
x.py has a syntax error
x.py has a syntax error
x.py has a syntax error
...
```
and not find that to be annoying.

bertjwregeer: > Why is logging wrong? Logging is user-configurable and allows you to agnostically write to the user's preferred output. As far as informational versus erroneous conditions, that why log levels exist -- you can use info() and error().

The logging is for the server itself, not for the runner. This is information the runner is providing, and is not at all related to the application/WSGI server itself.

I think that logging that with the logger is wrong, I don't expect output from my runner to show up in my logger/logging.

This is also related to a very small part of the runner, the `--reload` feature, which IMHO should not be used in anything but development.

mmerickel: > Why is logging wrong? Logging is user-configurable and allows you to agnostically write to the user's preferred output. As far as informational versus erroneous conditions, that why log levels exist -- you can use info() and error().

Sorry but when I say "wrong" I mean "this isn't the current convention in Pyramid". This is how all of the `p*` scripts in pyramid currently work. This issue (let alone line note) are not the place to discuss changing that. Getting this pr accepted is about conforming to the current standard. If you want to initiate/discuss a refactor of all the scripts you can open a separate issue/PR about that.

As far as the verbosity goes it would help if you interspersed in your examples some of the other messages output by pserve such as `%s changed, reloading ...` and others to get a feel for what you're talking about in your last example. I agree with most of your points however. I haven't tried to run the code yet, and your output isn't from real code so it's a little hard to envision your proposed change. We could easily be talking past each other or agreeing.

jessedhillon: Thanks for clarifying the logging point, I understand and agree re logging being for the application and framework, not for the runner.

Regarding the verbosity, the basic point I am making can be summed as this: only output things which are in response to events I've generate, or which I am expecting.

domenkozar: I think we should be printing SyntaxErrors on each reload. Not-so-small applications will quickly configure logging for their app.

Consider a following common use case:

- Developer changes file `foo.py` and makes a SyntaxError
- pserve prints the error message and doesn't reload
- Developer goes to the browser and reloads to see the new code
- Browsers makes requests to the server, prints out lots of logging about them
- Developer is confused and checks the console, but due to a bunch of logging, syntax error is hidden in history

TL;DR: printing out the error on each reload is annoying, but so it making a SyntaxError.

mmerickel: Right, I think it's important that the user is not confused about whether a particular error in the scrollback is fixed yet or not. If they see the error then some reloading text because they saved then if the file is still broken they should see the error again.

domenkozar: So, do we have any action-ables here?

mmerickel: We need to add a test and tweak the code to use the mockable `out` method instead of `print`.

mmerickel:
Ok I ran this code myself and see what @domenkozar was complaining about. Basically the error gets spammed out to the console every cycle of the poll (1s by default). It'd be really nice if we could fix that to only dump out when the file is saved. The suggestion by @jessedhillon doesn't quite work either as there is no updated output when you re-save the file because it is already in the list.

mmerickel:
I'm trying right now to come up with something that behaves more like I'd expect.

mmerickel:
I submitted #2049 against this branch to clean up a few things.

As I got into it I decided the print statements are fine. We currently have no tests for the monitor and it's always user-facing.

mmerickel:
@domenkozar any updates / thoughts? Do you want to merge #2049?

mmerickel:
@domenkozar Can you please review this? I'm about to cut 1.6b1 and would like to include this. I don't plan to do another beta release after this unless there are issues, given how long the rest of the tree has been baking. I'd like to have a final out by December 1st at the very latest.

domenkozar:
OK, reviewing.

Kilte:
Nice! What about import errors?

domenkozar:
We could use AST to parse themodified file and check for import errors, but this is going to get complicated :)

mmerickel:
The best answer I can come up with is to establish a multiprocessing protocol where the parent forks a child process B which calls `get_app()`  and then notifies the parent that it worked. The parent then kills the child process A and notifies the child process B to start the server.

Alternatively a poor man's implementation could just call `get_app` in the parent and if it works then reload. There's a fair amount of overhead in calling `get_app` twice per change though.

davisagli:
@domenkozar AST will just tell you about syntax errors; you'd have to import the modules yourself and that could have side effects. So yeah I think actually loading the app is the only way to know for sure if loading the app will work.

Do we really need to keep the old code running until the new code succeeds though? I think the important thing is catching any errors during configuration and keeping the monitor running so it can try again after you fix the problem; we don't necessarily need to keep the old code running (in fact probably better if we don't, to avoid the situation where someone spends an hour wondering why something doesn't work because they assume they're now using the new code but they aren't).

mmerickel:
@davisagli I like that approach, it's a great idea. You're right that the main thing we need is to just keep the monitor running and only try to start the app again after a new change is detected.

davisagli:
For bonus points, when the app's broken start another app that displays a big frowny face.

davisagli:
Or more helpfully, displays the error output.

bertjwregeer:
During import errors it would just be nice if the monitor kept running. I don't really care about the WSGI server itself.

domenkozar:
@davisagli you can extract imports and dynamically import them and test if modules are available.

bertjwregeer:
All of my import errors have actually been due to `maybe_dotted`.

I pass strings to `view_config` for things like `context`, and those get `maybe_dotted` when the configurator runs its scan.

It's at that point that it blows up. If you have a bad import statement somewhere in the code, that tends to get caught without issues already.

Issue #1656, opened 2015-05-05
Make BaseCookieSessionFactory require timeout to be a number


kengruven:
In pyramid/session.py, BaseCookieSessionFactory (and CookieSession) accept a "timeout" parameter.  This is stored in self._timeout, and later compared against a number (now - renewed).

With Python 3, comparing a number against a string is an immediate TypeError.  With Python 2, it's not an error (but probably not a meaningful result).

Thus, if someone were to run a Pyramid web server on Python 2, and passed the "timeout" parameter straight from their configuration file (as a string, unparsed), the session timeout would silently fail -- until they upgraded to Python 3, at which point everything would start failing.

One simple improvement might be to change:

    _timeout = timeout

to:

    _timeout = int(timeout)

(at https://github.com/Pylons/pyramid/blob/master/pyramid/session.py#L247) which would cause it to automatically upgrade a string like "1200" to the int 1200, or at least fail immediately (on all versions of Python) if you tried passing it something crazy.

tseaver:
I'd be happy to review a PR which applied this change:  it would need new tests for both valid and invalid string values.

RaHus:
The first two references above are a mistakes, please ignore. Sorry for the noise.

Should this change/feature also be mentioned on the param documentation?

stevepiercy:
> Should this change/feature also be mentioned on the param documentation?

@RaHus I don't see how this change would affect the intended behavior (IOW the docs are right, but the code has a bug), so I don't now whether adding something to its documentation would be appropriate.

But it would definitely be appropriate to add a note to [CHANGES.txt](https://github.com/Pylons/pyramid/blob/master/CHANGES.txt#L234) under "Bug Fixes".

Thank you for your diligence!

RaHus:
@stevepiercy I agree, i will add the relevant entry under "Bug Fixes" after checking if ``reissue_time`` and ``max_age`` are also affected.

Thanks for the great software!

bertjwregeer:
This has been merged in https://github.com/Pylons/pyramid/pull/2117. Please re-open if this is not the case.

PR #2045, opened 2015-10-28
Convert timeout argument to int values when applicable

RaHus:
In response to #1656.

This is my first contribution so i am not sure where to base the changes, currently on top of 1.6-branch.
Should i edit the parameter documentation also?


RaHus:
looks like i should have probably based on master.

stevepiercy:
@RaHus for the record, basing a PR on master is preferred, and if accepted, then we backport it to other releases when appropriate.

PR #2049, opened 2015-10-28
avoid spamming by outputting state only once when a file changes

mmerickel:
This reduces the output to only when a file changes. However, I'm able to get it into a weird state where it keeps spamming and I haven't figured out the fix for that yet.

1) start `pserve --reload`
2) save a syntax error to file A.py
3) save file B.py with no changes
4) you will see a continued output of "B.py changed ..." on the console.

It doesn't happen every time but it's somewhat consistent for me (especially when B.py is pserve.py).

### DIFF
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -18,6 +18,7 @@ import py_compile
 import re
 import subprocess
 import sys
+import tempfile
 import textwrap
 import threading
 import time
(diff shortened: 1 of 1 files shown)


bertjwregeer:
:+1:

pyramid/scripts/pserve.py
```diff
@@ -892,25 +893,34 @@ def check_reload(self):
...
+            return False
+        return True
+
+    def check_syntax(self, filename):
+        # check if a file has syntax errors.
+        # If so, track it until it's fixed.
+        try:
+            py_compile.compile(filename, doraise=True)
+        except py_compile.PyCompileError:
+            print("%s has a SyntaxError; NOT reloading." % filename)
```
domenkozar: I'd rather print the message for all files in syntax_error_files here, just to be explicit what is blocking the reload.

domenkozar:
One nitpick, otherwise :+1:

pyramid/scripts/pserve.py
mmerickel: Okay cool, I'll mess with that for a few and see how it feels. The reason I like it the way it is right now is that you only get notifications on the files that actually changed which I think it's somewhat intuitive.

domenkozar: But the information you're looking for is what files to fix to get reloader to work again

mmerickel:
Ok new version prints out the syntax error for the new file as well as the names of all the files that have errors. Printing out all the files really doesn't feel quite right but it's not a big deal since most people should only have one or two files with errors at a time.

mmerickel:
I created a temp file that py_compile uses to avoid changing the actual pyc file. It seems to have fixed the infinite loop issues.

domenkozar:
Shipit

PR #2050, opened 2015-10-28
Convert timeout argument to int values when applicable

RaHus:
In response to #1656.

Reapplied the changes on top of master.

### DIFF
--- a/pyramid/tests/test_authentication.py
+++ b/pyramid/tests/test_authentication.py
@@ -600,6 +600,15 @@ class TestAuthTktCookieHelper(unittest.TestCase):
         cookies.load(cookie)
         return cookies.get('auth_tkt')
 
+    def test_init_cookie_str_reissue_invalid(self):
+        self.assertRaises(ValueError, self._makeOne, 'secret', reissue_time='invalid value')
+
+    def test_init_cookie_str_timeout_invalid(self):
+        self.assertRaises(ValueError, self._makeOne, 'secret', timeout='invalid value')
+
+    def test_init_cookie_str_max_age_invalid(self):
+        self.assertRaises(ValueError, self._makeOne, 'secret', max_age='invalid value')
+
     def test_identify_nocookie(self):
         helper = self._makeOne('secret')
         request = self._makeRequest()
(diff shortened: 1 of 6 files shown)


stevepiercy:
@RaHus as an aside, I recently added [contributing.md](https://github.com/Pylons/pyramid/blob/master/contributing.md), which appears when [creating a new issue](https://github.com/Pylons/pyramid/issues/new) on GitHub. It's intended for new contributors to Pyramid.

1. Did you notice it?
2. If so, do you have suggestions for improvement?
3. If not, where did you find how to do a proper PR? Yours is the first one in a long time that includes a signed `CONTRIBUTORS.txt`.

Thanks.

RaHus:
@stevepiercy yeah i did notice it and it did include helpful information, thanks :)  
A possible improvement could be a list with the long lived branches of the project and a small description on how they fit in the pyramid git workflow.

mmerickel:
Can I convince you to fix `timeout` on the `AuthTktAuthenticationPolicy`[1] as well?

Have you investigated whether this is also an issue with the other parameters (reissue_time and max_age)?

[1] https://github.com/Pylons/pyramid/blob/master/pyramid/authentication.py#L858

RaHus:
@mmerickel Sure you can, i am glad i can help :).

It looks like [``reissue_time``](https://github.com/Pylons/pyramid/blob/master/pyramid/session.py#L31) is affected the same as ``timeout`` but i didn't have time to test it properly.

Next chance i get i'll investigate further and include the changes in this PR if needed.

bertjwregeer:
`max_age` is `int()`ed in WebOb. Doing it sooner would probably help people figure out what is going on sooner rather than later.

mmerickel:
@RaHus I'd appreciate it very much, thank you. As @bertjwregeer said, `max_age` would be good to check as well as it isn't checked until a request occurs as well.

RaHus:
``UnencryptedCookieSessionFactoryConfig`` doesn't support the reissue_time parameter, so i followed a previous authors example and copy/pasted the tests for the subclasses that support the ``reissue_time`` parameter. Since ``UnencryptedCookieSessionFactoryConfig`` is being deprecated and is alone in not supporting ``reissue_time`` maybe its better to keep the reissue tests Shared and DRY and instead override the problematic tests in ``TestUnencryptedCookieSession``. But that means future authors of reissue tests would need to maintain this list up to date, so i am not decided if that is indeed a better solution.

RaHus:
TravisCI fails due to missing ordereddict in python2.6 which is required by WebTest. see #2071

tseaver:
The failure is due to a WebTest [packaging bug](https://github.com/Pylons/webtest/issues/143).

tseaver:
The wheel is gone:  I restated the failed 2.6 test, and it passed.

RaHus:
I think i have included all the items that where discussed here. Any comments on the PR so far or any other item/changes/corrections i should consider? I would love to finalize this and pick another issue to work on.

pyramid/tests/test_authentication.py
```diff
@@ -758,6 +767,12 @@ def test_identify_cookie_timed_out(self):
         result = helper.identify(request)
         self.assertEqual(result, None)
 
+    def test_identify_cookie_str_timeout(self):
```
bertjwregeer: How does this test the timeout of the cookie? A bogus `auth_tkt` is already invalid and will always "timeout".

pyramid/tests/test_authentication.py
```diff
@@ -758,6 +767,12 @@ def test_identify_cookie_timed_out(self):
...
+        helper = self._makeOne('secret', timeout='1')
```
bertjwregeer: If you just want to test that `timeout` can be a `str`, then do a try/except block with a self.assertTrue(false) in the except block. So if it raises a ValueError then the test fails.

RaHus:
The rationale was to make sure it would behave the same with a valid str and int mostly due to me being unfamiliar with the code, but i see how this is needless. Prior to that i was just doing:
```python
def test_init_cookie_str_timeout(self):
    helper = self._makeOne('secret', timeout='1')
    self.assertTrue(helper.timeout == 1)
```
I followed the same rationale for other valid string parameter tests which i will change once someone confirms that the above (just check that the conversion to int is done) is enough

bertjwregeer:
Yes, what you put down is enough. If you make that change I'll be happy to +1 it.

bertjwregeer:
Thank you @RaHus, I've pulled the changes and made the modifications. Please see https://github.com/Pylons/pyramid/pull/2117

Issue #2071, opened 2015-11-03
Tox tests for python 2.6 fail due to WebTest requiring OrderedDict


RaHus:
OrderedDict was included in python 2.7 and WebTest requires it, so tests in python 2.6 fail with ImportError. We should include a backported version of OrderedDict in setup.tests_require when py_version is 2.6

stevepiercy:
@RaHus thanks for the explanation, and I agree. FWIW, I noticed TravisCI errors that recently cropped up on the branches master and 1.6-branch:
https://travis-ci.org/Pylons/pyramid/branches

Would you please submit a PR? You can use tox locally or watch TravisCI after submitting a PR to see results.

RaHus:
@stevepiercy Sure, i am on it

RaHus:
@stevepiercy Looks like its working now, should i touch CHANGES.txt also?

stevepiercy:
LGTM. I'd like one more core contributor to review.

Touching CHANGES.txt under "Bug Fixes" would be appropriate.

RaHus:
tseaver mentioned in the [PR](https://github.com/Pylons/pyramid/pull/2075#issuecomment-153348455) that this is actually a WebTest [packaging bug](https://github.com/Pylons/webtest/issues/143) and it should be fixed upstream.

PR #2075, opened 2015-11-03
Include backported ordereddict package from PYPI when using python 2.…

RaHus:
…6 since WebTest requires it. see #2071

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,9 @@ tests_require = [
 if not PY3:
     tests_require.append('zope.component>=3.11.0')
 
+if py_version == (2, 6):
+    tests_require.append('ordereddict')
+
 docs_extras = [
     'Sphinx >= 1.3.1',
     'docutils',


tseaver:
From looking at [WebTest's `setup.py`](https://github.com/Pylons/webtest/blob/master/setup.py#L22), this change seems unnecessary:  the latest WebTest version (2.0.19) already has that optional dependeny.

tseaver:
Our test failures are due to the fact that WebOb 2.0.19 released a `none-any` wheel which is not compatible with 2.6 (the source version is fine).  Let's get the WebTest [packaging bug](https://github.com/Pylons/webtest/issues/143) fixed instead of replicating its changes in Pyramid (we would be propagating the same error to downstreams if somebody released a Pyramid wheel).

RaHus:
@tseaver It sure didn't stick quite well since it started failing suddenly, i am closing this PR and adding a comment in #2071

Issue #235, opened 2011-07-14
When no mo files exist for a domain, pluralizer raises inscrutable error


mcdonc:
This sort of error will be raised when no mofile for a given domain exists, and someone attempts a pluralization against that domain:

```
Traceback (most recent call last):
  File "/home/chrism/projects/pyramid/pyramid/tests/test_i18n.py", line 487, in test_default_germanic_pluralization
    result = t.dungettext('messages', 'foo1', 'foos1', 2)
  File "/home/chrism/projects/pyramid/pyramid/i18n.py", line 337, in dungettext
    return self._domains.get(domain, self).ungettext(singular, plural, num)
  File "/home/chrism/opt/Python-2.6.5/lib/python2.6/gettext.py", line 409, in ungettext
    tmsg = self._catalog[(msgid1, self.plural(n))]
AttributeError: 'Translations' object has no attribute 'plural'
```

PR #2102, opened 2015-11-09
Fix non german plurals

quantum-omega:
When using translations for a language whose plurals are not defined the same way as german languages (n != 1), the default "plural" attribute of Translations is used even if it should instead use the one assigned to the Translations object by gettext. I came upon this as we have french translations and the plural form was used for zero items, even though this is incorrect for french.

My fix gives a name to the default plural (_DEFAULT_PLURAL) and checks whether it is still the value of Translations.plural when another Translations is added to it. If so, then instead of the default unset plural, the one from the added Translations object is used.


tseaver:
It doesn't seem obvious to me the "host" `Translations` instance should use the `plural` from the "added" instance:  if the user wants to override it, she should be setting it explicitly.

If we *do* want this behavior, then we should think about using a `plural` property, backed by a different attribute (e.g. `_plural`), and use `None` as the default value for the backing property:  we can then distinguish between "explicit" assignment of the value vs. the default, and override only in the latter case.

We should probably also be exposing / testing / documenting the default `plural` implementation (maybe even alternatives as we know of them).

quantum-omega:
My initial idea was to put this functionality in make_localizer as this is where Translations are mostly going to be used in pyramid and it is where it bites the most if it doesn't work well. One would expect the i18n module to just do what it's supposed to do when setup properly and they will let the framework call make_localizer when it needs to do so. Later, I figured the Translations class was where this functionality belonged as it made it work consistently when used directly and it just feels like it should be one of its concerns.

The user rarely ever works with Translations objects themselves. They rarely want (based on the use cases presented in the documentation) to change the plural property to another value; they want it to be set according to the locale name passed to make_localizer. This is something that the aggregate object that Translations is breaks.

I like the idea of using a property for plural but maybe the problem is deeper than this. Maybe, if Translations contains many other Translations objects, then it should return the right plural according to which sub-translations contained the actual translation string. But how to determine this? And does it even make sense to mix locales within the same Translations? Not the way it was designed in the gettext package.

I'm not sure I'm making myself quite clear here...

MatthewWilkes:
Hi @quantum-omega 

I was looking at this today, but I don't understand the specific case you're trying to solve. Is it for when you're trying to use the pluralize function but without creating a .po file for the language?

This change would only improve things for languages where pluralforms is 2, so I wonder if it'd be better to push people more towards translation files for plurals.

quantum-omega:
Well, it's been more than a year so details are blurry but the use case that uncovered this problem is one using translation files.

The Translations object's pluralize is assigned a default value in order to prevent #235. Translations is basically a composite of gettext.GNUTranslations and when a translation (.mo) file is added to the Translations object, the value of self.plural is not updated with the plural function from the added GNUTranslations. However, it is Translations' plural function that is used by the rest of i18n.py's translation mechanism in order to determine which plural form to use.

Thus, when the (let's assume only) language present in the .po files does not have a plural in the form n != 1, Pyramid fails to use the right plural. As the plural is read correctly from the files by gettext, my fix should at least make things better for any use case where only one language is present in a catalog, whether that language has two or more plural forms.

Mixing languages in a catalog is something that I do not remember being a common or even feasible use case, but it's been a while so I might be wrong. In any case, I cannot see how this could be working as the germanic plural forms are taking over no matter what. Translations.plural would have to be way more sophisticated to make that work. For one, it would need to know from which sub-translations a given translation text came from.

I would like to point out that, even with us knowing the correct plural forms for French and working primarily in that language, we did not see those errors for a long while as getting the wrong texts to show up can be corner cases and validating the texts displayed is not something one often does in automated tests.

ztane:
I guess the translations supported the case where you could layer many languages so that at least one message would be useful. And what should be tested is what happens if one uses Arabic, say - when that falls back to English, does it break with IndexError or what?

MatthewWilkes:
I think I'm missing some aspect of what the problem is, maybe if I share code it will be helpful?

`locale/en/LC_MESSAGES/hacking.po`
```po
# English translations for PACKAGE package
# This file is distributed under the same license as the PACKAGE package.
# Matthew Wilkes <matt@tiwaz.uk>, 2016.
msgid ""
msgstr ""
"Project-Id-Version: PACKAGE 1.0\n"
"POT-Creation-Date: 2016-12-05 10:17+0100\n"
"PO-Revision-Date: 2016-12-05 10:18+0100\n"
"Last-Translator: Matthew Wilkes <matt@tiwaz.uk>\n"
"Language-Team: English\n"
"Language: en\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Generated-By: Lingua 4.11\n"
"Plural-Forms: nplurals=2; plural=n>1;\n"

#: hacking/views.py:6
msgid "results_found"
msgid_plural ""
msgstr[0] "${number} result found"
msgstr[1] "${number} results found"
```

**Note:**  I used the french plural definition, rather than the Germanic one, despite this being declared as English.

and `views.py`:

```python
from pyramid.i18n import get_localizer
from pyramid.view import view_config


@view_config(route_name='home', renderer='templates/mytemplate.pt')
def my_view(request):
    n = int(request.GET.get('n', 0))
    number = get_localizer(request).pluralize(
        'results_found',
        '',
        n,
        domain='hacking',
        mapping={'number': n}
    )
    return {'project': 'hacking', 'number': number}
```

Then, when I get the view, I see the following results:

Query String | Text
-------------------
?n=0 | **0 result found**
?n=1 | 1 result found
?n=2 | 2 results found
?n=3 | 3 results found
?n=4 | 4 results found
?n=5 | 5 results found

This seems to be at odds with your latest message, which says: 

> Thus, when the (let's assume only) language present in the .po files does not have a plural in the form n != 1, Pyramid fails to use the right plural.

I'm sure I'm missing something here, but I'm having difficulty reproducing the bug.

ztane:
@MatthewWilkes I guess one of the problems now is that the fallbacks still would use the French plural? What happens if you use the Arabic expression and fall back to gettext strings?

ztane:
`nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 ? 4 : 5);` is the Arabic one, want to see if there would be index errors or sth for non-existent plurals

MatthewWilkes:
@ztane I don't understand, I'm sorry. What fallback are you talking about?

Also, when you say non-existent plurals, do you mean a .po file that declares 6 plurals but doesn't always provide 6? In that case I think it would be legitimate for that to raise exceptions.

ztane:
@MatthewWilkes meaning that the translation doesn't exist in the catalog

ztane:
Though, it seems that you're using the pluralize and msg_id the "zopey way" (I really don't know) above.

MatthewWilkes:
The pluralize method is utterly braindead in that it has two arguments, singular and plural, rather than a messageid. Is this problem with passing in literal strings to both arguments rather than using the plural support in gettext?

If so, I don't think we should support that, and I'd argue for limiting the API to not offer that option, as it makes it impossible to support nplurals != 2 in your code.

ztane:
no, it is not utterly braindead, the gettext is meant to use so that the source code is written in English and the actual msgid is the English language text. To distinguish between 2 different contexts. And those forms should be handled with n != 1.

quantum-omega:
Since it's been a while since I started this whole thing, I decided to remove the awful patching of pyramid.i18n.make_localizer we use in our application and see if the problem still presents itself. The code surrounding this didn't change but maybe an update in the gettext module or something might have had an impact.

So, with Pyramid 1.7 deployed on our test environment (Python 2.7.12), the problem is still there, meaning we get plural form index 1 (with ending s) in French on n == 0.

I'll play with @MatthewWilkes's snippet and see if I can reproduce it in that simpler context and if not try to figure how we might have messed up or confused pyramid.i18n somehow.

quantum-omega:
After a bit of trial and error and some confusion as @MatthewWilkes's code snippet worked perfectly fine, I found out how to make it fail. If, instead of using a domain for the application (here "hacking"), one uses the default domain name ("messages"), renaming the .po file to messages.po and the associated .mo file, and using domain="messages", it mysteriously stops working.

Using an application-specific domain name also seems to solve the problem for our application. We have been using the default domain out of convenience and it was on the list of things to do to use an application-specific one anyway, so that's a fix for our application and maybe an explanation as to why not many other people came across this. However, unless I misunderstood something about the "messages" domain, it seems to be buggy. @ztane seems quite knowledgeable about gettext and could maybe confirm whether it's normal or not to use "messages" as a domain for one's application.

MatthewWilkes:
@ztane Well, we'll agree to disagree on that, then. I personally think the assumption that source code is in English is a bad one, but that's more of a discussion over beers thing.

@quantum-omega Thanks for the hint at reproducing it, I'll take a look today.

MatthewWilkes:
@quantum-omega Thanks for the reproducer instructions, it was enough for me to isolate the problem. I have submitted a new PR #2859 that solves this in what I hope is a cleaner way.

quantum-omega:
@MatthewWilkes Thanks for looking into this. Your solution seems very good and I will close this PR.

MatthewWilkes:
@quantum-omega Thank **you** very much for finding this bug, reporting it, and being so responsive more than year later!

PR #2109, opened 2015-11-12
apply some suggestions from quantified code analysis

mmerickel:


pyramid/util.py
```diff
@@ -30,7 +30,7 @@
 
 class DottedNameResolver(_DottedNameResolver):
     def __init__(self, package=None): # default to package = None for bw compat
-        return _DottedNameResolver.__init__(self, package)
+        _DottedNameResolver.__init__(self, package)
```
tseaver: Seems like using `super` would be better.

PR #2115, opened 2015-11-12
Enable optional session encryption

tseaver:
Adds `pyramid.session.BlowfishPickleSerializer`. This class is an encrypting Webob cookie serializer, which uses the pickle protocol to dump Python data to bytes, then encrypts the value symmetrically using the Blowfish algorithm.

To encrypt sessions, pass an instance of this class as the `serializer` argument to `pyramid.session.BaseCookieSessionFactory`.

Before using this serializer, install `pyramid[encrypted_cookies]` to get the necessary dependency (`pycrypto`).

### DIFF
--- a/docs/narr/sessions.rst
+++ b/docs/narr/sessions.rst
@@ -71,6 +71,35 @@ using the :meth:`pyramid.config.Configurator.set_session_factory` method.
    security doesn't matter", and you are sure your application has no
    cross-site scripting vulnerabilities.
 
+.. _encrypting_the_session:
+
+Encrypting the Session
+----------------------
+
+Pyramid provides an optional session serializer which encrypts the pickled
+session data using the Blowfish algorithm.  You can use this serializer in
+your :app:`Pyramid` application by using the
+:meth:`pyramid.config.Configurator.set_session_factory` method.
+
+.. code-block:: python
+   :linenos:
+
+   from pyramid.session import BaseCookieSessionFactory
+   from pyramid.session import BlowfishPickleSerializer
+   serializer = BlowfishPickleSerializer('itsaseekreet')
+   my_session_factory = BaseCookieSessionFactory(serializer)
+
+   from pyramid.config import Configurator
+   config = Configurator()
+   config.set_session_factory(my_session_factory)
+
+.. note::
+
+   The :class:`BlowfishPickleSerializer` requires that you install
+   the optional ``pycrypto`` dependency, either directly (e.g., via
+   ``easy_install pycrypto``) or via the ``encrypted_cookies`` extra
+   (e.g., ``easy_install pyramid[encrypted_cookies]``).
+
 .. index::
    single: session object
 
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -3,12 +3,23 @@ import binascii
 import hashlib
 import hmac
 import os
+import struct
 import time
 
 from zope.deprecation import deprecated
 from zope.interface import implementer
 
 from webob.cookies import SignedSerializer
+
+try:
+    from Crypto import Random
+except ImportError:  # pragma: NO COVER
+    _HAS_CRYPTO = False
+else:
+    _HAS_CRYPTO = True
+    from Crypto.Cipher import Blowfish
+    BLOCK_SIZE = Blowfish.block_size
+    IV = Random.new().read(BLOCK_SIZE)
 
 from pyramid.compat import (
     pickle,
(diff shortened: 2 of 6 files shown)


mmerickel:
So I'm pretty strongly -1 on this. When I reworked the `SignedCookieSessionFactory` I went through many discussions with @mcdonc on this and he was always very anti adding any crypto code to pyramid, despite me adding all of the hooks for custom serializers. We ended up pushing all of the code signing down into webob's `SignedCookieProfile` and sharing it between auth_tkt and sessions.

On a more recent note I just put a lot of work into splitting out all of the pshell implementations (pyramid_ipython and pyramid_bpython) into separate packages and removing them from core. Adding in a significant portion of crypto code including pycrypto, which isn't even the recommended package I've seen posted latey (in favor of cryptography), seems like a dangerous road to go down.

Did you consider creating `pyramid_encrypted_sessions` on pypi, or `pyramid_blowfish` or something, versus pushing this to core?

bertjwregeer:
If we are pulling in pycrypto, why not go with AES-256? If pycrypto is using OpenSSL underneath, in most cases you get free CPU speed-ups (AES-NI) for encrypting/decrypting, also using a standard used by the US gov't for protecting things classified up to top secret means it should be adequate for encrypting cookies.

pyramid/session.py
```diff
@@ -10,6 +11,16 @@
 
 from webob.cookies import SignedSerializer
 
+try:
+    from Crypto import Random
```
bertjwregeer: This is a RNG implemented in Python, why not defer to getting randomness from the appropriate location (/dev/urandom?)

docs/narr/sessions.rst
```diff
@@ -71,6 +71,35 @@ using the :meth:`pyramid.config.Configurator.set_session_factory` method.
...
+your :app:`Pyramid` application by using the
+:meth:`pyramid.config.Configurator.set_session_factory` method.
+
+.. code-block:: python
+   :linenos:
+
+   from pyramid.session import BaseCookieSessionFactory
+   from pyramid.session import BlowfishPickleSerializer
+   serializer = BlowfishPickleSerializer('itsaseekreet')
+   my_session_factory = BaseCookieSessionFactory(serializer)
```
bertjwregeer: @raydeo just informed me this is not public API (BaseCookieSessionFactory)

mmerickel: Nah it is public... I forgot because we initially intended to keep it private and I think it slipped through the cracks and is now public.

pyramid/session.py
```diff
@@ -626,3 +637,46 @@ def SignedCookieSessionFactory(
...
+
+    To encrypt sessions, pass an instance of this class as the ``serializer``
+    argument to :class:`BaseCookieSessionFactory`.
+
+    .. note::
+
+       Before using this serializer, install ``pyramid[encrypted_cookies]`` to
+       get the necessary dependency (``pycrypto``).
+    """
+    def __init__(self, secret):
```
mmerickel: There is no signing of the content since it's using the `BaseCookieSessionFactory`. The signing needs to be done with a separate secret from the encryption key. There is also no salt to prevent dual-use of secrets in the application. See the `SignedCookieSessionFactory` for an example of the salt.

pyramid/session.py
```diff
@@ -626,3 +637,46 @@ def SignedCookieSessionFactory(
...
+        if not _HAS_CRYPTO:
+            raise ValueError("No crypto:  install 'pycrypto' first!")
+        self.secret = secret
```
bertjwregeer: This secret should NOT be used directly as the key to any symmetric cipher. You would want to extend it using a key derivation function (such as KDF2). You'll also want a unique salt as @mmerickel mentioned (luckily the KDF2 will happily take that too).

By using the secret directly you are making it easier to attack the crypto. KDF2 with 50,000 iterations would be much better, since this would also lengthen the existing key from what the user provided to the key size required by the cipher. This won't increase the entropy of the key, so if the key was only 12 bits of entropy, after KDF you'd still only have 12 bits of entropy, but the key space to search when brute forcing went from for example 20 bits to 128 bits.

bertjwregeer:
I too am a -1 on having this within Pyramid itself. We also open ourselves up to having to worry about export laws regarding crypto as Pyramid is developed in the US.

tseaver:
OK, I'll trash this one and put the functionality in an add-on, using `PyNaCl` as suggested by @bertjwregeer in IRC.

PR #2117, opened 2015-11-13
Supersedes: #2050 int conversion checks

bertjwregeer:


bertjwregeer:
Working on coverage...

bertjwregeer:
Missing coverage for the timeout variable, previous tests did not adequately verify that timeout was actually doing what it was supposed to be doing.

PR #2119, opened 2015-11-13
abort when using reload with daemon

### LINKED ISSUE #912: pserve --reload with --daemon

mmerickel:
Fix #912


bertjwregeer:
This looks good. +1. Waiting on travis to complete before merging.

PR #2141, opened 2015-11-18
Remove references to python-distribute.org

offby1:
That domain no longer holds the Guide to Packaging.

### DIFF
--- a/docs/narr/project.rst
+++ b/docs/narr/project.rst
@@ -680,8 +680,7 @@ testing, packaging, and distributing your application.
    ``setup.py`` is the de facto standard which Python developers use to
    distribute their reusable code.  You can read more about ``setup.py`` files
    and their usage in the `Setuptools documentation
-   <http://peak.telecommunity.com/DevCenter/setuptools>`_ and `The Hitchhiker's
-   Guide to Packaging <http://guide.python-distribute.org/>`_.
+   <http://peak.telecommunity.com/DevCenter/setuptools>`_.
 
 Our generated ``setup.py`` looks like this:
 
--- a/docs/narr/scaffolding.rst
+++ b/docs/narr/scaffolding.rst
@@ -22,9 +22,7 @@ found by the ``pcreate`` command.
 
 To create a scaffold template, create a Python :term:`distribution` to house
 the scaffold which includes a ``setup.py`` that relies on the ``setuptools``
-package.  See `Creating a Package
-<http://guide.python-distribute.org/creation.html>`_ for more information about
-how to do this.  For example, we'll pretend the distribution you create is
+package.  For example, we'll pretend the distribution you create is
 named ``CoolExtension``, and it has a package directory within it named
 ``coolextension``.
 


bertjwregeer:
Is this the same guide: http://docs.python-guide.org/en/latest/shipping/packaging/?

bertjwregeer:
Or this one: https://the-hitchhikers-guide-to-packaging.readthedocs.org/en/latest/ ?

stevepiercy:
We should replace, not simply remove, external references in our documentation.

This is the recommended documentation to replace the Guide to Packaging and The Hitchhikers Guide to Packaging:
https://packaging.python.org/en/latest/

according to its readme:
https://github.com/pypa/python-packaging-user-guide

as well as this issue:
https://bitbucket.org/tarek/hitchhiker-guide-packaging/issues/40/please-take-down-http-guidepython

However, the PyPUG documentation might not align with the Pyramid documentation in regards to installation, packaging, and distribution. If that's true, then maybe the Hitchhikers Guide is more relevant at this time. I really don't know which is better until we switch all our docs to use pip. Perhaps @dstufft would have a good suggestion?

pauleveritt:
Forgetting content, just on mechanics…should we do Intersphinx for these to make sure references breakage gets reported at build time?

—Paul

stevepiercy:
Intersphinx when possible. I found that sometimes docs lack a reference label at the top of every .rst file, making the page not show up in the inventory, [webob, for example](https://github.com/Pylons/webob/blob/master/docs/index.txt).

I've been adding them to the Intersphinx section in conf.py when I notice them and can make an Intersphinx link.

domenkozar:
The official (of course depends who you ask) new url is http://packaging.python.org/

bertjwregeer:
@stevepiercy What do I need to do to make Intersphinx work?

stevepiercy:
@bertjwregeer I'll try to set up a working example. This is for webob, correct?

bertjwregeer:
Yeah, what am I missing in WebOb so that Intersphinx can work correctly? PR’s with examples would be extremely helpful ;-)

mmerickel:
intersphinx just uses the `objects.inv` from the root url. http://docs.webob.org/en/latest/objects.inv in this example. You shouldn't have to change anything except add it to the `conf.py` in pyramid. (It's already there). packaging.python.org also has one, so just add it to pyramid's `conf.py` like all the others.

stevepiercy:
In addition to what @mmerickel said, if you want to use intersphinx to link to specific pages or sections in the *narrative* documentation, the reStructuredText markup in the destination must have [labels](http://sphinx-doc.org/ext/intersphinx.html#module-sphinx.ext.intersphinx).

> When using the “new” format (see below), you can even force lookup in a foreign set by prefixing the link target appropriately. A link like :ref:`comparison manual <python:comparisons>` will then link to the label “comparisons” in the doc set “python”, if it exists.

The labels will then be added to `objects.inv`, and intersphinx can use them.

As an example, I put the following in any documentation that is not webob's.

```rst
:ref:`webob docs <webob:index>`
```

Without a label in the webob documentation for its home page, no link is generated. But by adding a label to webob's index.txt:

```diff
+ .. _index:
+ 
WebOb
```

It will now render with a link.

I'll submit a PR to the webob docs in a minute or two.

stevepiercy:
Coming back to the original issue, @domenkozar confirmed my findings (http://packaging.python.org/ redirects to https://packaging.python.org/en/latest/), so I'll submit a separate PR to use the "official" link.

stevepiercy:
I take that back. I forgot my previous concern:

> However, the PyPUG documentation might not align with the Pyramid documentation in regards to installation, packaging, and distribution. If that's true, then maybe the Hitchhikers Guide is more relevant at this time. I really don't know which is better until we switch all our docs to use pip.

IOW, we have "official", but is it "right for Pyramid" at this moment in time?

docs/narr/scaffolding.rst
```diff
@@ -22,9 +22,7 @@ found by the ``pcreate`` command.
 
 To create a scaffold template, create a Python :term:`distribution` to house
 the scaffold which includes a ``setup.py`` that relies on the ``setuptools``
-package.  See `Creating a Package
-<http://guide.python-distribute.org/creation.html>`_ for more information about
-how to do this.  For example, we'll pretend the distribution you create is
```
stevepiercy: Proposed:

```
package.  See `Packaging and Distributing Projects
<https://packaging.python.org/en/latest/distributing/>`_ for more information about
how to do this.  For example, we'll pretend the distribution you create is
```

docs/narr/project.rst
```diff
@@ -680,8 +680,7 @@ testing, packaging, and distributing your application.
    ``setup.py`` is the de facto standard which Python developers use to
    distribute their reusable code.  You can read more about ``setup.py`` files
    and their usage in the `Setuptools documentation
-   <http://peak.telecommunity.com/DevCenter/setuptools>`_ and `The Hitchhiker's
-   Guide to Packaging <http://guide.python-distribute.org/>`_.
```
stevepiercy: Proposed:

```
<http://peak.telecommunity.com/DevCenter/setuptools>`_ and `Python Packaging
User Guide <https://packaging.python.org/en/latest/>`_.
```

stevepiercy:
I let the proposed changes simmer for 12 days, and I think it's done. I'm closing this PR in favor of merging #2141. I'll also rollback to earlier versions.

PR #2148, opened 2015-11-21
add optional step to update the Pyramid Dash docset

stevepiercy:


stevepiercy:
I'm leaving this open to discuss because:

* Dash is a commercial product on OS X or iOS only.
* docset generation is not scripted or automated yet.

stevepiercy:
I'm closing this PR because of the above mentioned reasons, it doesn't add value for the extra work it requires.

PR #2151, opened 2015-11-25
Add the abllity to set multiple groups

vanderlindenaj:
from a list and or from the requested users system defined groups.

### DIFF
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -167,6 +167,18 @@ class PServeCommand(object):
             dest='set_group',
             metavar="GROUP",
             help="Set the group (usually only possible when run as root)")
+        parser.add_option(
+            '--groups',
+            dest='set_groups',
+            metavar="GROUPS",
+            help="Set the groups (comma seperated with no spaces) (usually only possible when run as root) (only relevant with --user and --group)")
+        parser.add_option(
+            '--user-groups',
+            dest='user_groups',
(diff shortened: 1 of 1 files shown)


mmerickel:
Can I ask how you're using `pserve`? Are you aware that we are deprecating the daemon features? In my mind the user privileges are not useful without the daemon so I'm asking how you use pserve.

vanderlindenaj:
We are using the daemon option via init.d for startup.

Any sugestions?

mmerickel:
Sure, the recommended way to start daemon's in sysv init scripts is via the `start-stop-daemon` function which can do everything that you should want to do (set users, groups, pid files, etc).

@bertjwregeer I can't think of a reason why we didn't deprecate the set-user functionality when we deprecated daemon options. Can you?

bertjwregeer:
I can't think of a reason either, but we should. Also, I am a big :-1: to this change. This functionality does not and should not belong in `pserve`.

mmerickel:
I tend to agree. @vanderlindenaj you caught us at a time when we're attempting to strip `pserve` down to make it more focused on just starting up / reloading your code. Like I said the daemon functionality you're currently relying on has been deprecated in the latest 1.6 beta releases, and I think we're likely to deprecate the `set-user` functionality as well.

Thank you for the PR but I'm going to reject it as it's building on a features we are removing.

vanderlindenaj:
Cool thanks guys.

Keep up the good work.

👍

PR #2152, opened 2015-11-26
remove py2-docs from tox.ini and reorder so coverage is last

mmerickel:


mmerickel:
I already updated RTD to use py3 for building the docs to keep things consistent.

Issue #1845, opened 2015-06-15
add a chapter to the docs on security considerations and hardening a pyramid app


mmerickel:
This chapter should cover things like headers from trusted proxies and cache poisoning.

https://github.com/Pylons/pyramid/issues/1591 https://github.com/Pylons/pyramid/pull/1570 https://github.com/Pylons/pyramid/pull/1348

PR #2162, opened 2015-12-04
Implement OWASP session management recommendations

canni:
According to [OWASP Session management Cheat Sheet](https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Renew_the_Session_ID_After_Any_Privilege_Level_Change) renewing session is mandatory, on any privilege change, which is happening on all calls to `remember` or `forget` of `SessionAuthenticationPolicy`.

This change is backward-compatible, but probably
we should really empathize turning this on in docs.

I want to hear some feedback first, before I jump into updating docs, a covering this change with tests.

### TODO:
- [ ] docs updates
- [ ] unit-tests
- [ ] update `AuthTKTAuthenticationPolicy`

### DIFF
--- a/pyramid/authentication.py
+++ b/pyramid/authentication.py
@@ -1053,23 +1053,38 @@ class SessionAuthenticationPolicy(CallbackAuthenticationPolicy):
         steps.  The output from debugging is useful for reporting to maillist
         or IRC channels when asking for support.
 
+    ``invalidate``
+
+        Default: ``False``. If ``invalidate`` is ``True``, every call to the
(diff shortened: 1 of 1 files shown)


bertjwregeer:
https://github.com/Pylons/pyramid/pull/1570

Please read the comments on that PR, we have decided against merging something similar in the past, and don't believe we have figured out the right place to do this.

If you would like this, please sub-class your favourite authentication policy, or create your own that does this.

bertjwregeer:
Please also take a look at: https://github.com/Pylons/pyramid/issues/1845

PR #2168, opened 2015-12-08
Update views.py

Pomlon:
Previous version of views.py looked for templates in:
python3.4/site-packages/pyramid/home.pt

adding >tutorial:< makes it look for the template in the correct tutorial folder.

This bug is caused by not specifying the app-module too look for templates in.

### DIFF
--- a/docs/quick_tutorial/templating/tutorial/views.py
+++ b/docs/quick_tutorial/templating/tutorial/views.py
@@ -2,12 +2,12 @@ from pyramid.view import view_config
 
 
 # First view, available at http://localhost:6543/
-@view_config(route_name='home', renderer='home.pt')
+@view_config(route_name='home', renderer='tutorial:home.pt')
 def home(request):
     return {'name': 'Home View'}
 
 
 # /howdy
-@view_config(route_name='hello', renderer='home.pt')
+@view_config(route_name='hello', renderer='tutorial:home.pt')
 def hello(request):
     return {'name': 'Hello View'}


stevepiercy:
@Pomlon what do you mean "previous version of views.py"? When I follow the steps in Quick Tutorial in order, I cannot replicate this issue.

Pomlon:
Previous as in not fixed by me. Bear with me, I'm new to Pyramid.

I did follow the whole tutorial from the very beginning and unless I put in the modulename: clause in renderer, it looks for the templates in the wrong place.

Maybe it's platform-specific? Maybe it's because I'm using virtualenv instead of pyvenv as the tutorial says?

I'm using:
ubuntu 14.04 LTS
Python 3.4.3
virtualenv 13.1.2

mmerickel:
Where is your `views.py` located relative to the template? Are they not in the same folder?

Pomlon:
It didn't work no matter if I put it in project/tutorial or project/tutorial/templates. Also please notice it was looking in /python3.4/ not in the project folder for some reason.

The views.py is in project/tutorial

stevepiercy:
Let's check your setup. Did you follow the [Requirements](http://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/requirements.html) exactly, or go off-road with pip or activate?

mmerickel:
Can you put your repository on github to look at?

mmerickel:
Closing. I'm pretty sure this isn't an actual problem but rather just something in the author's project but without a reproducible example there's nowhere to go from here.

PR #2171, opened 2015-12-10
config.add cache buster

### LINKED ISSUE #2145: decouple cache buster from top-level asset spec
Currently the cache buster api is coupled to the asset spec defined in `config.add_static_view('myapp:static', '/static', cachebust=CacheBuster())`. This does not play well with `config.override_assets(to_override='myapp:static/theme/', override_with='mytheme:static/')` (a desirable pattern for static views). The issue is that some assets are pulled in from other packages which are not at `myapp:static` and thus cannot be properly cache busted without knowing what is overridden and what is not.

1) The `ICacheBuster` interface receives no information about the fact that this asset is overridden. The spec it received is `myapp:static`, never `mytheme:static`.
2) The 10-year cache expiration time is applied whether or not the asset is actually cache busted.

This should be decoupled such that raw paths can be busted. I'm proposing `config.add_cache_buster('mytheme:static/', cache_buster, cache_max_age)`.

Two problems with this API that we'll need to solve:

1) In multiple calls to `add_cache_buster` which cache buster is used on the matching side when a URL comes into the system?
2) For assets that match a cache busted pattern, how do we tell whether the asset is actually cache busted or not? Should we enhance `ICacheBuster.pregenerate` to return some info about whether it is busted? Check whether the parameters are mutated?

ping @bertjwregeer, @dstufft

mmerickel:
Remove `cachebust=` from `config.add_static_view` and provide a `config.add_cache_buster` API. The cache busters are tied directly to raw asset paths instead of high-level specifications. For example:

```python
config.add_static_view('static', 'myapp:static')
config.override_asset('myapp:static/', 'theme:static/')

def cb(pathspec, subpath, kw):
    # pathspec -> 'theme:static/foo.png'
    # subpath -> 'foo.png'
    # kw -> args to route_url
    return subpath, kw
config.add_cache_buster('theme:static', cb)
```

Note that the cache buster is tied to `theme:static` and not `myapp:static`.

I'm open to tweaking this API. For example, it might be more intuitive to tie the cache busters to the high-level specs in many cases. Perhaps it would be a better default to tie things to the high-level spec and provide an option to add_cache_buster such as `base_spec=True` or `raw_spec=True` for scenarios where the cache buster is a manifest tied directly to the specific assets. I don't think this would be very difficult to support and it'd be better to get a sane default to start with versus trying to change it. The advantage of using a high-level spec by default is that it would encompass overrides without having to add new cache busters per override.

This PR also drops the entire matching side of the cache busting and focuses purely on generating urls, allowing custom tweens, etc to serve the urls as they see fit.

Fixes #2145.

ping @mcdonc @bertjwregeer @dstufft

### DIFF
--- a/docs/narr/assets.rst
+++ b/docs/narr/assets.rst
@@ -557,7 +557,7 @@ use some of the following options to get started:
 
 * Configure your asset pipeline to rewrite URL references inline in
   CSS and JavaScript. This is the best approach because then the files
-  may be hosted by :app:`Pyramid` or an external CDN without haven't to
+  may be hosted by :app:`Pyramid` or an external CDN without having to
   change anything. They really are static.
 
 * Templatize JS and CSS, and call ``request.static_url()`` inside their
@@ -825,7 +825,7 @@ imagine you have overridden an asset defined in this manifest with a new,
 unknown version. By default, the cache buster will be invoked for an asset
 it has never seen before and will likely end up returning a cache busting
 token for the original asset rather than the asset that will actually end up
-being served! In order to get around this issue it's possible to attach a
+being served! In order to get around this issue, it's possible to attach a
 different :class:`pyramid.interfaces.ICacheBuster` implementation to the
 new assets. This would cause the original assets to be served by their
 manifest, and the new assets served by their own cache buster. To do this,
@@ -853,14 +853,14 @@ option. For example:
    theme_cb = ManifestCacheBuster('theme:static/manifest.json')
    config.add_cache_buster('theme:static', theme_cb, explicit=True)
 
-In the above example there is a default cache buster, ``my_cb`` for all assets
-served from the ``myapp:static`` folder. This would also affect
+In the above example there is a default cache buster, ``my_cb``, for all
+assets served from the ``myapp:static`` folder. This would also affect
 ``theme:static/background.png`` when generating URLs via
 ``request.static_url('myapp:static/background.png')``.
 
 The ``theme_cb`` is defined explicitly for any assets loaded from the
 ``theme:static`` folder. Explicit cache busters have priority and thus
 ``theme_cb`` would be invoked for
-``request.static_url('myapp:static/background.png')`` but ``my_cb`` would be
-used for any other assets like
+``request.static_url('myapp:static/background.png')``, but ``my_cb`` would
+be used for any other assets like
 ``request.static_url('myapp:static/favicon.ico')``.
--- a/docs/narr/assets.rst
+++ b/docs/narr/assets.rst
@@ -366,8 +366,7 @@ resource and requests the asset, regardless of any caching policy set for the
 resource's old URL.
 
 :app:`Pyramid` can be configured to produce cache busting URLs for static
-assets by passing the optional argument, ``cachebust`` to
-:meth:`~pyramid.config.Configurator.add_static_view`:
+assets using :meth:`~pyramid.config.Configurator.add_cache_buster`:
 
 .. code-block:: python
    :linenos:
@@ -376,14 +375,18 @@ assets by passing the optional argument, ``cachebust`` to
    from pyramid.static import QueryStringConstantCacheBuster
 
    # config is an instance of pyramid.config.Configurator
-   config.add_static_view(
-       name='static', path='mypackage:folder/static',
-       cachebust=QueryStringConstantCacheBuster(str(int(time.time()))),
-   )
+   config.add_static_view(name='static', path='mypackage:folder/static/')
+   config.add_cache_buster(
+       'mypackage:folder/static/',
+       QueryStringConstantCacheBuster(str(int(time.time()))))
 
-Setting the ``cachebust`` argument instructs :app:`Pyramid` to use a cache
-busting scheme which adds the curent time for a static asset to the query
-string in the asset's URL:
+.. note::
+   The trailing slash on the ``add_cache_buster`` call is important to
+   indicate that it is overriding every asset in the folder and not just a
+   file named ``static``.
+
+Adding the cachebuster instructs :app:`Pyramid` to add the current time for
+a static asset to the query string in the asset's URL:
 
 .. code-block:: python
    :linenos:
@@ -392,10 +395,7 @@ string in the asset's URL:
    # Returns: 'http://www.example.com/static/js/myapp.js?x=1445318121'
 
 When the web server restarts, the time constant will change and therefore so
-will its URL.  Supplying the ``cachebust`` argument also causes the static
-view to set headers instructing clients to cache the asset for ten years,
-unless the ``cache_max_age`` argument is also passed, in which case that
-value is used.
+will its URL.
 
 .. note::
 
@@ -410,7 +410,7 @@ Disabling the Cache Buster
 
 It can be useful in some situations (e.g., development) to globally disable all
 configured cache busters without changing calls to
-:meth:`~pyramid.config.Configurator.add_static_view`.  To do this set the
+:meth:`~pyramid.config.Configurator.add_cache_buster`.  To do this set the
 ``PYRAMID_PREVENT_CACHEBUST`` environment variable or the
 ``pyramid.prevent_cachebust`` configuration value to a true value.
 
@@ -419,9 +419,9 @@ configured cache busters without changing calls to
 Customizing the Cache Buster
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``cachebust`` option to
-:meth:`~pyramid.config.Configurator.add_static_view` may be set to any object
-that implements the interface :class:`~pyramid.interfaces.ICacheBuster`.
+Calls to :meth:`~pyramid.config.Configurator.add_cache_buster` may use
+any object that implements the interface
+:class:`~pyramid.interfaces.ICacheBuster`.
 
 :app:`Pyramid` ships with a very simplistic
 :class:`~pyramid.static.QueryStringConstantCacheBuster`, which adds an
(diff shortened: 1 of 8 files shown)


dstufft:
For my own personal use cases using the "real" asset spec is a better option than using the "high level" one. It seems to me when your cachebuster depends on information specific to the file (e.g. manifest cache buster) you'd want to use the real asset spec and when it's just generic information (latest git revision of the app?) you'd want to do the high level view.

However, it seems like it's easier to deal with the latest git revision case if it's tied to the real asset spec than it is to deal with the needing file specific information if it's tied to the high level asset spec.

You really just need to add duplicate calls to ``add_cache_buster`` with the same cachebuster instead of needing to write a wrapper that somehow discovered the "real" asset spec that is going to be served. IOW it's basically the situation from before, where the cachebuster doesn't have the required information.

One thing that might be a useful thing to do is to also pass in the high level spec to the cache buster in addition to the low level so people can select which one they want to base their cache busting off of?

dstufft:
Oh, and does this do away with the automatically setting a +10 year expiration on cache busted files, or does that still happen? I think that could work very well with if cache busters are using the real asset specs instead of the high level ones, it'd let you have the cachebuster decide on a url by url basis if it gets the super long expires or not (returns a ``None`` or an extra boolean or something to specify if it's being cache busted or not).

mmerickel:
> Oh, and does this do away with the automatically setting a +10 year expiration on cache busted files, or does that still happen?

This PR removes any and all cache busting logic from the serving/matching side of static assets. If you want to add a longer expiration you can do so in a tween that detects requests with cache busting tokens.

mmerickel:
Tying things to the base asset spec seems unintuitive to me in almost every situation except this theme package we are trying to help you with. That's why I'm concerned that we're making it the default behavior.

I think I'd prefer it default to the top-level and only if you pass some option to `add_cache_buster` then you can set your own.

dstufft:
What would happen if you mix and match cache busters with and without said option?

dstufft:
> This PR removes any and all cache busting logic from the serving/matching side of static assets. If you want to add a longer expiration you can do so in a tween that detects requests with cache busting tokens.

So the argument for why I think ``add_static_view`` should still be smart about applying an expires header is because the obvious way is wrong when you add in asset overrides. The obvious way to add the long expires is using the ``cache_max_age`` parameter to the ``add_static_view`` configuration option. However that's going to do the wrong thing if an un-cachebusted file gets served via an asset override. You *can* add it using a tween, but then there's this convenient footgun (IMO) laying there in ``add_static_view`` that is going to just silently do the wrong thing when dealing with overrides.

mmerickel:
Yeah that's the part I'm struggling to enumerate.

My best answer would just be to allow the base-spec to win if there is one. Assuming the api is `add_cache_buster(cb, spec=None, base_spec=None)`. This basically means that if a theme package says `add_cache_buster(cb, base_spec='themepkg:static/')` then it would have priority over `add_cache_buster(cb, 'myapp:static/'). This isn't bad because the app can still override it by adding their own `add_cache_buster(cb, base_spec='themepkg:static/') if they want to but by default they are just specifying a high-level spec with lower priority.

dstufft:
>  However that's going to do the wrong thing if an un-cachebusted file gets served via an asset override. 

I'd like point out this isn't specific to the manifest cache buster either. Even if you're using a random value on each deployment like ``/static/foo.png?cachetoken=<blah>`` you still generally want ``/static/foo.png?cachetoken=<blah>`` to be cached for a long time, but you typically don't want ``/static/foo.png`` cached for a long time.

dstufft:
@mmerickel That seems like a more complicated way to just manually specify having it do something like:

1. Is there a cache buster for the base spec that matches this asset? If yes use that.
2. Otherwise is there a cache buster for the top level spec? If yes use that.
3. Else, do nothing and assume there is no cache busting for this URL.

mmerickel:
I don't see your expiration stuff as a concern because the assets will still receive a cache header like 3600 if you specify one. You can still control it via `add_static_view` for your app configuration. If you have some assets that are busted and some that are not then that's on you to deal with, and there are several easy ways to do it. 1) add all those assets to a subfolder with a new static view. 2) override specific assets in a tween like I suggested.  We voted in #2145 to remove the matching side.

mmerickel:
> Is there a cache buster for the base spec that matches this asset? If yes use that.
> Otherwise is there a cache buster for the top level spec? If yes use that.
> Else, do nothing and assume there is no cache busting for this URL.

I think that's exactly what I described. I went further and discussed how a user would fix it if they didn't want option 1 to have an effect.

mmerickel:
I believe it was even your idea to remove the matching side... That is the side that controls caching headers.

bertjwregeer:
> So the argument for why I think `add_static_view` should still be smart about applying an expires header is because the obvious way is wrong when you add in asset overrides.

This is no longer possible because we removed the `match()` API. As of right now the serving side knows nothing about what it is serving or that it is cache-busted or not. This means that currently we can't set the expires headers differently.

bertjwregeer:
Noticed too late that @mmerickel had already mentioned that.

dstufft:
Right. I got confused and forgot the static view is going to get the already cache busted URL not the original spec.  I blame the fact I've been working on moving all week and that frying my brain :). I rescind that statement because you're right it's  it possible to know if we're cache busted or not. A possible future thing might be to allow the cache_max_age take a callable that can be called to return and integer instead of just an integer. That way projects can still dynamically set it without having to use a tween.

mmerickel:
I'm going to switch up the behavior like I described and add an `explicit=False` option that people like @dstufft can set to `True` if they seriously want to tie the manifest to those exact files on disk.

mmerickel:
or `exact=True`, any one care about naming?!?!!

bertjwregeer:
exact over explicit.

mmerickel:
so far I like explicit because it lines up with `config.add_tween(..., explicit=True)`

dstufft:
``underlying=True``? ``overridden=True``? I dunno, neither exact or explicit really says to me that it's dealing with the real spec not the high level one.

mmerickel:
Interesting tidbit is that when dealing with the real spec it's possible it'll be an absolute path, not a package-relative spec. This won't happen when dealing with high level specs.

mmerickel:
`raw=True` ?

bertjwregeer:
Just pick one and go with it. We can bikeshed this to death otherwise.

docs/narr/assets.rst
```diff
@@ -376,14 +375,18 @@ assets by passing the optional argument, ``cachebust`` to
...
-   )
+   config.add_static_view(name='static', path='mypackage:folder/static/')
+   config.add_cache_buster(
+       'mypackage:folder/static/',
+       QueryStringConstantCacheBuster(str(int(time.time()))))
+
+.. note::
+   The trailing slash on the ``add_cache_buster`` call is important to
+   indicate that it is overriding every asset in the folder and not just a
+   file named ``static``.
```
bertjwregeer: Do we still allow a cache buster to be added for a single file? Also, is this really considered "overriding"? It looks like we are adding a cache buster for the current static view, not overriding anything.

pyramid/config/views.py
```diff
@@ -2102,3 +2100,86 @@ def register():
         intr['spec'] = spec
 
         config.action(None, callable=register, introspectables=(intr,))
+
+    def add_cache_buster(self, config, spec, cachebust, explicit=False):
+        if config.registry.settings.get('pyramid.prevent_cachebust'):
+            return
```
bertjwregeer: Should we move this check down into register() so that if you do config.add_cache_buster(), config.registry.settings['pyramid.prevent_cachebust'] = True it works the same way? I can imagine scenarios where a user may want to disable cache busting but may have already called config.include() or something along those lines.

docs/narr/assets.rst
mmerickel: No we don't allow it for a single file. You reminded me I need to update the docs. Thanks. Maybe just look at the api for now.

bertjwregeer:
I think it would be a good idea to add documentation with an example of the asset override, and registering a second cache buster for that.

Other than the comment above about moving the disabled check down to `register()` I am a +1.

pyramid/config/views.py
mmerickel: Great point, I will move the check.

bertjwregeer:
:ship: :+1:

docs/narr/assets.rst
```diff
@@ -536,25 +541,38 @@ CSS and JavaScript source and cache busting
...
-* Templatize JS and CSS, and call ``request.static_url()`` inside their
-  template code.
+Now that you are able to generate static URLs within :app:`Pyramid`,
+you'll need to handle URLs that are out of our control. To do this you may
+use some of the following options to get started:
 
-* Pass static URL references to CSS and JavaScript via other means.
+* Configure your asset pipeline to rewrite URL references inline in
+  CSS and JavaScript. This is the best approach because then the files
+  may be hosted by :app:`Pyramid` or an external CDN without haven't to
```
stevepiercy: s/haven't/having

docs/narr/assets.rst
```diff
@@ -793,3 +811,56 @@ when an override is used.
...
+:meth:`pyramid.config.Configurator.add_static_view` can affect your cache
+busting strategy when using any cache busters that are asset-aware such as
+:class:`pyramid.static.ManifestCacheBuster`. What sets asset-aware cache
+busters apart is that they have logic tied to specific assets. For example,
+a manifest is only generated for a specific set of pre-defined assets. Now,
+imagine you have overridden an asset defined in this manifest with a new,
+unknown version. By default, the cache buster will be invoked for an asset
+it has never seen before and will likely end up returning a cache busting
+token for the original asset rather than the asset that will actually end up
+being served! In order to get around this issue it's possible to attach a
```
stevepiercy: Insert comma:
issue, it's

docs/narr/assets.rst
```diff
@@ -793,3 +811,56 @@ when an override is used.
...
+   # override an asset
+   config.override_asset(
+       to_override='myapp:static/background.png',
+       override_with='theme:static/background.png')
+
+   # override the cache buster for theme:static assets
+   theme_cb = ManifestCacheBuster('theme:static/manifest.json')
+   config.add_cache_buster('theme:static', theme_cb, explicit=True)
+
+In the above example there is a default cache buster, ``my_cb`` for all assets
```
stevepiercy: insert comma:
buster, ``my_cb``, for

docs/narr/assets.rst
```diff
@@ -793,3 +811,56 @@ when an override is used.
...
+served from the ``myapp:static`` folder. This would also affect
+``theme:static/background.png`` when generating URLs via
+``request.static_url('myapp:static/background.png')``.
+
+The ``theme_cb`` is defined explicitly for any assets loaded from the
+``theme:static`` folder. Explicit cache busters have priority and thus
+``theme_cb`` would be invoked for
+``request.static_url('myapp:static/background.png')`` but ``my_cb`` would be
```
stevepiercy: insert comma:
background.png')``, but

stevepiercy:
@mmerickel excellent work. I added some line notes for you to consider in the docs.

mmerickel:
Good stuff @stevepiercy, thank you for reading through it!

Issue #2104, opened 2015-11-11
Documentation change: easy_install to pip


stevepiercy:
Update the documentation to use pip instead of easy_install.

Work should be pushed to the branch [docs/easy-install-to-pip.2104](https://github.com/Pylons/pyramid/tree/docs/easy-install-to-pip.2104).

Branch `docs/easy-install-to-pip.2104` will eventually be merged to `master` upon completion of the items listed below.

Add omitted items to the lists.

# List of related issues and PRs.
- https://github.com/Pylons/pyramid/issues/575
- https://github.com/Pylons/pyramid/issues/1454
- https://github.com/Pylons/pyramid/issues/121
- https://github.com/Pylons/pyramid/pull/2024

# List of documentation and official tutorials
(Pretty much every step for tutorials needs updating, so only the indices are shown for brevity.)
- [x] [Glossary](http://docs.pylonsproject.org/projects/pyramid/en/master/glossary.html)
- [x] [Quick Tour](http://docs.pylonsproject.org/projects/pyramid/en/master/quick_tour.html)
- [x] [Quick Tutorial](http://docs.pylonsproject.org/projects/pyramid/en/master/quick_tutorial/index.html)
- [x] [SQLAlchemy + URL Dispatch Wiki Tutorial](http://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/wiki2/index.html)
- [x] [ZODB + Traversal Wiki Tutorial](http://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/wiki/index.html)
- [x] [Running a Pyramid Application under mod_wsgi](http://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/modwsgi/index.html)
- [x] narr/commandline.rst
- [x] narr/extending.rst
- [x] narr/installation.html
- [x] narr/project.rst
- [x] narr/testing.rst

mmerickel:
A couple notes on how to do this:

1. `$VENV/bin/python setup.py develop` is now `$VENV/bin/pip install -e .` (the `.` on the end is important).
2. `$VENV/bin/python setup.py install` is now `$VENV/bin/pip install .

One big issue when updating to pip is that `python setup.py test` is no longer a great option for running tests. Specifically because the `tests_require` option will be installed using setuptools/easy_install instead of pip. We need to come up with a common pattern for this.

Two options (please chime in with some alternatives) I'm aware of that would work alright:

[...]

PR #2173, opened 2015-12-11
add developing and contributing section

stevepiercy:
- wrap to 79 columns


stevepiercy:
In reference to: https://github.com/Pylons/pyramid/issues/2104#issuecomment-163464303

PR #2176, opened 2015-12-11
route_prefix as an alternative

enomado:


bertjwregeer:
I'm sorry, what are you changing/adding and what are you attempting to document? It's in the wrong place, it should be placed above the `.. _virtual_root_support:` if it belongs with the previous section.

docs/narr/vhosting.rst
```diff
@@ -80,6 +80,8 @@ In the above configuration, we root a :app:`Pyramid` application at
 
 .. _virtual_root_support:
 
+Similar behaviour gives `route_prefix` `Configurator's` param. It will add prefix to all routes.
```
stevepiercy: * Do not split indices and labels from their section headings.
* We use American spellings. s/b "behavior".
* Wrap to 79 columns.
* Improve the grammar.

Aside from those issues, I do not understand what concept you are trying to express here.

enomado:
Thanks. I'm sorry about my first commit. It was impulsive.

btw if route_prefix is unrelated to this document its okay, I can ask about differences on stackoverflow. I just wanted to mention this option because it help me to do the same thing

stevepiercy:
@enomado you might be interested in these two documents http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#using-a-route-prefix-to-compose-applications and http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html?highlight=route_prefix

If either of those documents sound helpful to your use case, then to enhance your commit, I would suggest using rst syntax that provides a link to the relevant documentation section, like so:

```rst
:ref:`route_prefix`
```

enomado:
@stevepiercy, ```` :ref:`route_prefix` ```` is not the same thing. It goes to `using-a-route-prefix-to-compose-applications ` but my remark was about using it in Configurator(your second link). Is there something in sphinx to generate url to Configurator class and highlight=route_prefix ?

stevepiercy:
```
:mod:`pyramid.config`
```

See Sphinx docs about roles: http://sphinx-doc.org/domains.html?#role-py:mod

stevepiercy:
What errors do you see when building the docs? I have a difficult time reading your screen from here. :stuck_out_tongue_winking_eye: 

Did you follow the instructions in [HACKING.txt](https://github.com/Pylons/pyramid/blob/master/HACKING.txt) for building docs?

enomado:
@stevepiercy oh I've tried to build docs by `make html` from docs folder. There were errors about not existing classes. Its strange but tox do the same thing. Anyway thanks. Its fine now. Thanks

stevepiercy:
Circling back to this, there is no clear `route_prefix` section in the API docs for this parameter, so we cannot generate a named anchor link to it. The best we can do is:

```
:class:`pyramid.config.Configurator`
```

I also tried the experimental plugin sphinx-paramlinks, too, but it seemed not to add much functionality above and beyond what exists in Sphinx 1.3.1.

stevepiercy:
@enomado would you please correct the grammar, then add a code example, as follows:

```rst
An alternative approach is to add a prefix to all routes by using the
`route_prefix` parameter in :class:`pyramid.config.Configurator`. For
example::

    example
```

Finally, if you have not yet done so, please sign https://github.com/Pylons/pyramid/blob/master/CONTRIBUTORS.txt

Other than those items, I think it's fine to add this note here, unless there is something I overlooked.

Thank you!

mmerickel:
This doesn't seem to be going anywhere. Gonna close this.

PR #2178, opened 2015-12-12
Update url.py

srisankethu:
Just made a few changes like in line 104,105 is nested under  if statement in line 102

if port is not None: statement in line 100 is added to else: statement in line 111

Combined url=... from line 99 and 120

### DIFF
--- a/pyramid/url.py
+++ b/pyramid/url.py
@@ -98,6 +98,8 @@ class URLMethodsMixin(object):
                     port = '80'
         if host is None:
             host = e.get('HTTP_HOST')
+            if host is None:
+                host = e['SERVER_NAME']
         if port is None:
             if ':' in host:
                 host, port = host.split(':', 1)
--- a/pyramid/url.py
+++ b/pyramid/url.py
@@ -98,8 +98,6 @@ class URLMethodsMixin(object):
                     port = '80'
         if host is None:
             host = e.get('HTTP_HOST')
-            if host is None:
-                host = e['SERVER_NAME']
         if port is None:
             if ':' in host:
                 host, port = host.split(':', 1)
--- a/pyramid/url.py
+++ b/pyramid/url.py
@@ -96,19 +96,17 @@ class URLMethodsMixin(object):
             if scheme == 'http':
                 if port is None:
                     port = '80'
-        url = scheme + '://'
-        if port is not None:
-            port = str(port)
         if host is None:
             host = e.get('HTTP_HOST')
-        if host is None:
-            host = e['SERVER_NAME']
+            if host is None:
+                host = e['SERVER_NAME']
         if port is None:
             if ':' in host:
                 host, port = host.split(':', 1)
             else:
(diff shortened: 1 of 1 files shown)


sontek:
Can you add a test to show why this change was necessary?

srisankethu:
I was just organising the code.
Nested if on line 104 looks meaningful
No need to update url twice(line 99 and line 120) as there is no change in it's value between line 99 and line 120
Unnecessary if port is None: on line 100 as else statement on line 110 is the same.

pyramid/url.py
```diff
@@ -96,19 +96,17 @@ def _partial_application_url(self, scheme=None, host=None, port=None):
...
-            host = e['SERVER_NAME']
+            if host is None:
+                host = e['SERVER_NAME']
         if port is None:
             if ':' in host:
                 host, port = host.split(':', 1)
             else:
                 port = e['SERVER_PORT']
         else:
+            port = str(port)
```
stevepiercy: Why is this line necessary? Please provide a test to prove its necessity.

pyramid/url.py
```diff
@@ -96,19 +96,17 @@ def _partial_application_url(self, scheme=None, host=None, port=None):
...
                     port = '80'
-        url = scheme + '://'
-        if port is not None:
-            port = str(port)
         if host is None:
             host = e.get('HTTP_HOST')
-        if host is None:
-            host = e['SERVER_NAME']
+            if host is None:
+                host = e['SERVER_NAME']
```
stevepiercy: This is OK with me. With the nesting, I read this implicitly as "if host is *still* None, then...".

pyramid/url.py
```diff
@@ -117,7 +115,7 @@ def _partial_application_url(self, scheme=None, host=None, port=None):
         elif scheme == 'http':
             if port == '80':
                 port = None
-        url += host
+        url = scheme + '://' + host
```
stevepiercy: Condensing this url assignment into one line from the previous two lines (old lines 99 and 120) is cleaner. :+1:

pyramid/url.py
srisankethu: That was already a part of your code.I just rearranged it's place

stevepiercy: Ah, I see now. I'm OK with this change, too.

stevepiercy:
If one of the core contributors gives a :+1: on this PR, then I'll merge it, backport it to 1.6 and 1.5, and make a change note.

@srisankethu would you please sign CONTRIBUTORS.txt, and push it to this PR, too? You don't need to delete what you've already done.

mmerickel:
There's no need to backport this change, it's just a cleanup versus adding any actual features.

mmerickel:
Anyway, +1, I'll merge this and add a changelog entry. Thanks!

mmerickel:
I lied... need you to add your name to CONTRIBUTORS.txt first. :-)

srisankethu:
Ha....Thanks man

Issue #1385, opened 2014-07-16
remove pyramid.reload_assets setting


mmerickel:
As discovered in #1380 it is not actually used for anything in the codebase.

mmerickel:
I did some more digging on this and it was used in the codebase prior to the removal of mako/chameleon from core. The setting is still used in the pyramid_chameleon package. There is also a bunch of documentation[1] I never read before explaining its intended purpose. Now I'm not sure what to do.

Further thinking, it seemed like it would be interesting to support some sort of caching of `override_asset` such that it didn't need to go through the asset search mechanism for a file every time. The mapping could use a LRU cache when `reload_assets == False`. Anyway maybe it's harmless to leave the setting in and use it in the future.

[1] http://docs.pylonsproject.org/projects/pyramid/en/1.4-branch/narr/environment.html#understanding-the-distinction-between-reload-templates-and-reload-assets

mmerickel:
This setting is still depended on by `pyramid_chameleon`[1] as well. @bertjwregeer or @mcdonc can you add some thoughts on what to do here?

[1] https://github.com/Pylons/pyramid_chameleon/blob/dd1f4cfa913da7e6af5d6189d613aed279dad761/pyramid_chameleon/renderer.py#L123

bertjwregeer:
I just helped rip that code base out of Pyramid itself, I am not at all familiar with the pyramid_chameleon renderer, it would take me some time to understand what it is doing and why.

mmerickel:
Okay, judgement call. It was premature to want to remove the setting as it serves a real, documented, purpose and is being used by pyramid_chameleon for that purpose.

PR #2219, opened 2016-01-08
Remove pyramid.reload_assets

marioidival:
#1385 

While moving, I noticed that there was another key named `reload_resources`, which is a alias for `reload_assets`...

let me know if has any errors.


mmerickel:
Sorry @marioidival but as I said in #1385 I think it was a mistake to try and remove this setting.

PR #2227, opened 2016-01-13
expect py3 and special-case py2 behavior

mmerickel:
previously py3 was the special case, but as we move toward the future we
want the py2 codebase to be considered legacy with py3 as the default

### DIFF
--- a/pyramid/compat.py
+++ b/pyramid/compat.py
@@ -20,7 +20,7 @@ try:
 except ImportError:  # pragma: no cover
     import pickle
 
-# True if we are running on Python 3.
+# PY3 is left as bw-compat but PY2 should be used for most checks.
 PY2 = sys.version_info[0] == 2
 PY3 = sys.version_info[0] == 3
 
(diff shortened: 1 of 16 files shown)


stevepiercy:
This is a good idea. Off the top of my head, I can't think of any Python version specific docs that require updating.

tseaver:
LGTM

bertjwregeer:
While I don't disagree with the change, I am wondering if this actually changes any aspect of the runtime efficiency? Does Python do any optimizations based upon which side of the branch it needs to take?

The code change looks good to me.

mmerickel:
Almost all of the changes are import-time. As far as if/else efficiency I've never heard of python attempting to do any branch prediction on top of what the CPU would be doing itself.

abdulmuneer:
if there is any explicit version check for python 3 (like sys.version[0]==3), it is going to execute python 2 code when python version reaches 4.

bertjwregeer:
Before that happens, we will have deprecated Python 2.x.

Issue #2255, opened 2016-01-20
pip no longer installs in Python 3.2


bertjwregeer:
This is currently breaking tox test runs. Dropping Python 3.2 support may be required.

bertjwregeer:
https://travis-ci.org/Pylons/pyramid/jobs/103424405

mmerickel:
yes, I'm +1 on dropping python versions as the PSF does as well

mmerickel:
I'll say I'm not interested in backporting this change to older Pyramid versions however. I think we should be stuck figuring out how to handle 1.5/1.6 (extra motivation to release 1.7/1.8).

stevepiercy:
@mmerickel which versions do you consider to be "older Pyramid versions"?

mmerickel:
Currently 1.5/1.6 like I said. We consider the last 2 releases supported.

mmerickel:
Fixed via #2256, #2269, #2270, #2271.

PR #2256, opened 2016-01-20
Remove Python 3.2 support

bertjwregeer:
This removes Python 3.2 support from the testing path, and from setup.py. At this point testing it becomes more difficult because pip no longer installs/runs under Python 3.2, and tox pulls in the latest version of pip.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ matrix:
           env: TOXENV=py26
         - python: 2.7
           env: TOXENV=py27
-        - python: 3.2
-          env: TOXENV=py32
         - python: 3.3
           env: TOXENV=py33
         - python: 3.4
--- a/docs/narr/install.rst
+++ b/docs/narr/install.rst
@@ -15,8 +15,8 @@ You will need `Python <http://python.org>`_ version 2.6 or better to run
 .. sidebar:: Python Versions
 
     As of this writing, :app:`Pyramid` has been tested under Python 2.6, Python
-    2.7, Python 3.2, Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3.
-    :app:`Pyramid` does not run under any version of Python before 2.6.
+    2.7, Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3. :app:`Pyramid`
+    does not run under any version of Python before 2.6.
 
 :app:`Pyramid` is known to run on all popular UNIX-like systems such as Linux,
 Mac OS X, and FreeBSD as well as on Windows platforms.  It is also known to run
--- a/docs/narr/introduction.rst
+++ b/docs/narr/introduction.rst
@@ -860,11 +860,11 @@ Every release of Pyramid has 100% statement coverage via unit and integration
 tests, as measured by the ``coverage`` tool available on PyPI.  It also has
 greater than 95% decision/condition coverage as measured by the
 ``instrumental`` tool available on PyPI.  It is automatically tested by the
-Jenkins tool on Python 2.6, Python 2.7, Python 3.2, Python 3.3, Python 3.4,
-Python 3.5, PyPy, and PyPy3 after each commit to its GitHub repository.
-Official Pyramid add-ons are held to a similar testing standard.  We still find
-bugs in Pyramid and its official add-ons, but we've noticed we find a lot more
-of them while working on other projects that don't have a good testing regime.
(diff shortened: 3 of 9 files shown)


bertjwregeer:
Closes https://github.com/Pylons/pyramid/issues/2255

Not sure if we want to backport this to 1.5/1.6, but if we don't we need to figure out how install an older version of pip in the Python 3.2 tox environment.

stevepiercy:
This PR also needs to include updates to docs. I'm working on those now.

stevepiercy:
Docs updated for Pyramid, except for one mention of Waitress running on 3.2+ in `docs/glossary.rst`.

stevepiercy:
From https://www.python.org/dev/peps/pep-0392/#lifespan

> ...security updates (source only) will be released until 5 years after the release of 3.2 final, which will be February 2016.

I think this should definitely be merged into master for Pyramid 1.7 release.

Since docs for 1.6/latest is the version most hit from search engine results, and pip is where everyone is headed, I'd say backport to 1.6.

For 1.5, I'd say remove testing for 3.2 to get it to pass, but we leave the docs as is to indicate that it runs on 3.2.

Also for future thought, we might want to declare our plan for support of Python 3.3, which is scheduled to cease receiving security updates in September 2017:
https://www.python.org/dev/peps/pep-0398/#lifespan

bertjwregeer:
I've added the 1.7 milestone for this. We should definitely add a document somewhere to declare our intentions and how long we will support Python versions. Python 2.6 is kind of a special snowflake, because of CentOS 6.x. It was last released in 2008!!!

bertjwregeer:
@mmerickel I am going to answer your questions on the original ticket here:

https://github.com/pypa/packaging/pull/58

We can use this for Python 3.2 support on Travis, this way we can continue building on travis for 1.5/1.6 branches. Local building becomes a HUGE pain.

mmerickel:
I'm much less worried about the complexity of building locally for backports. Anyone who is actually stuck building those versions probably has a machine where it's easier to operate in that enivronment. All backports should be tested via travis using such a fix. Any new features belong on master where we drop 3.2 and local testing is not difficult.

stevepiercy:
Is it OK to merge this PR onto master, or is there something else we need to do or are waiting for?

Here's a statement of support of Python, although I don't know where it should go. It might also be a Pylons Project policy.

# Python Support Policy

At the time of each release of Pyramid, each version supports all versions of Python to the end of their lifespans. The end-of-life for a given version of Python is when security updates are no longer released.

* [Python 3.2 Lifespan](https://www.python.org/dev/peps/pep-0392/#lifespan) ends February 2016.
* [Python 3.3 Lifespan](https://www.python.org/dev/peps/pep-0392/#lifespan) ends September 2017.
* [Python 3.4 Lifespan](https://www.python.org/dev/peps/pep-0429/) TBD.
* [Python 3.5 Lifespan](https://www.python.org/dev/peps/pep-0478/) TBD.
* [Python 3.6 Lifespan](https://www.python.org/dev/peps/pep-0494/#id4) December 2021.

# Python Support in Pyramid
## Pyramid 1.5
* Python 2.6
* Python 2.7
* Python 3.2
* Python 3.3
* Python 3.4
* Python 3.5
* PyPy
* PyPy3

## Pyramid 1.6
* Python 2.6
* Python 2.7
* Python 3.2
* Python 3.3
* Python 3.4
* Python 3.5
* PyPy
* PyPy3

## Pyramid 1.7
* Python 2.6
* Python 2.7
* Python 3.3
* Python 3.4
* Python 3.5
* PyPy
* PyPy3

bertjwregeer:
I don't want to merge this myself. I'm leaving it to @mmerickel. We need to apply the fix above for the 1.6-branch and 1.5-branch so that we can continue to build Python 3.2 on Travis for those branches.

mmerickel:
> We need to apply the fix above for the 1.6-branch and 1.5-branch so that we can continue to build Python 3.2 on Travis for those branches.

Can you add PRs that build for those branches?

bertjwregeer:
Yes, when I get back Monday. I am currently out at SCALE14X.

bertjwregeer:
#2269 #2270

stevepiercy:
For "Python Support Policy", how about placing the content here?
http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/upgrading.html#deprecation-and-removal-policy

I also modified the content to point to each branch's tox.ini to find Python support. Much cleaner.

mmerickel:
@stevepiercy that seems like a logical place.

PR #2257, opened 2016-01-20
Do not use trailing . on versionadded rst directive. It automatically…

stevepiercy:
… adds a .


stevepiercy:
OK to merge, see https://github.com/Pylons/pyramid/pull/2256

PR #2270, opened 2016-01-22
1.5-branch: Fix building Py32 on Travis

bertjwregeer:


mmerickel:
Only had to tell travis to re-build this 3 times I think before it worked. Some stale state in the VM or something?

bertjwregeer:
It looked like the Travis infrastructure was having some issue in general, like docker containers weren't being started correctly.

PR #2283, opened 2016-01-24
add rstcheck for docs

uralbash:
@stevepiercy 

Added rst linter to ``docs/Makefile``. I think it might help to look for mistakes, especially if it is added to travis-ci.
It works with https://github.com/myint/rstcheck from master branch.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -11,8 +11,10 @@ BUILDDIR      = _build
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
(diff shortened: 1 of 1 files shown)


mmerickel:
travis-ci already builds the docs (with warnings turned into errors) every time. What benefit does this add?

uralbash:
@mmerickel Yes, you're right, it seems ``sphinx-build`` is doing the same thing

stevepiercy:
@mmerickel @uralbash This would have been helpful before the release of Sphinx 1.3.5. See https://github.com/sphinx-doc/sphinx/issues/2264 Now I think this would be redundant in regards to syntaxes.

The rst heading over/underline length mismatches is helpful, but I think that is better handled in the editor (PyCharm's rst syntax checker is very helpful). travis-ci used to miss rst heading over/underline length mismatches, but I haven't tried checking that in a while because I pretty much snuffed them all out. Ultimately the sphinx-build for html forgives this syntax error and renders such headings just fine.

PR #2298, opened 2016-01-28
ensure csrf tokens are compared as bytes

mmerickel:


mmerickel:
@bertjwregeer this is ready for review

bertjwregeer:
I'm good on this.

PR #2328, opened 2016-02-02
small typo

FooBarQuaxx:


stevepiercy:
The current phrase is correct, not a typographical error.

PR #2376, opened 2016-02-26
Add ApplicationCreated.config variable.

wichert:
I have some common code which is used in multiple applications. One of things I would like to do there is setup an event handler that logs application startup and dump a bunch of used package versions. To make that more useful I would like to know which application was started. I can get that using ``config.package``, but that is not accessible through the `ApplicationCreated` event.

### DIFF
--- a/pyramid/events.py
+++ b/pyramid/events.py
@@ -156,13 +156,19 @@ class ContextFound(object):
 AfterTraversal = ContextFound # b/c as of 1.0
 
 @implementer(IApplicationCreated)
-class ApplicationCreated(object):    
+class ApplicationCreated(object):
     """ An instance of this class is emitted as an :term:`event` when
     the :meth:`pyramid.config.Configurator.make_wsgi_app` is
-    called.  The instance has an attribute, ``app``, which is an
-    instance of the :term:`router` that will handle WSGI requests.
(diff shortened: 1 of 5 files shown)


pyramid/events.py
```diff
@@ -156,13 +156,19 @@ def __init__(self, request):
...
-class ApplicationCreated(object):    
+class ApplicationCreated(object):
     """ An instance of this class is emitted as an :term:`event` when
     the :meth:`pyramid.config.Configurator.make_wsgi_app` is
-    called.  The instance has an attribute, ``app``, which is an
-    instance of the :term:`router` that will handle WSGI requests.
-    This class implements the
-    :class:`pyramid.interfaces.IApplicationCreated` interface.
+    called. The instance has two attribuets:
+
```
stevepiercy: spelling: attributes

mmerickel:
The configurator is a throw-away object that is not intended to have any gettable state on it. There are actually many config objects created (one per call to `config.include(..)` for example). The shared state is the `registry` which is already passed here via `event.app.registry`. Possibly it could be improved by making `event.registry` available but whatever, it's there.

From the registry you can use `registry.package_name` to get the root package that the top-level configurator was created from.

Issue #2368, opened 2016-02-23
Drop Python 2.6 support


tseaver:
Per [Nick Coghlan's blog post](http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html).

tl;dr: folks still running Python 2.6 can't reasonably expect to be updating to latest-and-greatest of everything else, especially if they aren't paying for it with blood or treasure.

bertjwregeer:
For 1.7 we could totally just drop it.

mmerickel:
I have no problem dropping 2.6 considering it was EOL 2 years ago. The only argument against that I know are 1) according to @dstufft's numbers there are still tons of installations of 2.6 using PyPI 2) it's not hard to support - at least so far.

Next question: soft or hard drop? The current `setup.py` raises a `RuntimeError` on < 2.6 because Pyramid uses things like the `with` statement that don't work on 2.5 whereas 2.6 feels more like "it may work or it may not, but we aren't testing it anymore".

mmerickel:
Actually I think the `with` statement is 2.5+... Maybe byte strings are a better example. Point remains.

bertjwregeer:
We should remove it from setup.py, besides that, let's keep testing on Python 2.6 until we decide it's time to drop it.

mmerickel:
Sorry what do you mean "remove it from setup.py" ? The trove classifier? A hard fail?

bertjwregeer:
Remove it from the trove classifier.

tseaver:
ISTM that dropping it from tox/Travis should go hand-in-hand with dropping it from Trove:  we get quicker turnarounds for builds, plus we can start using [2.7-only features / syntax](https://docs.python.org/2/whatsnew/2.7.html), e.g. improved unittest assertions, set literals, dict comprehensions, etc.  I'm +0 to dropping it for 1.7 (since we haven't deprecated it ahead) and +1 for whatever release comes afterward (1.8 or 2.0).

dstufft:
I personally think Pyramid can probably ditch Python 2.6, but I'll let y'all make that decision. However, data is cool, so here is the downloads for Pyramid from 2016-02-22:

| python_version | download_count |
|----------------|----------------|
| 2.7            | 1855           |
| 3.4            | 270            |
| 3.5            | 175            |
| 2.6            | 123            |
| Unknown  | 75             |
| 3.3            | 67             |
| 3.2            | 8              |

dstufft:
[...]

PR #2379, opened 2016-02-27
wiki2 authentication bug fix and improvement against timing attack

stevepiercy:
ping @mmerickel @bertjwregeer for review and comment (thanks to @ztane @Themanwithoutaplan and @wichert for IRC help)

- Bytes type does not have encode method. The `expected_hash` retrieved from the database is a bytes object.
- Use `hmac.compare_digest()` instead of `==` to avoid timing attacks as a recommended security best practice. See https://www.python.org/dev/peps/pep-0466/ https://bugs.python.org/issue21306 and https://codahale.com/a-lesson-in-timing-attacks/ for details.

  Note, however, `hmac.compare_digest()` was not backported to py2.6. For a tutorial, I am OK with stating this will not work on Python 2.6 with a clear warning note at the start of the tutorial and on the authentication step.

### DIFF
--- a/docs/tutorials/wiki2/src/authentication/tutorial/models/user.py
+++ b/docs/tutorials/wiki2/src/authentication/tutorial/models/user.py
@@ -1,5 +1,4 @@
 import bcrypt
-import hmac
 from sqlalchemy import (
     Column,
     Integer,
@@ -26,5 +25,5 @@ class User(Base):
         if self.password_hash is not None:
             expected_hash = self.password_hash
             actual_hash = bcrypt.hashpw(pw.encode('utf8'), expected_hash)
-            return hmac.compare_digest(expected_hash, actual_hash)
+            return expected_hash == actual_hash
         return False
--- a/docs/tutorials/wiki2/src/authentication/tutorial/models/user.py
+++ b/docs/tutorials/wiki2/src/authentication/tutorial/models/user.py
@@ -1,4 +1,5 @@
 import bcrypt
+import hmac
 from sqlalchemy import (
     Column,
     Integer,
@@ -23,7 +24,7 @@ class User(Base):
 
     def check_password(self, pw):
         if self.password_hash is not None:
-            expected_hash = self.password_hash.encode('utf8')
+            expected_hash = self.password_hash
             actual_hash = bcrypt.hashpw(pw.encode('utf8'), expected_hash)
-            return expected_hash == actual_hash
+            return hmac.compare_digest(expected_hash, actual_hash)
         return False
--- a/docs/tutorials/wiki2/src/authorization/tutorial/models/user.py
+++ b/docs/tutorials/wiki2/src/authorization/tutorial/models/user.py
@@ -23,7 +23,7 @@ class User(Base):
 
     def check_password(self, pw):
         if self.password_hash is not None:
-            expected_hash = self.password_hash.encode('utf8')
+            expected_hash = self.password_hash
             actual_hash = bcrypt.hashpw(pw.encode('utf8'), expected_hash)
             return expected_hash == actual_hash
         return False
--- a/docs/tutorials/wiki2/src/models/tutorial/models/user.py
+++ b/docs/tutorials/wiki2/src/models/tutorial/models/user.py
@@ -23,7 +23,7 @@ class User(Base):
 
     def check_password(self, pw):
         if self.password_hash is not None:
-            expected_hash = self.password_hash.encode('utf8')
+            expected_hash = self.password_hash
             actual_hash = bcrypt.hashpw(pw.encode('utf8'), expected_hash)
             return expected_hash == actual_hash
         return False
--- a/docs/tutorials/wiki2/src/tests/tutorial/models/user.py
+++ b/docs/tutorials/wiki2/src/tests/tutorial/models/user.py
@@ -23,7 +23,7 @@ class User(Base):
 
     def check_password(self, pw):
         if self.password_hash is not None:
-            expected_hash = self.password_hash.encode('utf8')
+            expected_hash = self.password_hash
             actual_hash = bcrypt.hashpw(pw.encode('utf8'), expected_hash)
             return expected_hash == actual_hash
         return False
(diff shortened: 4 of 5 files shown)


stevepiercy:
Oh, and if approved, I will do the necessary work on the docs:
- [ ] add references to avoiding timing attack best practices in authentication
- [ ] carry forward changes in `user.py` to later tutorial steps

Themanwithoutaplan:
FWIW Python 2.6 probably shouldn't be in use on any servers anyway because it's not getting security updates.

mmerickel:
@bertjwregeer can back me up on this but there is no evidence that timing attacks are possible against password hashing algorithms (at least bcrypt) because the string comparison is on the resulting hash itself. Any change to the password results in a completely new hash which would take longer or shorter to compare thus it's infeasible to use timing as an attacker to progressively guess more and more characters of the string. [1]

This also breaks py26 and some versions of py27 (not gonna happen).

As far as the bytes vs unicode issue - I was previously testing on py27 so didn't notice this. Just spun up a virtualenv on py35 and reproduced the issue. If you want to submit that fix in a separate PR I will merge it.

[1]  http://security.stackexchange.com/questions/46212/does-bcrypt-compare-the-hashes-in-length-constant-time

stevepiercy:
@mmerickel @bertjwregeer I could be wrong, but my understanding is that bcrypt provides a constant-length, but when comparing the two hashes, we still need to use hmac or XOR to avoid the timing attack. See the last two sections "Why does the hashing code on this page compare the hashes in 'length-constant' time?" and "How does the SlowEquals code work?"
https://crackstation.net/hashing-security.htm

While we're discussing the topic, how come we don't use a salt? I could add that, too, and make the necessary updates to the db init script, etc.

Finally, advocating best security practices is not consistent with supporting versions of Python that don't receive security updates. I'm not sure what we should do in this regard, but at least mentioning implications would be the responsible thing to do.

mmerickel:
>  I could be wrong, but my understanding is that bcrypt provides a constant-length, but when comparing the two hashes, we still need to use hmac or XOR to avoid the timing attack.

There's a lot of crud floating around about storing passwords given "best practices" are basically re-learned by every developer ever who chooses to blog about it. I'm not saying your link is wrong, and there are benefits to constant-time comparisons in many situations however bcrypt makes these attacks practically quite unfeasible. See the comments in https://github.com/pyca/bcrypt/issues/8 and https://github.com/pyca/bcrypt/pull/60 for examples. If the bcrypt folks in PyCA say its unfeasible I'm liable to believe them. If they decide to add a comparison function to the library as one of those issues discusses, then we should update our example to use it.

> While we're discussing the topic, how come we don't use a salt?

We are using a salt. Note the `bcrypt.gensalt()` used in the `set_password` method. You can use your own application-level salts and such but they provided only limited benefit over per-password salts.

> Finally, advocating best security practices is not consistent with supporting versions of Python that don't receive security updates.

We've already started the discussion to deprecate Python 2.6 in #2368, however we cannot stop supporting any versions of Python 2.7 regardless of their state as they are actively being bundled by OSs and are not aggressively updated to patch releases.

stevepiercy:
> If the bcrypt folks in PyCA say it's unfeasible I'm liable to believe them.

OK, I can go along with that discussion.

> We are using a salt. Note the bcrypt.gensalt() used in the set_password method. You can use your own application-level salts and such but they provided only limited benefit over per-password salts.

OK, I was looking at the db schema only, not the code, and wondered why there weren't per-password salts. I've done that in my apps outside of Python.

Given the foregoing, my final point is moot and I'll pull it from the table.

I'll revert the hmac stuff, and roll the source forward in later steps. Thanks!

mmerickel:
> OK, I was looking at the db schema only, not the code, and wondered why there weren't per-password salts.

The bcrypt result we store as `password_hash` is actually an ascii string containing several of the input parameters along with the resulting hash. That's why `bcrypt.hashpw(pw.encode('utf8'), expected_hash)` works. It is borrowing the input parameters (salt and work factor) stored in `expected_hash` and using them to hash the new `pw`.

bertjwregeer:
You do not need to have a constant time compare for password hashes.

bertjwregeer:
Oh $DEITY, that site https://crackstation.net/hashing-security.htm is terrible.

bertjwregeer:
@stevepiercy

Constant time compare is needed when an attacker can manipulate the hash being compared. For example, signed cookies. The cookie is actually <value><hash (aka signature)>. It's secure so long as the attacker can't modify <value> and generate a valid <hash>. However <hash> is only X amount of bytes long. The attacker can modify the <hash> byte by byte, and if it isn't using a constant time compare, use timing information to derive a valid <hash> for their chosen <value>.

In the case of passwords, the hash is stored server side, and the client sends up plaintext information. The attacker can't modify any hashes. Server side we then take the salt, and bcrypt and out comes a hash. We can then compare that hash against the one we had in our database and see if they match. Let's say they don't, all an attacker gleans is that the password is invalid, as soon as they change a byte in the plaintext input, the avalanche property of a good hash function will cause the output hash to change so significantly that an attacker won't glean anything and has to bruteforce the full password space.

Avalanche in action:
Password1 == 2ac9cb7dc02b3c0083eb70898e549b63
password1 == 7c6a180b36896a0a8c02787eeafb0e4c

stevepiercy:
@bertjwregeer thanks, that makes sense.

@mmerickel I made the updates, and merged the PR. I'll resume work on `feature/alchemy-scaffold-update` branch, including the last two pages of docs, the feature to redirect when adding a page that already exists feature, and circle back to spruce up the affected docs.

Themanwithoutaplan:
@bertjwregeer am I correct in thinking that `bcrypt.hashpw(…)` works in constant time? Seems to me this would be the relevant call for a timing attack. Comparing two strings of the same length should have minimal variance as the stackexchange article indicates.

bertjwregeer:
I have no idea if that is constant time or not, and even if it isn't constant time, nothing is lost.

There is no timing attack in comparison of password hashes.

ztane:
@stevepiercy notice also that the *salt* is not revealed to the attacker, so the attacker cannot even know the hashes of their chosen passwords.

PR #2392, opened 2016-03-02
Note's link is dead

Slevender:
The link in note does not give an expected result.

I also would like to thank everyone who have contributed to this amazing tutorial. Having a blast learning!

### DIFF
--- a/docs/quick_tutorial/routing.rst
+++ b/docs/quick_tutorial/routing.rst
@@ -30,7 +30,7 @@ Previously we saw the basics of routing URLs to views in Pyramid.
     be explicit in ordering. Pyramid also gives facilities to avoid the
     problem.  It's relatively easy to build a system that uses implicit route
     ordering with Pyramid too.  See `The Groundhog series of screencasts 
-    !<http://bfg.repoze.org/videos#groundhog1>`_ if you're interested in
+    <http://bfg.repoze.org/videos#groundhog1>`_ if you're interested in
     doing so.
 
 Objectives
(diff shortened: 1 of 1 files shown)


stevepiercy:
@mcdonc @pauleveritt or @tseaver http://bfg.repoze.org/videos is dead. Is there a new URL for the videos, or are they lost forever?

@Slevender the .rst syntax in this PR is bad. Make sure you can build the docs locally before submitting a PR.

Slevender:
@stevepiercy Okay. Sorry. I'm new to this stuff. Will try to be more careful next time.

bertjwregeer:
There are no changes in this PR, should this to be a ticket instead?

mmerickel:
The link and reference simply needs to be removed. I'm unaware of an alternate location of this content we could link to.

stevepiercy:
@Slevender no problem. For instructions on how to build the docs, see https://github.com/Pylons/pyramid/blob/master/HACKING.txt

@bertjwregeer there is a `!` which causes the docs not to build.

bertjwregeer:
@stevepiercy He added another commit that reverts that addition of a !. This makes this PR a no-op. Hence my question as to wether this would be better as a standard ticket and not a PR.

tseaver:
We took down `bfg.repoze.org` because it was attracting abuse.  The videos are actually hosted at:

  http://static.repoze.org/casts/videotags.html

stevepiercy:
I'll take this on tonight.

stevepiercy:
@tseaver I could only get the second video to play in Chrome. I didn't try other browsers.

PR #2393, opened 2016-03-03
request.invoke exception view

mmerickel:
Introduce a new API for invoking exception views manually. This can be done for several reasons but the primary use-case is for error handling inside of tweens where it is desirable to utilize the registered exception views of the pyramid app to render errors instead of defining some other hook for rendering the error as is required when using wsgi middleware.


bertjwregeer:
Much excitement =)

:+1: from me. I just gave it a cursory glance though.

mmerickel:
Hey @tseaver I'm about to merge this, do you have any final objections?

tseaver:
I guess I'm -0 on checking anything in that leaves the traceback-local binding around until GC kicks in.  Not enough to block it (and we may be doing it elsewhere), but it still makes me uncomfortable.

mmerickel:
AFAICT based on https://github.com/Pylons/pyramid/pull/2393#commitcomment-16464132 there should be no need for GC here as there are no circular refs. The only possible issue in this api is if a user themselves calls `sys.exc_info()` and passes it to me but at that point they're responsible for handling that - I can't stop them.

PR #2397, opened 2016-03-04
configurable view deriver tweaks

mmerickel:
I put some work into #2021 tonight and here's what I've come up with:

- New api for view derivers is `deriver(view, info)` where info is a new `IViewDeriverInfo` object with attributes including `package`, `registry`, `orig_view`, `predicates` and `options`.
- Define `options` as almost everything that was passed to `config.add_view` and a few other bonuses like `phash`, `order`, etc. We'll need to document this.
- Removed all default values from view derivers.

TODO

- Add a new `add_view_option` API that supports setting default values. View options will be the way to provide default values for options passed to `config.add_view`. This could be used for predicates values or view deriver values so it would then be possible to have a predicate default to on as well. Envision `config.add_view_option('require_csrf', default=True)`.

### DIFF
--- a/pyramid/config/derivations.py
+++ b/pyramid/config/derivations.py
@@ -154,8 +154,8 @@ class DefaultViewMapper(object):
 
 
 def wraps_view(wrapper):
-    def inner(view, value, **kw):
-        wrapper_view = wrapper(view, value, **kw)
+    def inner(view, info):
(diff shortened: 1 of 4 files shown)


bertjwregeer:
:+1: This looks great, and I like the idea of having an object with stuff on it that is well defined.

Also like the `add_view_option` idea.

PR #2398, opened 2016-03-07
Pass os.environ to logging.config.fileConfig

msabramo:
This allows one to set up a logging configuration that is parameterized based on environment variables, and thus in the spirit of http://12factor.net/config

e.g.: the application .ini file could have:

```ini
[logger_root]
level = %(LOGGING_LOGGER_ROOT_LEVEL)s
handlers = console

[handler_console]
class = StreamHandler
args = (sys.stderr,)
level = %(LOGGING_HANDLER_CONSOLE_LEVEL)s
formatter = generic
```

Log levels are one of the main things that differ in a `development.ini` vs. a `production.ini`. I am working towards the goal of having only one config file for a Pyramid app, because having two configs is error-prone and implies that there are only two possible configurations for an app, which is somewhat limiting. Again, this is inspired largely by http://12factor.net/config

Cc: @mmerickel, @sontek, @sudarkoff, @sseg, @aconrad

### DIFF
--- a/pyramid/paster.py
+++ b/pyramid/paster.py
@@ -67,10 +67,12 @@ def setup_logging(config_uri, fileConfig=fileConfig,
     parser.read([path])
     if parser.has_section('loggers'):
         config_file = os.path.abspath(path)
-        return fileConfig(
-            config_file,
-            dict(__file__=config_file, here=os.path.dirname(config_file))
+        vars = dict(
+            __file__=config_file,
+            here=os.path.dirname(config_file)
             )
+        vars.update(os.environ)
+        return fileConfig(config_file, vars)
 
 def _getpathsec(config_uri, name):
     if '#' in config_uri:


sontek:
Rather than passing all environment variables to the configuration it might be easier for you to just pass the env vars to the command line when launching the application:

```
$ export HTTPPORT=8800
$ bin/pserve development.ini httpport=$HTTPPORT
```

and then just use the standard `%(httpport)s` syntax

msabramo:
@sontek: Yeah, I also thought of using the command-line vars because they're already there. I tried that approach actually before I started to hack on Pyramid. It didn't work though because the code that configures logging is currently unaware of those variables (those vars are currently only passed to `paste.deploy.{loadapp,loadserver}`).

The PR could be changed to do that of course.

One problem with that is it makes launching the app more complex, as the person or system launching the app now has to know what variables are required and it could potentially be a long list.

mmerickel:
Passing the entire environ to the INI file without any sort of namespacing is not going to be a acceptable. Prefixing things with `env:<key>` might be more interesting but the lack of support for fallbacks would still be a far cry from ideal, but maybe an improvement. I'm -0 on such a feature at this time.

mmerickel:
I think this is not going to be a workable approach in light of the other changes we're talking about making to the config system in #2419.

PR #2399, opened 2016-03-07
Pass vars to logging.config.fileConfig

msabramo:
This allows one to set up a logging configuration that is parameterized based on variables specified on the command-line.

e.g.: the application .ini file could have:

```ini
[logger_root]
level = %(LOGGING_LOGGER_ROOT_LEVEL)s
handlers = console

[handler_console]
class = StreamHandler
args = (sys.stderr,)
level = %(LOGGING_HANDLER_CONSOLE_LEVEL)s
formatter = generic
```

This app could be launched with:

```
pserve development.ini LOGGING_LOGGER_ROOT_LEVEL=DEBUG LOGGING_HANDLER_CONSOLE_LEVEL=DEBUG
```

Cc: @mmerickel, @sontek, @sudarkoff, @sseg, @aconrad

### DIFF
--- a/docs/api/paster.rst
+++ b/docs/api/paster.rst
@@ -11,4 +11,4 @@
 
     .. autofunction:: get_appsettings(config_uri, name=None, options=None)
 
-    .. autofunction:: setup_logging(config_uri)
+    .. autofunction:: setup_logging(config_uri, global_conf=None)
--- a/pyramid/paster.py
+++ b/pyramid/paster.py
@@ -52,7 +52,8 @@ def get_appsettings(config_uri, name=None, options=None, appconfig=appconfig):
         relative_to=here_dir,
         global_conf=options)
 
-def setup_logging(config_uri, fileConfig=fileConfig,
+def setup_logging(config_uri, global_conf=None,
+                  fileConfig=fileConfig,
                   configparser=configparser):
     """
     Set up logging via the logging module's fileConfig function with the
@@ -61,16 +62,22 @@ def setup_logging(config_uri, fileConfig=fileConfig,
 
     ConfigParser defaults are specified for the special ``__file__``
     and ``here`` variables, similar to PasteDeploy config loading.
+    Extra defaults can optionally be specified as a dict in ``global_conf``.
     """
     path, _ = _getpathsec(config_uri, None)
     parser = configparser.ConfigParser()
     parser.read([path])
     if parser.has_section('loggers'):
         config_file = os.path.abspath(path)
-        return fileConfig(
-            config_file,
-            dict(__file__=config_file, here=os.path.dirname(config_file))
-            )
+        if global_conf:
+            # Copy to avoid side effects
+            global_conf = dict(global_conf)
+        else:
+            global_conf = {}
+        global_conf.update(
+            __file__=config_file,
+            here=os.path.dirname(config_file))
+        return fileConfig(config_file, global_conf)
 
 def _getpathsec(config_uri, name):
     if '#' in config_uri:
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -27,6 +27,12 @@ unreleased
   where you may need more control over the request.
   See https://github.com/Pylons/pyramid/pull/2393
 
+- Allow using variable substitutions like ``%(LOGGING_LOGGER_ROOT_LEVEL)s``
+  for logging sections of the .ini file and populate these variables from
+  the ``pserve`` command line -- e.g.:
+  ``pserve development.ini LOGGING_LOGGER_ROOT_LEVEL=DEBUG``
+  See https://github.com/Pylons/pyramid/pull/2399
+
 1.6 (2015-04-14)
 ================
 
--- a/pyramid/scripts/pserve.py
+++ b/pyramid/scripts/pserve.py
@@ -362,7 +362,7 @@ a real process manager for your processes like Systemd, Circus, or Supervisor.
             log_fn = None
         if log_fn:
             log_fn = os.path.join(base, log_fn)
-            setup_logging(log_fn)
+            setup_logging(log_fn, global_conf=vars)
 
         server = self.loadserver(server_spec, name=server_name,
                                  relative_to=base, global_conf=vars)
--- a/pyramid/tests/test_paster.py
+++ b/pyramid/tests/test_paster.py
@@ -105,17 +105,37 @@ class Test_get_appsettings(unittest.TestCase):
         self.assertEqual(result['foo'], 'baz')
 
 class Test_setup_logging(unittest.TestCase):
-    def _callFUT(self, config_file):
+    def _callFUT(self, config_file, global_conf=None):
         from pyramid.paster import setup_logging
         dummy_cp = DummyConfigParserModule
-        return setup_logging(config_file, self.fileConfig, dummy_cp)
+        return setup_logging(
+            config_uri=config_file,
+            global_conf=global_conf,
+            fileConfig=self.fileConfig,
+            configparser=dummy_cp,
+            )
 
-    def test_it(self):
+    def test_it_no_global_conf(self):
         config_file, dict = self._callFUT('/abc')
         # os.path.abspath is a sop to Windows
         self.assertEqual(config_file, os.path.abspath('/abc'))
         self.assertEqual(dict['__file__'], os.path.abspath('/abc'))
         self.assertEqual(dict['here'], os.path.abspath('/'))
+
+    def test_it_global_conf_empty(self):
+        config_file, dict = self._callFUT('/abc', global_conf={})
+        # os.path.abspath is a sop to Windows
+        self.assertEqual(config_file, os.path.abspath('/abc'))
+        self.assertEqual(dict['__file__'], os.path.abspath('/abc'))
+        self.assertEqual(dict['here'], os.path.abspath('/'))
+
+    def test_it_global_conf_not_empty(self):
+        config_file, dict = self._callFUT('/abc', global_conf={'key': 'val'})
+        # os.path.abspath is a sop to Windows
+        self.assertEqual(config_file, os.path.abspath('/abc'))
+        self.assertEqual(dict['__file__'], os.path.abspath('/abc'))
+        self.assertEqual(dict['here'], os.path.abspath('/'))
+        self.assertEqual(dict['key'], 'val')
 
     def fileConfig(self, config_file, dict):
         return config_file, dict


msabramo:
Though maybe if we combine the code for passing vars derived from the command line options with code for populating those vars with values from environment variables, then we get something pretty nice.

msabramo:
Added:

```
[logging_global_conf 71bbd6a] pserve: get_options from env vars
 1 file changed, 4 insertions(+), 1 deletion(-)
```

With that change, one can use environment variables in both the logging sections and in the non-logging sections.

For example:

```ini
# development.ini

app:main]
use = egg:fakeproject

pyramid.reload_templates = %(PYRAMID_RELOAD_TEMPLATES)s

...

[logger_root]
level = %(LOGGING_LOGGER_ROOT_LEVEL)s
handlers = console
```

This app can be launched with:

```
$ PYRAMID_RELOAD_TEMPLATES=false LOGGING_LOGGER_ROOT_LEVEL=DEBUG LOGGING_HANDLER_CONSOLE_LEVEL=DEBUG pserve development.ini
pyramid.reload_templates = 'false'
Starting server in PID 6457.
serving on http://127.0.0.1:6543
```

msabramo:
I changed the implementation a bit so that it has less impact on Pyramid's tests and the loading of the environment variables is something you have to opt into by setting a special environment variable, `PSERVE_LOAD_ENV_VARS`.

With this configuration file:

```
# app.ini
[DEFAULT]
PYRAMID_RELOAD_TEMPLATES = false
LOGGING_HANDLER_CONSOLE_LEVEL = NOTSET
LOGGING_LOGGER_ROOT_LEVEL = WARN
FAKEPROJECT_FOO = bar

[app:main]
use = egg:fakeproject
fakeproject.foo = %(FAKEPROJECT_FOO)s
```

Without setting the environment variable, one gets the default value for `FAKEPROJECT_FOO` in the `DEFAULT` section: `bar`

```
$ pserve app.ini
settings.get('fakeproject.foo') = 'bar'
config.registry.settings.get('fakeproject.foo') = 'bar'
Starting server in PID 41111.
serving on http://127.0.0.1:6543
```

By setting 2 environment variables, we can override the default value of `FAKEPROJECT_FOO` from `bar` to `dog`:

```
$ PSERVE_LOAD_ENV_VARS=true FAKEPROJECT_FOO=dog pserve app.ini
settings.get('fakeproject.foo') = 'dog'
config.registry.settings.get('fakeproject.foo') = 'dog'
Starting server in PID 41658.
serving on http://127.0.0.1:6543
```

bertjwregeer:
@inklesspen was working on something similar with montague, have you looked at that?

msabramo:
Note that one of the advantages of using environment variables is that there are various other programs that help you to manage environment variables.

No environment variables set:

```
$ pserve app.ini
settings.get('fakeproject.foo') = 'bar'
config.registry.settings.get('fakeproject.foo') = 'bar'
Starting server in PID 43123.
serving on http://127.0.0.1:6543
```

But then using [direnv](http://direnv.net/):

```
$ cat > .envrc
export FAKEPROJECT_FOO=cool
export PSERVE_LOAD_ENV_VARS=true
direnv: error .envrc is blocked. Run `direnv allow` to approve its content.

$ direnv allow
direnv: loading .envrc
direnv: export +FAKEPROJECT_FOO +PSERVE_LOAD_ENV_VARS

$ pserve app.ini
settings.get('fakeproject.foo') = 'cool'
config.registry.settings.get('fakeproject.foo') = 'cool'
Starting server in PID 43481.
serving on http://127.0.0.1:6543
```

msabramo:
And one can set environment variables when launching Docker containers.

So that's cool because you could package a Pyramid app as a Docker image, but then deploy that image to lots of different environments by passing in different environment variables to configure the app for various environments.

Build once, run everywhere.

msabramo:
@bertjwregeer: Yes, I looked at [montague](https://github.com/inklesspen/montague) but it was a long time ago. I haven't looked at it recently. I remember not being clear on how it integrated with Pyramid.

I probably should look again. Is there a possibility that Pyramid might switch from PasteDeploy to montague?

inklesspen:
Montague is in a usable state for people who are fine with getting their hands dirty, but there's kind of a catch-22. I want to have real feedback from users that what I'm building is useful before I do much more work on it, but a lot of users generally don't want to try using it until some of the rough edges are gone.

I think the next step is as follows: package up montague versions of pserve and pshell so they can be installed, then pick a config format (JSON? YAML?) and probably package up a scaffold fork that uses it.

Does this seem like the right approach to anyone else?

msabramo:
A big drawback of what I have here is that it only works for pserve and not for other things with PasteDeploy support like gunicorn.

```
$ PSERVE_LOAD_ENV_VARS=true FAKEPROJECT_FOO=dog pserve app.ini
settings.get('fakeproject.foo') = 'dog'
config.registry.settings.get('fakeproject.foo') = 'dog'
...

$ PSERVE_LOAD_ENV_VARS=true FAKEPROJECT_FOO=dog gunicorn --paste=app.ini
...
settings.get('fakeproject.foo') = 'bar'
config.registry.settings.get('fakeproject.foo') = 'bar'
...
```

That of course limits the usefulness of it.

msabramo:
OTOH, I guess you could have a `[server:gunicorn]` setting in your PasteDeploy ini file and then use `pserve` to launch `gunicorn`. Does gunicorn still support (and recommend) doing that? I know they deprecated some stuff in the past, though I think what I'm thinking of is they used to have a `gunicorn_paster` command (and analogous ones for Django and what not -- https://github.com/benoitc/gunicorn/issues/589). Maybe launching gunicorn from PasteDeploy is still a recommended thing? In that case, my argument about this not being useful for gunicorn may not be applicable.

inklesspen:
gunicorn actively recommends against using pastedeploy (or montague) to load the wsgi server. the `gunicorn_paster` command uses the ini file to load the wsgi app only, and launches gunicorn normally.

So I'm adding a third TODO on my above list: have `montague_gunicorn` which works like `gunicorn_paster`.

I'll get started on `montague_pyramid_scripts` tonight; hopefully someone will volunteer to test it out.

msabramo:
I feel like PasteDeploy does a good job of letting you compose a system out of various applications, filters, servers, etc. and that's probably mainly what it was invented for -- as a system for composing WSGI pieces into a whole.

It also allows you to pass configuration to your applications and that part works reasonably well for a lot of folks, but it doesn't nicely support things like Twelve Factor.

One solution is to make PasteDeploy (or a similar system like Montague) better support stuff like environment variables.

Another is to leave application configuration to the application. In that case, no work would be needed on Pyramid, though it might be nice to add some recipes to the documentation about this. http://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/configuration/index.html would probably be a good place for that.

inklesspen:
As a note, montague already has great support for environment variables; you simply turn your config file into a mako template, and montague will render the environment variables into it, and then load the resulting config. (Docs are lacking, though.)

mmerickel:
> As a note, montague already has great support for environment variables; you simply turn your config file into a mako template, and montague will render the environment variables into it, and then load the resulting config.

I have to bring mako into the mix to get env variables? Support maybe, but great support?

inklesspen:
@mmerickel I needed some templating solution, and i wanted one that was used widely enough to not have too many warts. it would be entirely possible to do the same thing with jinja or tempita or whatever has a decent chance of being adopted. I crave your suggestions.

msabramo:
I feel like we went on a little bit of a tangent with Montague here. I think Montague might be interesting as a replacement for PasteDeploy, but my focus here is simply to come up with some kind of simple pattern for using environment variables to configure Pyramid apps in a Twelve Factorish way. Ideally it would be easy to understand and wouldn't require app developers to change their apps much. 

@mmerickel: What do you think of this PR?

msabramo:
Any comments on this PR?

mmerickel:
Similar comments to #2398. The lack of namespacing is an issue when injecting the environ vars. They NEED to be explicitly used by the file.

I *thought* this PR would be simply about making the CLI options available to the logging setup, but apparently this is also trying to mess with env vars making two features into one and hard to judge.

msabramo:
@mmerickel You're right -- I tacked env vars stuff on to the original PR of just making the CLI options available to the logging setup, which confused things.

I just removed those commits, so now this PR is just about making the CLI options available to the logging setup.

msabramo:
Oh I just spotted a problem. I'm not adding `__file__` and `here` to `global_conf` when vars are provided. I'll add that.

msabramo:
Needs tests too, if folks like the idea.

msabramo:
Fixed the `__file__` and `here` problem with 21eefe5. Kept as a separate commit so reviewers can see history. I can squash later if we decide to merge this.

msabramo:
Added tests in 32c0c3d.

Let me know if there are further comments on this...

pyramid/paster.py
```diff
@@ -53,7 +53,7 @@ def get_appsettings(config_uri, name=None, options=None, appconfig=appconfig):
         global_conf=options)
 
 def setup_logging(config_uri, fileConfig=fileConfig,
-                  configparser=configparser):
+                  configparser=configparser, global_conf=None):
```
mmerickel: If you look at `docs/api/paster.rst` you will see that `fileConfig` and `configparser` are private apis used for testing. This option should be documented and go ahead of the private options.

msabramo: OK. Push some new commits to handle these issues:

- 	Change arg order to setup_logging …			6e52898
-	docs/api/paster.rst: Add global_conf arg to setup_logging			e9ac5ad
-	Document new global_conf arg to setup_logging			5bc0669

docs/api/paster.rst
```diff
@@ -11,4 +11,4 @@
 
     .. autofunction:: get_appsettings(config_uri, name=None, options=None)
 
-    .. autofunction:: setup_logging(config_uri)
+    .. autofunction:: setup_logging(config_uri, global_conf)
```
mmerickel: I think this needs to be `global_conf=None`.

mmerickel:
This PR seems fine to me to improve the `setup_logging` support. +1

docs/api/paster.rst
msabramo: Oops, thanks! Fixed in 78904fc.

mmerickel:
If you want to touch things up or rebase or anything prior to merge go for it. I'll merge later today.

msabramo:
Rebased and squashed down to one commit. 8f26879b221cb9914b57a03cdfcac71e3e411365

msabramo:
@mmerickel Did you want me to add an entry to `CHANGES.txt`?

msabramo:
Went ahead and added a `CHANGES.txt` entry.

mmerickel:
Thanks @msabramo!

msabramo:
So there is one bummer to this because of the stdlib configparser module. I don't see any easy way to solve it.

It would be nice to specify default values in a `[DEFAULT]` section in the `.ini` file and then have the values specified on the command-line override these.

Unfortunately, it doesn't work that way. Because of the way configparser works, the values in the `[DEFAULT]` section always take precedence.

Don't see any way to fix this without reimplementing stuff from the stdlib.

PR #2400, opened 2016-03-08
Extra colon in docs breaks code block highlighting.

ianjosephwilson:
2cc0710ecf9b47e5a6f41caf6b1c56b29bab2db0

### DIFF
--- a/pyramid/static.py
+++ b/pyramid/static.py
@@ -229,7 +229,7 @@ class ManifestCacheBuster(object):
 
     By default, it is a JSON-serialized dictionary where the keys are the
     source asset paths used in calls to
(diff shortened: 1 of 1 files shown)


mmerickel:
LGTM, ping @stevepiercy

stevepiercy:
In addition to removing the extra colon, the lexer is incorrect. It should be [pycon](http://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer) for Python console.

I used a grep search `\:\:\s+\.\.` to find any other potential candidates for correction, and found no others.

mmerickel:
nice I didn't know about that lexer option

stevepiercy:
Actually only needed backporting to 1.6, as this code was added in 1.6.

Issue #1264, opened 2014-03-07
accept handling during view lookup is unpredictable


mmerickel:
As per the examples in #1259, it should be clear that the accept handling in Pyramid is somewhat unpredictable. It is a smell that the view invoked is somewhat dependent on the ordering of a `set`. When defining some views with accept predicates, a user should be able to have a solid understanding of when each view will be invoked, and possibly be able to influence what happens what a client simply asks for `*/*` or does not include an `Accept` header.

mmerickel:
I think that the `accept=` predicate should be updated to support a quality measurement. For example `accept='application/json;q=1.0'` vs another view with `accept='application/json;q=0.5'`. We could leave the default quality at 1.0. It'd also be useful if ambiguous views could be detected and emit a warning (possibly a `ConfigurationError` in pyramid 2.0).

bertjwregeer:
Those two that you just mentioned should conflict. They are both of type 'application/json' and the second would NEVER match.

`accept='application/json;q=1.0'` vs `accept='text/html;q=0.5'` should mean that `application/json` gets picked.

mmerickel:
Yep, bad example on my part.

mmerickel:
The server quality is already supported by webob and so we could start supporting it with only minimal changes to pyramid's accept handling.

https://github.com/Pylons/webob/blob/31620bb8b7749c92693e4d6d7233d9547b942a57/webob/acceptparse.py#L126

I'm interested in arguments for or against this.

mmerickel:
The first argument against I can think of is that the interaction between server quality and client quality is a little hard to understand because they are multiplied together to come up with the final quality / sorted order of views (it's not defined at config time but rather per request). Regardless of that, I think it's not a bad solution.

bertjwregeer:
In most cases the server quality will be used mainly to disambiguate which one is preferred ("default") server side.

mmerickel:
> In most cases the server quality will be used mainly to disambiguate which one is preferred ("default") server side.

Are you saying that the quality option is overkill then? You'd rather just have a server-side weighting that is not combined with the client-side weighting?

bertjwregeer:
I would like the option of having it take into account client side weighting (so that for example if my client sends `application/json` first, and `text/html` with q=0.5 then `application/json` could win if I have set up a view with `text/html;q=1.0` and another with `application/json;q=0.9`. But I think that would be more difficult to document if we had to fully explain how accept matching works in the first place.

[...]

PR #2409, opened 2016-03-14
Do not use a set for accept dedup in MultiView.add

romuald:
Starting python3.3, string hash are randomised by default. Meaning the self.accept list order was random.

In case of a MultiView accepting multiple types, the first matching type of that list is used.

The result was that the method actually used was random based on the initial seed.

Example bellow. When using GET without an accept header (or */*), the method called was randomly get_json or get_text

```python
@view_defaults(route_name='example')
class ExampleView(object):
    def __init__(self, request):
        self.request = request

    @view_config(renderer='json', accept='application/json')
    def get_json(self):
        print('GET JSON')
        return 'json'

    @view_config(renderer='string', accept='text/plain')
    def get_text(self):
        print('GET TEXT')
        return 'text'
```

### DIFF
--- a/pyramid/config/views.py
+++ b/pyramid/config/views.py
@@ -568,9 +568,8 @@ class MultiView(object):
             else:
                 subset.append((order, view, phash))
                 subset.sort(key=operator.itemgetter(0))
-            accepts = set(self.accepts)
-            accepts.add(accept)
-            self.accepts = list(accepts) # dedupe
+            if accept not in self.accepts:
+                self.accepts.append(accept)
 
     def get_views(self, request):
         if self.accepts and hasattr(request, 'accept'):


romuald:
I have no idea how to actually test this. Since the hash see is random and initialized during interpreter start, the test would be randomly failing or succedding randomly

Themanwithoutaplan:
You should be able to write some unit tests for the relevant branch of this method. And you should do this before the code change. `set()` can only have unique items so the code you have proposed only affects the order of items.

pyramid/config/views.py
```diff
@@ -568,9 +568,8 @@ def add(self, view, order, accept=None, phash=None):
             else:
                 subset.append((order, view, phash))
                 subset.sort(key=operator.itemgetter(0))
-            accepts = set(self.accepts)
-            accepts.add(accept)
-            self.accepts = list(accepts) # dedupe
```
Themanwithoutaplan: accepts = sorted(list(set(accepts))

romuald: Not sure I agree with the sort logic: in that case the "default" accept will be the lower one (application/foo vs text/bar)

The list/append method have the advantage of defaulting to the first declared view accept

Themanwithoutaplan: Then a relevant sort function can be passed in. The use of set() already indicates that order does not matter but you have an edge case where two different views can match the same request in some circumstances (missing accepts-header).

bertjwregeer:
This a known issue.

Right now with `.scan()` with `accept=` across different files there is no sane ordering either, so even if we were to change this to not use `set`, there would still not be a good way to know which one takes priority over the other.

What if for example I wanted `text/html` over `application/json` if the user agent sends `*/*`, how can I guarantee that? Right now there is no way to do so.

Also of note, with a sorted list as specified in a comment above, we end up with `['application/json', 'text/html']`.

I am a :-1: on merging this.

pyramid/config/views.py
bertjwregeer: There is no concept of "default" accept, and if you use `.scan()` across multiple files for example, it becomes even murkier because it highly depends on which one gets scanned first as to which one becomes the "default". This is a known issue.

bertjwregeer:
See this thread for more information: https://github.com/Pylons/pyramid/issues/1264

mmerickel:
This PR is not a good solution to the problem. The accept data is intentionally unordered and is relying on webob to select the best match. @bertjwregeer linked to the correct issue where this should be discussed. Currently it's simply not possible to put enough information into the `accept=` predicate to unambiguously sort the views.

PR #2411, opened 2016-03-14
polish Invoking an Exception View docs

stevepiercy:
- add index entry
- minor grammar, syntax


mmerickel:
LGTM thank you @stevepiercy

Issue #2367, opened 2016-02-23
replace check_csrf with require_csrf


mmerickel:
The `check_csrf` view predicate as it currently stands is not considered useful because it behaves as a predicate. Thus if the CSRF check fails the normal scenario is that a `PredicateMismatch` is raised and results in a 404.

In 1.7 with the introduction of view derivations in #2021 we can add a new `require_csrf` derivation that properly raises `BadCSRFToken` which can be caught and handled in an exception view. The `check_csrf` can be deprecated at this point and potentially removed in the future.

bertjwregeer:
:+1:

mmerickel:
Should `require_csrf=` with `True` / `False` be the only option? The other issue I can think of is wanting to use the deriver on a view that accepts both get and post, at which point we may want to only require csrf on non-get/head requests (post, put, patch, delete, etc). It seems like it might be a little tough to generalize but I think what I said would work. Maybe this could be overridden by `require_csrf='always'` instead of `True` or we could just define a few strings instead of true/false. Anyone have some thoughts on this?

sontek:
Could we allow `require_csrf` to be a callable that inspects that request and returns True/False?  Might be nice to be able to dynamically enable/disable CSRF checks

mmerickel:
well... here's what I didn't mention yet that will influence my answer here: Because of issues like this, the view deriver will probably not be added to your app by default but rather just another thing people can use if they want, similar to the default session factories etc. For example:

```python
config.add_view_deriver(`require_csrf`, `pyramid.session.require_csrf_view`)
```

If you want something that can do totally custom logic at the point of view declaration then you probably just want `decorator=require_csrf`. Passing a callable is better served by the existing `decorator` deriver. This PR is about a more declarative approach where you define the policy globally and then you can pass options to it per-view. That's the goal, I'm not saying we'll get there but I don't want to add an API with knobs on it if I don't have to. The idea is that you can already just swap out the `require_csrf` view deriver we're talking about here for a different one that may do a better job with your callable.

bertjwregeer:
I think that the CSRF view derivation that we ship with core should come with seat-belts enabled. If you load it, CSRF checking is mandatory for all `POST`.

`PUT`/`DELETE` are not required because they can't be used cross site without OPTIONS. Anyone that needs more flexibility should adapt the code from core and modify it as necessary and load it as a view derivation.

I am a -1 on knobs.

mmerickel:
[...]

PR #2413, opened 2016-03-22
require_csrf to replace check_csrf

mmerickel:
Fixes https://github.com/Pylons/pyramid/issues/2367

- [x] Merge https://github.com/Pylons/pyramid/pull/2021 and rebase this PR on that.
- [x] Deprecate `check_csrf`.
- [x] Fix any other docs and docstrings referencing `check_csrf` to point to `require_csrf` instead.
    - [x] Exclude `docs/whatsnew-*`
    - [x] `docs/api/session.rst`
    - [x] `docs/narr/hooks.rst`
    - [x] `docs/narr/sessions.rst`
    - [x] `docs/narr/viewconfig.rst`
    - [x] `pyramid/session.py`
    - [x] `pyramid/view.py`
    - [x] `pyramid/config/predicates.py`
    - [x] `pyramid/config/views.py`
    - [x] `pyramid/tests/tests_session.py`
    - [x] `pyramid/tests/test_predicates.py`
- [ ] Update the request processing diagram to show the csrf checks just below authorization in the view pipeline. See #2473.
- [x] ~~Update scaffolds to set `pyramid.require_default_csrf = yes` in `development.ini` and `production.ini`~~. See branch https://github.com/Pylons/pyramid/tree/docs/easy-install-to-pip.2104.


docs/narr/hooks.rst
```diff
@@ -1660,18 +1665,19 @@ token unless ``disable_csrf=True`` is passed to the view:
    from pyramid.response import Response
    from pyramid.session import check_csrf_token
 
-   def require_csrf_view(view, info):
+   def csrf_view(view, info):
```
bertjwregeer: This should probably match the actual `csrf_view` we ship more closely (related to the `val` stuff), and add just the check for `request.method == POST`.

mmerickel:
I've rebased this PR on master.

Two remaining questions on the impl:
1. Should I place the `csrf_view` above or below the `secured_view`? Right now I have it above but I'm not so sure anymore.
2. Should I take this approach of leaving it entirely opt-in on a per-view basis? My alternative right now is to add a setting such as `pyramid.default_require_csrf = yes`. If someone sets it then they will get the example code I put in the docs where it's on by default and only on POST. We could then add this setting to the scaffolds so that it's on by default for new projects.

bertjwregeer:
1. I would argue that it should be below secured_view. If you don't have permission to the view in the first place, raising an not authorized makes more sense than raising BadCSRFToken. They couldn't submit to that view in the first place anyway.
2. I like the opt-in with the setting to enable it by default, so long as on a particular view a user can turn it off. Then again, I am biased because it means I won't need to write my own that does it ;-)

mmerickel:
I will reverse the order. I'm working on implementing this feature now... it touches several more systems than the simple implementation but I think it'll be worth it. I'm adding `pyramid.require_default_csrf` which will turn on CSRF checking for an app's POST requests. It can be overridden by a local call to `add_view(require_csrf=...)`.

The only quirk I've added is that if you set `require_csrf=True` and `pyramid.require_default_csrf=foo` then the token used will be `foo` instead of `csrf_token`. This means if the view didn't specify a token then it'll fallback to whatever was in `require_default_csrf`. If that value is truthy then it'll fallback to `csrf_token`.

bertjwregeer:
That sounds good to me :-).

mmerickel:
Is there any interest in discussing how to enable/disable this feature based on whether authorization is done via headers or cookies? As I understand it, CSRF is only required when auth is via cookies that a browser will send automatically. I'm not too worried about this but I wanted to just hypothesize how an app could turn this on or off on a per-request basis. Like maybe someone could override this thing with one that supports a per-request setting to turn it off that they could set in their auth policy or something.

mmerickel:
Okay, I've done that work. I'd appreciate some review of the implementation.

docs/narr/sessions.rst
```diff
@@ -389,12 +404,51 @@ header named ``X-CSRF-Token``.
...
+query string and a submitted form body.
+
+It is always possible to pass the token in the ``X-CSRF-Token`` header as well.
+There is currently no way to define an alternate name for this header without
+performing CSRF checking manually.
+
+If CSRF checks fail then a :class:`pyramid.exceptions.BadCSRFToken` exception
+will be raised. This exception may be caught and handled by an
+:term:`exception view` but, by default, will result in a ``400 Bad Request``
+resposne being sent to the client.
```
bertjwregeer: typo on response

bertjwregeer:
Besides the typo, this looks great :-)

bertjwregeer:
:+1: this looks fantastic! Thanks @mmerickel!

mmerickel:
Okay so the only point remaining is to fix the scaffolds which I'm afraid to do while @stevepiercy is messing with the easy_install vs pip stuff. I need some guidance on when that might be done so we can add it. We just need to slip in a little `pyramid.require_default_csrf = yes` into the ini files. No other changes or docs required.

mmerickel:
> We just need to slip in a little pyramid.require_default_csrf = yes into the ini files. No other changes or docs required.

Well that's not quite right. If we do this we also need to setup the scaffolds to configure a session factory. I'm fine with that but it's a new thing. The session will require adding another secret to the settings at least, as well as a bit of code. I think that's fine for the alchemy and zodb scaffolds... I'm a little more wary of adding it to the starter scaffold which may not want a session.

bertjwregeer:
> The session will require adding another secret to the settings at least, as well as a bit of code.

Do we currently have any "secrets" in the settings? I am now worried that if we add this secret, and keep it the default, that people will end up running their applications using the config file and be vulnerable to remote code execution...

stevepiercy:
@mmerickel I'm down to the last file to update, so today.

I've moved `HACKING.txt` and Pyramid's `setup.py` into a separate issue #2474, too, since that is outside of docs scope and deserves special attention from developers working on the core.

mmerickel:
> Do we currently have any "secrets" in the settings?

Yes, the alchemy tutorial added the auth policy secret to the settings as the tutorial was previously using a hard-coded string which was even worse.

bertjwregeer:
Hmmm, okay, guess adding one more secret isn't that big of a deal then.

mmerickel:
The bigger issue is like I said it requires adding a session to the app which means invalidating it on boundaries as well as possibly explaining such things. On the bonus side it means we could possibly add flash messages to the tutorial in the future to make it nicer. I had previously envisioned adding this stuff in a new chapter but the invalidation might just need to go earlier.

The alternative is to leave this stuff out of the scaffolds for this release and stew on whether to update the scaffolds / tutorials.

bertjwregeer:
I hadn't thought of the fact that we'd have to add session to the basic scaffolds, not sure it is worth adding that by default especially since a lot of people prefer to use `pyramid_beaker` for instance to store data server side versus using cookie sessions.

Could we possibly leave it in the .ini commented out with a comment that you'll need to enable a session before enabling the feature?

mmerickel:
I think we should punt that change for a later version and just ship this feature as something people can turn on. We should consider (I already wanted to do this but it was getting out of scope) adding a session chapter to the tutorial.

bertjwregeer:
Could you merge master into this please? Then I'll click the merge button.

PR #2421, opened 2016-03-31
Docs/easy install to pip.2104

stevepiercy:
Another pair of eyes would be much appreciated, especially on the installation of add-ons, like pyramid_chameleon.


stevepiercy:
Grrr... GitHub UI is tricksy. Closing in favor of #2422.

PR #2433, opened 2016-04-06
setup_logging: global_conf => extra_global_conf

msabramo:
In `setup_logging` function, rename the argument `global_conf` to `extra_global_conf`.

This lets us have a somewhat simpler and nicer implementation than what I had in https://github.com/Pylons/pyramid/pull/2399.

Cc: @mmerickel

### DIFF
--- a/pyramid/paster.py
+++ b/pyramid/paster.py
@@ -69,15 +69,12 @@ def setup_logging(config_uri, global_conf=None,
     parser.read([path])
     if parser.has_section('loggers'):
         config_file = os.path.abspath(path)
-        if global_conf:
-            # Copy to avoid side effects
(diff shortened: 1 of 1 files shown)


mmerickel:
Nah, this is inconsistent with the `load_app`, `load_server` and `load_context` apis which all accept `global_conf`. If you want to cleanup the impl that's fine, but the public api will remain consistent as it was before this PR.

msabramo:
Ah right, good idea to be consistent with those others.

msabramo:
OK, changed so the argument name doesn't change; just the implementation. New impl gets rid of an `else` and copying dict to avoid side effects.

PR #2435, opened 2016-04-07
separate viewderiver module

mmerickel:
This is some work to move the view derivers into `pyramid.viewderivers`. This moves the implementations out into the `pyramid.viewderivers` package as they are not technically configuration. They are implementations used by the configurator as defaults. Similarly we should in a separate PR move the predicates out of the `pyramid.config` package as well.

This PR allows overriding every builtin deriver. Previously everything could be overridden except `mapped_view`. While I'm not saying it's a good idea, at least it is treated the same now as all the other derivers.

Also, I've introduced `pyramid.viewderivers.INGRESS` and `pyramid.viewderivers.MAPPED_VIEW`. I do not like the name `MAPPED_VIEW` but anything else doesn't seem to fit very well. I considered `VIEW_MAPPER` and `VIEW` (analogous to `pyramid.tweens.MAIN`).

Finally, it is now required to explicitly define `over` if you define `under` and vice versa. The only time it's not required is if you set both to `None` at which point it'll be over `rendered_view` and under `decorated_view`. This is all a little crazy but I'm struggling to come up with a better idea.

Oh yeah, and I also merged `authdebug_view` with `secured_view` because it was just annoying that they were separate imo.

I've left this as a separate PR for now to make it easier to review the changes against the currently working view deriver in #2021.


pyramid/config/views.py
```diff
@@ -1028,17 +1031,15 @@ def _check_view_options(self, **kw):
...
 
     def _apply_view_derivers(self, info):
-        d = pyramid.config.derivations
-        # These derivations have fixed order
+        d = pyramid.viewderivers
+
+        # These derivers are not really derivers and so have fixed order
         outer_derivers = [('attr_wrapped_view', d.attr_wrapped_view),
                           ('predicated_view', d.predicated_view)]
-        inner_derivers = [('mapped_view', d.mapped_view)]
```
bertjwregeer: This is a change I personally struggled with when I was working on this branch. I added `inner_derivers` back at one point after discussing it with you @mmerickel, but ultimately it makes sense that even this can be overridden at will.

pyramid/config/views.py
```diff
@@ -1028,17 +1031,15 @@ def _check_view_options(self, **kw):
             raise ConfigurationError('Unknown view options: %s' % (kw,))
 
     def _apply_view_derivers(self, info):
-        d = pyramid.config.derivations
-        # These derivations have fixed order
+        d = pyramid.viewderivers
+
+        # These derivers are not really derivers and so have fixed order
         outer_derivers = [('attr_wrapped_view', d.attr_wrapped_view),
```
bertjwregeer: It still feels weird to me to have something that is specific to the view machinery and HAS to exist, is applied only during the view derivation stage because it happens to need to wrap the view.

Not that I can think of a better place, but just wanted to air my feelings on it.

bertjwregeer:
:+1:

I'm happy with this :-)

mmerickel:
> Finally, it is now required to explicitly define over if you define under and vice versa. The only time it's not required is if you set both to None at which point it'll be over rendered_view and under decorated_view. This is all a little crazy but I'm struggling to come up with a better idea.

I'm considering changing this such that the default values are simply `over='rendered_view'` and `under='decorated_view'`. If you change one you are stuck with the other (including the cycle) until you specify both but at least it's a bit more straightforward to document I think.

bertjwregeer:
Would this remove the INGRESS/VIEW_MAPPER?

mmerickel:
They are still required to put something between rendered_view and mapped_view or between ingress and secured_view.

bertjwregeer:
Alright, I am a +1 on changing that to simplify docs.

mmerickel:
I think the only way to get rid of the ingress sentinel value is to use `None` for it. For example  `under=None` means you don't care what it's under so it's clamped to the start. The only way it breaks down is if you want to pass a list of constraints like `under=(INGRESS, 'foo')`. You could say `under=(None, 'foo')`....... I don't have a large opinion on that one.

I will get rid of the mapped_view though, because it's unnecessary versus just using the documented string that is now in the list of valid names to use since it isn't tacked on via the `inner_derivers` later.

bertjwregeer:
Cool. Looking forward to the final changes here to merge them in :-)

mmerickel:
Alright, I'm done with this. I defined a new sentinel `VIEW` instead of `MAPPED_VIEW` to make it possible to do `over=VIEW`. This is the sanest impl I've come up with so far so let's roll with it!

PR #2442, opened 2016-04-08
rough draft for wiki2, replace setup.py with pip. See #2104.

stevepiercy:


docs/tutorials/wiki2/installation.rst
```diff
@@ -242,79 +346,65 @@ On Windows
...
-   OK
+======================== test session starts ========================
+platform Python 3.5.1, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
+rootdir: /Users/stevepiercy/projects/pyramidtut/tutorial, inifile:
+plugins: cov-2.2.1
+collected 2 items
+
+tutorial/tests.py ..
+------------------ coverage: platform Python 3.5.1 ------------------
+Name                               Stmts   Miss  Cover
```
Themanwithoutaplan: You'll want to add the `--cov-report=term-missing` option to the call to get the same kind of report as previously.

docs/tutorials/wiki2/installation.rst
```diff
@@ -215,25 +214,130 @@ On Windows
...
+.. code-block:: bash
+
+   $ $VENV/bin/pip install -e ".[testing]"
+
+On Windows
+^^^^^^^^^^
+
+.. code-block:: ps1con
+
+   c:\pyramidtut\tutorial> %VENV%\Scripts\pip install -e ".[testing]"
```
Themanwithoutaplan: Because this is *so* much easier and less likely to have typos than `pip install -r testing_requirements.txt`! NB this will attempt to install the package again and tell the use that it's already up to date.

docs/tutorials/wiki2/installation.rst
stevepiercy: Done in https://github.com/Pylons/pyramid/pull/2442/commits/aa51dcc629028fda64ee83340d9b3981be7fbb29

docs/tutorials/wiki2/installation.rst
stevepiercy: I think this pattern got lost in the comment on [hate against requirements.txt](https://github.com/Pylons/pyramid/issues/2104#issuecomment-200040138). Perhaps the discussion about the advantages and disadvantages of `testing_requirements.txt` would be better on issue #2104 where there are more participants?

bertjwregeer: @Themanwithoutaplan I am a -1 on splitting out into a "testing_requirements.txt" and a "requirements.txt" and having setup.py. Currently there is a way for us to easily add extra dependencies, it's in the same file as all other dependencies, and makes it simple to understand.

mmerickel: Yes, all dependencies of the project belong in the `setup.py` file. We will not be requiring anyone to use tox as a test runner. tox shines in libraries where a matrix is required but for apps it is much less widely used.

docs/tutorials/wiki2/installation.rst
```diff
@@ -215,25 +214,130 @@ On Windows
...
+         url='',
+         keywords='web wsgi bfg pylons pyramid',
+         packages=find_packages(),
+         include_package_data=True,
+         zip_safe=False,
+         test_suite='tutorial',
+         extras_require={
+             'testing': testing_extras,
+         },
+         tests_require=tests_require,
```
mmerickel: if we are moving the test runner out of setup.py (we are no longer advocating `python setup.py test`) then we should stop using the `tests_require` and `test_suite` options in the `setup.py`. It should be simplified to simply the extras dependency. This means also merging `tests_require` and `testing_extras` into a single list up above.

stevepiercy: @mmerickel done in https://github.com/Pylons/pyramid/pull/2442/commits/c239c4a68b41dab9d3f4477ea7de6431ec7548dd Please review.

mmerickel: it looks like test_suite is still in there

docs/tutorials/wiki2/installation.rst
```diff
@@ -215,25 +214,126 @@ On Windows
...
+           "Topic :: Internet :: WWW/HTTP :: WSGI :: Application",
+           ],
+         author='',
+         author_email='',
+         url='',
+         keywords='web wsgi bfg pylons pyramid',
+         packages=find_packages(),
+         include_package_data=True,
+         zip_safe=False,
+         test_suite='tutorial',
```
bertjwregeer: This line is no longer necessary.

bertjwregeer:
:+1: Looks good.

mmerickel:
This looks like a good start to me. We will need to update the scaffolds using this approach as well as the other wiki tutorial. Changing the scaffold is a big deal because it affects lots of line numbers in the tutorial itself as well as the code for each chapter of the tutorial. So don't do that until the scaffold is finalized!

stevepiercy:
I'm merging this first big step now. I'll work on the remaining steps in the wiki2 tutorial in the next few days. Thanks for the help on this PR everyone!

PR #2467, opened 2016-04-11
Remove: py26

bertjwregeer:
In accordance with this: https://github.com/Pylons/pyramid/issues/2368

Here's a PR that removes Python 2.6 support. If we want to drop Python 3.3 support I can modify this PR as necessary.

### DIFF
--- a/docs/narr/install.rst
+++ b/docs/narr/install.rst
@@ -21,9 +21,9 @@ the following sections.
 
 .. sidebar:: Python Versions
 
-    As of this writing, :app:`Pyramid` has been tested under Python 2.6, Python
-    2.7, Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3. :app:`Pyramid`
-    does not run under any version of Python before 2.6.
+    As of this writing, :app:`Pyramid` has been tested under Python 2.7,
+    Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3. :app:`Pyramid` does
+    not run under any version of Python before 2.7.
 
 :app:`Pyramid` is known to run on all popular UNIX-like systems such as Linux,
 Mac OS X, and FreeBSD, as well as on Windows platforms.  It is also known to
--- a/docs/narr/introduction.rst
+++ b/docs/narr/introduction.rst
@@ -859,14 +859,15 @@ Testing
 Every release of Pyramid has 100% statement coverage via unit and integration
 tests, as measured by the ``coverage`` tool available on PyPI.  It also has
 greater than 95% decision/condition coverage as measured by the
-``instrumental`` tool available on PyPI.  It is automatically tested by the
-Jenkins tool on Python 2.6, Python 2.7, Python 3.3, Python 3.4, Python 3.5,
-PyPy, and PyPy3 after each commit to its GitHub repository. Official Pyramid
-add-ons are held to a similar testing standard.  We still find bugs in Pyramid
-and its official add-ons, but we've noticed we find a lot more of them while
-working on other projects that don't have a good testing regime.
+``instrumental`` tool available on PyPI. It is automatically tested by Travis,
+and Jenkins on Python 2.7, Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3
+after each commit to its GitHub repository. Official Pyramid add-ons are held
+to a similar testing standard.  We still find bugs in Pyramid and its official
+add-ons, but we've noticed we find a lot more of them while working on other
+projects that don't have a good testing regime.
 
-Example: http://jenkins.pylonsproject.org/
+Travis: https://travis-ci.org/Pylons/pyramid
+Jenkins: http://jenkins.pylonsproject.org/job/pyramid/
 
 Support
 ~~~~~~~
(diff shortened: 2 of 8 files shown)


bertjwregeer:
Ping @stevepiercy @tseaver @mmerickel @dstufft

bertjwregeer:
(Did not run tests locally, so if travis breaks I will fix)

docs/narr/install.rst
```diff
@@ -9,14 +9,14 @@ Installing :app:`Pyramid`
...
-You will need `Python <http://python.org>`_ version 2.6 or better to run
+You will need `Python <http://python.org>`_ version 2.7 or better to run
 :app:`Pyramid`.
 
 .. sidebar:: Python Versions
 
-    As of this writing, :app:`Pyramid` has been tested under Python 2.6, Python
-    2.7, Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3. :app:`Pyramid`
-    does not run under any version of Python before 2.6.
+    As of this writing, :app:`Pyramid` has been tested under, Python 2.7,
```
stevepiercy: remove comma after "under,"

docs/narr/introduction.rst
```diff
@@ -859,14 +859,14 @@ Testing
...
-
-Example: http://jenkins.pylonsproject.org/
+``instrumental`` tool available on PyPI. It is automatically tested by the
+Travis on Python 2.7, Python 3.3, Python 3.4, Python 3.5, PyPy, and PyPy3
+after each commit to its GitHub repository. Official Pyramid add-ons are held
+to a similar testing standard.  We still find bugs in Pyramid and its official
+add-ons, but we've noticed we find a lot more of them while working on other
+projects that don't have a good testing regime.
+
+Example: https://travis-ci.org/Pylons/pyramid
```
stevepiercy: We test on both Jenkins and Travis. Why exclude or replace one? I admit, I never check Jenkins, unless someone tells me to do so.
* http://jenkins.pylonsproject.org/job/pyramid/
* https://travis-ci.org/Pylons/pyramid

stevepiercy:
@bertjwregeer I'd suggest waiting until the branch https://github.com/Pylons/pyramid/tree/docs/easy-install-to-pip.2104 is merged to master, as there are dozens of small changes I have already made on that branch that will conflict with this PR.

Other than the timing, I would :heart_eyes: to see 2.6 and 3.3 support go away.

docs/narr/introduction.rst
bertjwregeer: I honestly didn't realise that Jenkins was still online, and I only ever verify testing against Travis. I can revert that part of it if you'd like, or mention both.

stevepiercy: @tseaver pointed out a failure on Jenkins as a result of a PR I once made, which passed on Travis-CI. I honestly don't know how Jenkins is used. Perhaps notifications?

bertjwregeer:
Okay, fixed comma, and updated the testing introduction to include both Travis and Jenkins.

dstufft:
Looks good to me, :shipit:

docs/narr/introduction.rst
tseaver: http://jenkins.pylonsproject.org/ is still running tests for pyramid, waitress, etc.:  it sends e-mails on failure / restore-to-passing to the pyramid-checkins@lists.repoze.org mailing list.

bertjwregeer: @tseaver I have added the information about Jenkins back in. See latest changes.

setup.py
```diff
@@ -26,8 +26,8 @@
     if py_version < (3, 3) and not is_pypy: # PyPy3 masquerades as Python 3.2...
         raise RuntimeError('On Python 3, Pyramid requires Python 3.3 or better')
 else:
-    if py_version < (2, 6):
-        raise RuntimeError('On Python 2, Pyramid requires Python 2.6 or better')
+    if py_version < (2, 7):
+        raise RuntimeError('On Python 2, Pyramid requires Python 2.7 or better')
```
mmerickel: I'm a little concerned about adding 2.6 as a hard fail considering pyramid does still run on 2.6 just fine. I'd prefer to wait to add a hard fail until (if) we start using 2.7-only features like set/dict comprehensions, etc. Even if the test suite stops working on 2.6 I don't see a reason to hard fail until the core library uses 2.7-only features.

bertjwregeer: I'd be concerned about us introducing changes that only work on Py 2.7 and since we no longer test on Py 2.6 we don't catch it...

mmerickel: What does that have to do with a hard fail? I don't particularly like refusing to run our software on a system just because we don't support it.

bertjwregeer: Let's say we introduce set/dict comprehensions for example, and don't catch it because we no longer test on Py 2.6 but still allow people to install and attempt to run on Py 2.6, this would break things. I'd rather have a clear cut off.

If you'd prefer to have a soft cut-off, I will happily make the change, I am not dead set on it one way or the other, but we have the protection at the moment and it seemed like updating it would be the right thing to do.

bertjwregeer:
Closes #2368

Issue #2406, opened 2016-03-10
forward port the webob httpexception changes to support json mime types


mmerickel:
https://github.com/Pylons/webob/pull/230

bertjwregeer:
Taking this. Need to release WebOb 1.6.

bertjwregeer:
If only Pyramid had lifted the top-level exception wholesale ;-)

bertjwregeer:
So I just figured out that MIMEAccept in WebOb is busted as it is currently written.

Accept header:

`application/json, text/html, */*;q=0.8`

Server offers the following variants:

`text/plain, text/html, application/json`

with `text/plain` being the default if no offers match at all.

Current `MIMEAccept.best_match()` will return `text/html`, even though `application/json` is preferred by the client.

Fudging the numbers to `application/json;q=1.0, text/html;q=0.9, */*;q=0.8` which is exactly the same thing I expressed above, but implicitly rather than explicitly and `MIMEAccept.best_match()` now does the right thing.

bertjwregeer:
Blocked by: https://github.com/Pylons/webob/issues/239

PR #2489, opened 2016-04-13
Feature: JSON exceptions

bertjwregeer:
This is to bring Pyramid in-line with the changes that were made to WebOb's exceptions and to allow for generation of JSON exceptions.

https://github.com/Pylons/pyramid/issues/2406


bertjwregeer:
@mmerickel Take a look at this.

To understand why I was concerned WebOb was broken (and IMHO the ordering is wrong, but nobody but Apache implements it, and Apache implements it the same way as WebOb...) see this long issue: https://github.com/Pylons/webob/issues/239

Either way, this is ready to be merged as soon as tests come back satisfactory.

mmerickel:
This looks great. Could you please add a changelog that mentions any possible incompatibilities?

bertjwregeer:
Should be no incompatibilities. It now does the right thing when the client provides an `Acccept` of `application/json` and returns a JSON response.

I can add a CHANGELOG announcing the change.

bertjwregeer:
I did match the signature for `json_formatter` to be the same as the one used by WebOb. This should allow users to easily move those between projects.

Issue #2362, opened 2016-02-21
Pyramid 2.0 possible feature list


mmerickel:
I'm starting here a little 2.0 wishlist. I have no idea if an issue will be a good medium for the discussion but here goes. None of these are guaranteed to make the cut. There is no timeline for this yet. Personally I'm interested in MINOR incompatibilities that are large enough that we have refused to do until now (i.e. nothing that will require a fundamental redesign of a Pyramid app for a user).

**Please keep the discussion away from here and in separate issues. Let's focus on simply compiling a list of possibilities.**

**likely**
- Drop Python 2 support. #2903
- Switch the `ISession` contract from pickleable objects to something more secure and portable like json. #2709
- Remove `check_csrf` view predicate in favor of `require_csrf` (already deprecated).
- Remove `custom_predicates` from `config.add_view` and `view_config` (already deprecated).
- ~Remove `config.set_request_property` (already deprecated).~
- Remove `pcreate` and default scaffolds. #2384

**interesting, let's open an issue and talk about it**
- ~Separate CSRF tokens from the `ISession` into something else implementable without a session. #1573~
- Something to do with route prefixes so that it's possible to use `config.include` without the base url inside having a `/` suffix. #406, #1639 
- ~Drop PasteDeploy in favor of a better loading system. #2419~
- Fix predicate factories to accept a standardized `info` object instead of `config`. Currently the only purpose of this argument is for `config.maybe_dotted` but it's poor practice to pass the ephemeral `config` object around. #2535 
- Drop the `use_tweens` argument to `request.invoke_subrequest` in order to make the nested request pipeline more sane. #2606

**interesting but unlikely without a champion to start the discussion**
- Deprecate `remember`/`forget` APIs in favor of a pattern that fits better with non-cookie workflows.
- Something to do with identity and [auth-api post mortem](http://plope.com/pyramid_auth_design_api_postmortem)?
- Add a public API at the top-level pyramid namespace. For example `pyramid.view_config`, `pyramid.Configurator`, etc. This would greatly assist in maintainability of the public API for people using IDEs like pycharm where they may not be checking the docs to ensure an API is public. Probably use a lazy-instantiator like [apipkg](https://pypi.python.org/pypi/apipkg) for this. #2387 
- Re-work `view_config` to namespace arguments to be more clear about options versus predicates.
[...]

PR #2496, opened 2016-04-14
Switch to sha512 in AuthTktAuthenticationPolicy

bertjwregeer:
This matches the TODO located at: https://github.com/Pylons/pyramid/blob/master/TODO.txt#L123

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,22 @@
 unreleased
 ==========
 
+Backward Incompatibilities
+--------------------------
+
+- Following the Pyramid deprecation period (1.4 -> 1.6),
+  AuthTktAuthenticationPolicy's default hashing algorithm is changing from md5
+  to sha512. If you are using the authentication policy and need to continue
+  using md5, please explicitly set hashalg to 'md5'.
+
+  This change does mean that any existing auth tickets (and associated cookies)
+  will no longer be valid, and users will no longer be logged in, and have to
+  login to their accounts again. No other backwards incompatible changes have
+  been made.
+
+Features
+--------
+
 - Pyramid HTTPExceptions will now take into account the best match for the
   clients Accept header, and depending on what is requested will return
   text/html, application/json or text/plain. The default for */* is still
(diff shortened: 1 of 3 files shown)


mmerickel:
As I said in https://github.com/Pylons/pyramid/issues/2362 I don't want to do this until 2.0.

bertjwregeer:
Okay :-) Missed that.

tseaver:
@mmerickel Why is something we've warned about for three major releases a 2.0-only changes.

mmerickel:
@tseaver My concern is that the warnings don't actually show up for people. Python 2.7 changed the default behavior to not show DeprecationWarnings unless you explicitly set an environ variable or interpreter flag. It's ridiculous.

For this reason, and that the behavior of this change is something that will affect people who upgrade by logging out every user of their website if they didn't see the warnings... It feels like a bad idea.

We should do a better job with our deprecation warnings, we've never changed how it worked (the current system worked great on Python 2.6) but simply doesn't work well now.

tseaver:
@bertjwregeer pep8 failure:

> pyramid/authentication.py:8:1: F401 'warnings' imported but unused

bertjwregeer:
@tseaver Yeah, noticed that. But unless we agree to ship this soon, I was not planning on fixing this until 2.0.

mmerickel:
The deprecation is weird because it's changing the default. It's not removing a feature where user's code will error but rather just flipping it to something else. Super subtle.

tseaver:
IMHO, having user logins evaporate during a "major" software upgrade (and I would count Pyramid 1.6.x -> 1.7 as such) doesn't sound like a big enough issue to warrant deferring the change until 2.0.   We won't be "breaking" users software, in the sense that they will be forced to make code changes to get up and running:  users who care will be able to a) change the code ahead of time such that they pass an explicit `hashalg`, and thus won't see the cookies disappear;  b) schedule the upgrade for a time when they deem that having login cookies disappear is tolerable.

mmerickel:
Alright, I'm satisfied since we've at least discussed the issue. 2.0 was my conservative feeling as no one had expressed interest otherwise until now. We'll have to make the release notes very clear here.

bertjwregeer:
@mmerickel Updated CHANGES.txt with specific text. Also, re-organised it a little (forgot that I was on a feature branch).

mmerickel:
`hashalg` docstring is not updated on the policy

PR #2500, opened 2016-04-15
Increase the protection provided by the CSRF checks

dstufft:
Instead of only checking on the CSRF on `POST` requests, this will instead have the new, automatic CSRF checking occur on any HTTP method which is not defined as safe by RFC 2616.

### DIFF
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -106,13 +106,14 @@ def check_csrf_token(request,
                      header='X-CSRF-Token',
                      raises=True):
     """ Check the CSRF token in the request's session against the value in
-    ``request.params.get(token)`` or ``request.headers.get(header)``.
-    If a ``token`` keyword is not supplied to this function, the string
-    ``csrf_token`` will be used to look up the token in ``request.params``.
-    If a ``header`` keyword is not supplied to this function, the string
-    ``X-CSRF-Token`` will be used to look up the token in ``request.headers``.
+    ``request.POST.get(token)`` (if a POST request) or
+    ``request.headers.get(header)``. If a ``token`` keyword is not supplied to
+    this function, the string ``csrf_token`` will be used to look up the token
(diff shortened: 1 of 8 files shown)


pyramid/session.py
```diff
@@ -128,7 +129,24 @@ def check_csrf_token(request,
...
+    # any valid CSRF token, so that if we fail to locate a CSRF token then
+    # CSRF checks will fail.
+    supplied_token = ""
+
+    # If this is a POST request, then we'll check the POST body to see if it
+    # has a token. We explicitly use request.POST here because CSRF tokens
+    # should never appear in an URL as doing so is a security issue. We also
+    # explicitly check for request.POST here as we do not support sending form
+    # encoded data over anything but a request.POST.
+    if request.method == "POST":
```
mmerickel: `request.POST` can be used for any request that is uploaded using a form-encoding (multipart or urlencoded). A better check here would probably be `if token in request.POST` which will fail on any request where `request.POST` cannot be parsed.

bertjwregeer: It is perfectly acceptable for a PUT request to have a body that is form encoded, thus a PUT request could have the CSRF token in `request.POST`.

This would be a b/w incompatible change.

dstufft: I originally did this because I was told that WebOb will copy the body on ``PUT`` requests if we unconditionally accessed ``request.POST`` and that was one of the reasons that this couldn't be done. However looking at the code now in WebOb it appears it will *only* do that if the Content-Type matches, so I think that is probably reasonable then.

bertjwregeer: I had forgotten that was changed in WebOb, it used to do it unconditionally, but it was fixed because it would replace it with a FakeCGIBody, even on non form data.

bertjwregeer:
:+1:

PR #2501, opened 2016-04-16
In addition to CSRF token, verify the origin too

dstufft:
Add an additional layer of protection against CSRF by verifying the actual origin of the request in addition to the CSRF token. We only do this check on sites hosted behind HTTPS because only HTTPS sites have evidence to show that the Referrer header is not being spuriously removed by random middleware boxes.

Note, to prevent any sort of backwards incompatibilities and since the CSRF predicate has been deprecated, I've only added this to the new view derivation form of CSRF.

I still need to write documentation and tests (working on that now), just thought I'd throw this up here incase people were interested. This is same technique can be seen in [Django](https://github.com/django/django/blob/df8d8d4292684d6ffa7474f1e201aed486f02b53/django/middleware/csrf.py#L136-L189).

### DIFF
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -16,11 +16,19 @@ from pyramid.compat import (
     text_,
     bytes_,
     native_,
+    urlparse,
     )
 
-from pyramid.exceptions import BadCSRFToken
+from pyramid.exceptions import (
+    BadCSRFOrigin,
+    BadCSRFToken,
+)
 from pyramid.interfaces import ISession
-from pyramid.util import strings_differ
+from pyramid.settings import aslist
+from pyramid.util import (
+    is_same_domain,
+    strings_differ,
+)
 
 def manage_accessed(wrapped):
     """ Decorator which causes a cookie to be renewed when an accessor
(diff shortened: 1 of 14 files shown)


pyramid/session.py
```diff
@@ -101,6 +102,67 @@ def signed_deserialize(serialized, secret, hmac=hmac):
...
 
+
+def check_csrf_origin(request):
+    """
+    Check the Origin of the request to see if it is a cross site request or
+    not.
+
+    Note that this function will do nothing if request.scheme is not https.
+
+    ..versionadded:: 1.7
```
bertjwregeer: IIRC you need one more space after the initial two periods.

pyramid/session.py
```diff
@@ -16,11 +16,12 @@
     text_,
     bytes_,
     native_,
+    urlparse,
     )
 
-from pyramid.exceptions import BadCSRFToken
+from pyramid.exceptions import BadCSRFOrigin, BadCSRFToken
```
bertjwregeer: I would prefer these on multiple lines, with tuple style importing, listed in alphabetical order.

pyramid/viewderivers.py
```diff
@@ -6,7 +6,7 @@
     )
 
 from pyramid.security import NO_PERMISSION_REQUIRED
-from pyramid.session import check_csrf_token
+from pyramid.session import check_csrf_origin, check_csrf_token
```
bertjwregeer: Same here, imports across multiple lines, alphabetical order. Most of the Pyramid source code has that for multiple imports.

dstufft:
Ok, unless code review finds something that needs addressed I think this should be ready to merge!

bertjwregeer:
:+1:

@mmerickel I don't believe that adding this will break anything for anyone, and I've compared it against the Django version and it is similar (and has the same effect).

pyramid/session.py
```diff
@@ -101,6 +108,77 @@ def signed_deserialize(serialized, secret, hmac=hmac):
 
     return pickle.loads(pickled)
 
+
+def check_csrf_origin(request):
```
mmerickel: I would like to see this function more reusable in a similar vein to `check_csrf_token`. For example a default signature like `check_csrf_origin(request, trusted_origins=None, raises=True)` where trusted_origins overrides the usage of the setting.

pyramid/session.py
```diff
@@ -101,6 +108,77 @@ def signed_deserialize(serialized, secret, hmac=hmac):
...
+        # Ensure that our Referer is also secure.
+        if originp.scheme != "https":
+            raise BadCSRFOrigin(
+                "Referer checking failed - Referer is insecure while host is "
+                "secure."
+            )
+
+        # Determine which origins we trust, which by default will include the
+        # current origin.
+        trusted_origins = request.registry.settings.get(
```
mmerickel: Where is this being cast from an ini string syntax to a list? I can't find it if it's being done right now. This probably needs to be wrapped in `pyramid.settings.aslist`.

mmerickel:
Not very happy about all the whitespace changes here. It made this significantly more tedious to review and github's little `?w=1` doesn't appear to be working. This type of change belongs in a separate PR.

mmerickel:
Found a couple issues in comments but otherwise the feature looks good.

dstufft:
Ok. I adjusted this to take into account the feedback and the tests are passing again.

mmerickel:
Nice, thank you!

PR #2517, opened 2016-04-18
disable csrf checking on all exception views unless explicitly turned on

mmerickel:
Use case: User turns on auto csrf checks and registers exception views.
Issue: User must remember to turn off csrf checks on exception views or they will raise csrf errors instead of handling the original exception.
Solution in this PR: Disable csrf checks on exception views by default.
Control: User can still turn on csrf checks on exception views if they really want by setting `require_csrf=True` explicitly on the exception view.

Basically the csrf checks on exception views are too strict causing inadvertent raising of `BadCSRFToken` exceptions while handling the original exception resulting in the new csrf exception being propagated out of pyramid and up the middleware chain.

ping @dstufft @bertjwregeer

### DIFF
--- a/pyramid/tests/test_viewderivers.py
+++ b/pyramid/tests/test_viewderivers.py
@@ -1297,6 +1297,64 @@ class TestDeriveView(unittest.TestCase):
         result = view(None, request)
         self.assertTrue(result is response)
 
+    def test_csrf_view_skipped_by_default_on_exception_view(self):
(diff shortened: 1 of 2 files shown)


bertjwregeer:
:+1: This is something I ran into while working on my app, I ended up explicitly setting `require_csrf` on my views, so that my exception views wouldn't get touched, but this is better.

PR #2518, opened 2016-04-18
replace pyramid.require_default_csrf setting with config.set_default_csrf_options

mmerickel:
I realized that the csrf options should not really change per-environment and were more similar to default permissions. I'm removing the INI setting and adding `config.set_default_csrf_options` instead. Bonus aspect is that it will allow changing the default header and safe methods if desired.

I don't currently plan to remove the `pyramid.csrf_trusted_origins` setting but it's also an option. I do expect the origins to be different per-environment (dev vs prod) but people could handle it themselves by passing it to `set_default_csrf_options`.

Thoughts?

### DIFF
--- a/pyramid/viewderivers.py
+++ b/pyramid/viewderivers.py
@@ -15,6 +15,7 @@ from pyramid.response import Response
 from pyramid.interfaces import (
     IAuthenticationPolicy,
     IAuthorizationPolicy,
+    IDefaultCSRFOptions,
     IDebugLogger,
     IResponse,
     IViewMapper,
@@ -22,7 +23,6 @@ from pyramid.interfaces import (
     )
 
 from pyramid.compat import (
-    string_types,
     is_bound_method,
     is_unbound_method,
     )
@@ -38,10 +38,6 @@ from pyramid.exceptions import (
     PredicateMismatch,
     )
 from pyramid.httpexceptions import HTTPForbidden
-from pyramid.settings import (
-    falsey,
-    truthy,
-)
 from pyramid.util import object_description
 from pyramid.view import render_view_to_response
 from pyramid import renderers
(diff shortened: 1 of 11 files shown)


bertjwregeer:
Leaving `pyramid.csrf_trusted_origins` makes sense, this way configuration can still be defined with the rest of the configuration.

mmerickel:
I'm going back and forth on how deep the integration should go. I'd like to have `check_csrf_token` use these options automatically for every call to it similar to how `check_csrf_origin` works right now (always checks the setting unless overridden). That way the new defaults are always used unless you explicitly specify them. That's a minor backward-incompatibility but does require opt-in (by calling `config.set_default_csrf_options` from the user to trigger so it feels like a win to me.

If I went that route then I would add another parameter like `config.set_default_csrf_options(autocheck=True, token='csrf_token', header='X-CSRF-Token', autocheck_ignore_safe_request_methods=['GET', 'HEAD', 'TRACE', 'OPTIONS'])`. A user could then set `autocheck=False` but still influence the token/header being set.

bertjwregeer:
I like the idea that I as a user can influence the header/token name from a setting. Other than that, I really have no real preference either way.

mmerickel:
Alright I've overhauled the automatic csrf api to be simpler.

- I removed the `pyramid.require_default_csrf` setting.
- I removed the ability to specify a token name to `add_view(require_csrf=...)`.
- `require_csrf` may now be `True`, `False`, or `None`.
- If the value is `None` then it falls through to the setting used in `config.set_default_csrf_options(require_csrf=..)`.
- The token name, header and safe methods can only be overridden globally via the new `set_default_csrf_options`.

Please review!

pyramid/viewderivers.py
```diff
@@ -464,48 +460,38 @@ def decorated_view(view, info):
...
-    explicit_val = _parse_csrf_setting(
-        info.options.get('require_csrf'),
-        'View option "require_csrf"')
-    resolved_val = explicit_val
-    if (explicit_val is True and default_val) or explicit_val is None:
-        resolved_val = default_val
-    if resolved_val is True:
-        resolved_val = 'csrf_token'
+    explicit_val = info.options.get('require_csrf')
+    defaults = info.registry.queryUtility(IDefaultCSRFOptions)
```
bertjwregeer: Wouldn't it be simpler to always register a default utility, that can be overridden, rather than having to redefine the defaults here?

If we ever decide to change the safe_methods or the header, we need to change it in two places.

bertjwregeer:
Besides a minor nitpick about duplication, this looks great. 👍

pyramid/viewderivers.py
mmerickel: Pyramid tends to be fairly lazy about this stuff. I think no matter what you will have 2 spots unless I move things out of the signature.

bertjwregeer: No worries :-)

PR #2519, opened 2016-04-18
Don't fail CSRF check with subdomains

di:
This fixes a bug introduced in #2501 which fails the CSRF origin check due to incorrectly using `request.domain` instead of `request.host`.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -266,3 +266,5 @@ Contributors
 - Marcin Raczyński, 2016/01/26
 
 - Arian Maykon de A. Diógenes, 2016/04/13
+
+- Dustin Ingram, 2016/04/18
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -182,10 +182,7 @@ def check_csrf_origin(request, trusted_origins=None, raises=True):
                     "pyramid.csrf_trusted_origins", [])
             )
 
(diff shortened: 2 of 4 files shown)


dstufft:
@di Is this to fix the CSRF error on Warehouse?

di:
@dstufft It would fix that, I believe. Am I missing something, or is this definitely a bug in Pyramid?

dstufft:
Ah. The reason I ask is because I think that #2512 should fix Warehouse (It just hasn't been released yet, and I hadn't gone to the effort to pull in the branch into Warehouse instead of using 1.7a1 from PyPI). I was curious if you had tried Pyramid@master and found it still broken (and thus my fix was incomplete) or if it was just a race condition between two fixes.

Looking at your PR here, I *think* that it's essentially a no-op as far as what would pass or fail, except that a request with a Referrer of ``https://example.com/`` and a Host of ``example.com:443`` would fail with it (but I don't know of any clients that actually send the port if it is {80, 443}).

di:
Ah, @dstufft you're right. I didn't see #2512 and the test I wrote is bad.

Sorry for the noise!

dstufft:
No worries! FWIW if you're working on Warehouse and want to bypass the problem locally, you can just stick a ``return view`` at the start of the function here: https://github.com/pypa/warehouse/blob/master/warehouse/csrf.py#L22

PR #2520, opened 2016-04-19
Fix tests on Windows

Cito:
Two tests do not pass under Windows where backslashes instead of slashes are used in file paths. This fix makes the tests platform independent.


Cito:
Actually wanted to push a fix for the issue with int ports in session.py but it was already fixed :+1:

mmerickel:
Thanks, I need to look into configuring appveyor to get our test suite running on windows. This will serve as a good reminder.

mmerickel:
Cool, I got appveyor setup and tests pass. Thank you!

Cito:
:+1:

PR #2523, opened 2016-04-23
import/docstring adjustments in decorator module

int3l:
Straight forward change:
- import adjustment
- docstring adjustment + additional note

### DIFF
--- a/pyramid/decorator.py
+++ b/pyramid/decorator.py
@@ -19,15 +19,14 @@ class reify(object):
     And usage of Foo:
 
     >>> f = Foo()
-    >>> print(f.jammy)
+    >>> v = f.jammy
     'jammy called'
+    >>> print(v)
     1
-    >>> print(f.jammy)
+    >>> f.jammy
     1
     >>> # jammy func not called the second time; it replaced itself with 1
-
-    Note: reassignment is possible
-
+    >>> # Note: reassignment is possible
     >>> f.jammy = 2
     >>> f.jammy
     2
--- a/pyramid/decorator.py
+++ b/pyramid/decorator.py
@@ -1,4 +1,4 @@
-import functools
+from functools import update_wrapper
 
 
 class reify(object):
@@ -19,17 +19,22 @@ class reify(object):
     And usage of Foo:
 
     >>> f = Foo()
-    >>> v = f.jammy
+    >>> print(f.jammy)
     'jammy called'
-    >>> print(v)
     1
-    >>> f.jammy
+    >>> print(f.jammy)
     1
     >>> # jammy func not called the second time; it replaced itself with 1
+
+    Note: reassignment is possible
+
+    >>> f.jammy = 2
+    >>> f.jammy
+    2
     """
     def __init__(self, wrapped):
         self.wrapped = wrapped
-        functools.update_wrapper(self, wrapped)
+        update_wrapper(self, wrapped)
 
     def __get__(self, inst, objtype=None):
         if inst is None:


pyramid/decorator.py
```diff
@@ -19,17 +19,22 @@ def jammy(self):
     And usage of Foo:
 
     >>> f = Foo()
-    >>> v = f.jammy
+    >>> print(f.jammy)
```
tseaver: This change doesn't feel right:  having `jammy` do the print means we should be doing something like assigning `v = f.jammy`, and then maybe `print(v)`.

tseaver:
Except for my concern about mixing the side-effect `print` with the return value from `jammy()`, LGTM.

tseaver:
I should also express that I am also worried about having example snippets in docstrings which don't get exercised via Tox/Travis:  we aren't running `sphinx -b doctests` in either one.

stevepiercy:
@tseaver I've had the same concern, but didn't know it was possible until your comment. Do you have a configuration example where you run `sphinx -b doctests` that we could reuse for Pyramid?

I don't see a builder for doctests in Pyramid's `Makefile`, although I found the documentation for [sphinx.ext.doctest](http://www.sphinx-doc.org/en/stable/ext/doctest.html).

tseaver:
@stevepiercy the `zope.*` packages routinely test any doctest examples which are pulled into Sphinx, e.g.: [zope.interface](https://github.com/zopefoundation/zope.interface/blob/ab5b2448bd901e5d6a25a61cda67c44c0da7f904/tox.ini#L43).

stevepiercy:
@tseaver cool! That revealed some code blocks in .rst files that don't pass doctest.

[Sorry for hijacking this issue, but...] For a snippet like the following in `docs/narr/hooks.rst`, should the docs be "fixed" so tests pass, or use the lexer `.. code-block:: pycon` so that doctest ignores it:

```pycon
   >>> request.total(1, 2, 3)
   6
   >>> request.prop
   getting the property
   the property
   >>> request.prop
   the property
```

Currently running `make doctest SPHINXBUILD=$VENV/bin/sphinx-build` results with the traceback:

```
Warning, treated as error:
WARNING: **********************************************************************
File "narr/hooks.rst", line ?, in default
Failed example:
    request.total(1, 2, 3)
Exception raised:
      File "/Users/stevepiercy/.pyenv/versions/3.5.0/lib/python3.5/doctest.py", line 1320, in __run
        compileflags, 1), test.globs)
      File "<doctest default[0]>", line 1, in <module>
        request.total(1, 2, 3)
    NameError: name 'request' is not defined
```

I imagine if we decide to test snippets in .rst files, then we'll need to write `.. testsetup::` code, too. I'm willing to doing the work, but wondered whether it's a common or best practice.

int3l:
@tseaver I see your point, thanks for the comments.

stevepiercy:
The proposed changes offer another example. Running `make doctest SPHINXBUILD=$VENV/bin/sphinx-build` yields:

```python
Warning, treated as error:
WARNING: **********************************************************************
File "api/decorator.rst", line 29, in default
Failed example:
    f = Foo()
Exception raised:
      File "/Users/stevepiercy/.pyenv/versions/3.5.0/lib/python3.5/doctest.py", line 1320, in __run
        compileflags, 1), test.globs)
      File "<doctest default[0]>", line 1, in <module>
        f = Foo()
    NameError: name 'Foo' is not defined


make: *** [doctest] Error 1
```

Without knowing any better or the history of using doctest in Pyramid, I'd be inclined to change this to use the `pycon` lexer.

tseaver:
@stevepiercy even better would be make the definition of `class Foo` also use doctest format.  Still, not a reason to block this PR, as it is broken the same way on `master.

Issue #2533, opened 2016-04-26
figure out how to normalize view introspectables


mmerickel:
The `permission=` and `config.set_default_permission` are normalized in `config.add_view` where the default permission is actually pulled out and set as the value in `info.options`. This is inconsistent with the `require_csrf=` and `config.set_default_csrf_options` where if the default is used then `require_csrf` does not show up in the introspectable (not good). Unfortunately if we normalize the value in `add_view` then:

1) It's not possible for anyone else to do this except core.
2) It's no longer possible to tell if `require_csrf` was set on the view or not. This is necessary to determine whether to enforce the checks on the exception view, which are only used if the user explicitly sets the value.

I'm currently thinking it may be necessary to define `info.raw_options` versus `info.options` or something, to tell exactly what options were passed to the `config.add_view` versus which ones were normalized. However, again, this normalization is something only Pyramid can do (inside `config.add_view`), which is not ideal when thinking about extension authors.

We could also stop normalizing permission. However, I'd still want the introspectable to show the final value. How to do this? We could put the introspectable in the view deriver info, and allow the derivers to write to it. So far that's the best option I have.

mmerickel:
Well after digging a bit deeper I discovered that the introspectable did not actually show the final value, which means I was able to refactor the logic to work exactly like csrf does. #2534 demonstrates this.

The second part of this issue is still valid which is that it'd be nice if view derivers could affect the introspectable. If they could, we would be able to move a fairly large chunk of logic out of `config.add_view` and into the derivers themselves. This would allow config.add_view to focus almost exclusively on the crazy MultiView it is constructing, and less so on bookkeeping around derivers. For example, almost all of this junk at the end could be in the individual derivers: https://github.com/Pylons/pyramid/blob/master/pyramid/config/views.py#L1030-L1065

mmerickel:
I might need to abandon my hopes of having view derivers register their own introspectables. The initial thought was to add `info.view_intr` which the derivers could use to relate their own introspectables to the view. I've been staring at it for a while now and there are a few issues.

1. There may not be a `view_intr` to relate to in the case of calling `config.derive_view` directly instead of coming from `config.add_view`. In this case no introspectables can be registered.
[...]

PR #2534, opened 2016-04-26
do not enforce default permissions on exception views

mmerickel:
- this normalizes the behavior to work similar to require_csrf

- if an explicit `permission=` is set on the view it will still be enforced,
  this just affects a default permission via `config.set_default_permission`

- `permission=NO_PERMISSION_REQUIRED` was already forced on for notfound
  and forbidden views, this just helps out with other exception views

This partially solves concerns with #2533


bertjwregeer:
:+1:

PR #2559, opened 2016-05-08
backport scaffold defaults for py.test and coverage; link checking

stevepiercy:
1.7 branch rebase master


bertjwregeer:
This seems to be pulling in an awful lot of commits...

stevepiercy:
Indeed. Yet the end result under https://github.com/Pylons/pyramid/pull/2559/files looks good. I might need some help with git commands to do it the right way.

mmerickel:
What? Did you just rebase 1.7-branch on top of master? I'm confused what your comment says but if it's that, then no.

mmerickel:
If there are some commits needing to be backported to 1.7-branch then here's the approach I would take.

1) make a new branch (foo) off of 1.7-branch
2) cherry pick the missing commits into foo.
3) rebase that branch however you see fit so that the changes are in a small number of logical commits (usually the commit should link back to the original PR on master via a comment). This squashing via rebase isn't enforced very much by our project, but that's how you can make it as clean as possible.
4) when you're happy with that branch, submit a PR to 1.7-branch with those commits.

Basically after you get all the commits into a branch, you can do an interactive rebase (git rebase --interactive <base commit>` on those commits to reorder them and squash them as you see fit. The `base commit` is the current 1.7-branch. The cherry picking is the hardest part. You can do it with `git rebase --onto` but it's pretty hairy to get around merge commits and so I tend to just manually cherry pick the missing changes via diffing commits to see if the changes are missing.

stevepiercy:
closing this in favor of another clean PR, coming soon.

PR #2591, opened 2016-05-20
Change column `Page.data` to `Text` in wiki2 tutorial

viniciusban:
Column `data` in `Page` class is `Integer` but should be `Text` in wiki2 tutorial.

### DIFF
--- a/docs/tutorials/wiki2/src/authentication/tutorial/models/page.py
+++ b/docs/tutorials/wiki2/src/authentication/tutorial/models/page.py
@@ -14,7 +14,7 @@ class Page(Base):
     __tablename__ = 'pages'
     id = Column(Integer, primary_key=True)
     name = Column(Text, nullable=False, unique=True)
-    data = Column(Integer, nullable=False)
+    data = Column(Text, nullable=False)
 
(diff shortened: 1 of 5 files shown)


mmerickel:
How'd that get through? Anyway, nice catch. LGTM

viniciusban:
Because of this: http://www.sqlite.org/different.html

tl;dr: almost everything is stored as string. See the "Manifest typing" section.

mmerickel:
I'm aware of sqlite's typing issues but usually SQLAlchemy will complain somewhere along the lines. In this case it looks like it just shuffles along the data from the driver however.

viniciusban:
Indeed. :-(

stevepiercy:
Huh, that's interesting. Using PyCharm's database browser on the database `tutorial.sqlite`, it shows `data` as type `INTEGER`. I expected `TEXT`.

I'll want to review all these docs again as I am certain there are more occurrences such as in the output of database initialization, but I'll create a new issue and do that in another PR.

stevepiercy:
Thanks @viniciusban

viniciusban:
Anytime, Sir.

PR #2605, opened 2016-05-31
Update user.py to prevent attribute error in check_password function

pranaysashank:
The original file is throwing Attribute Error in check_password function with python 3 interpreter. (didn't check with python 2)
password_hash is of type bytes, bytes object doesn't have attribute encode.
The error is evaded by removing the encoding function.


bertjwregeer:
Please see this PR as to why the change was made: https://github.com/Pylons/pyramid/pull/2576

We need to convert this to a `LargeBinary` and make sure that it works on both Python 2/3.

stevepiercy:
Closing in favor of new issue #2623

Issue #735, opened 2012-11-29
Enhancement: can ACLAuthorizationPolicy support __acl__ as a callable?


twillis:
Being a python newb I did not know that any error in property getters result in an AttributeError. I understand the rationale behind it, but it makes me a little anxious to have my "security" code blowing up silently and leaving the traversal "layer" in an unexpected state. 

What would be nice is if  the ACLAuthorizationPolicy could look out for **acl** that is a callable and deal with it that way. then errors will bubble up properly.

For now, I've embraced and extended for batterii's innovation platform, but I'm of the opinion it might be desired in pyramid as well. 

https://gist.github.com/4169503

PR #2613, opened 2016-06-02
Avoid silent dismissal of __acl__ AttributeErrors

mike-perdide:
With current code, an AttributeError raised in an __acl__ property will be silently dismissed, making code debug harder: what looks like a permission problem can actually be a bug. Instead, here we check if the location has the __acl__ attribute, and if not we keep looking.

I have run the test suite for this modification and it passed but I'd be happy to write whatever checks seem necessary to make this change acceptable as I think this part of the Pyramid code can lead to costly debug sessions :)

### DIFF
--- a/pyramid/authorization.py
+++ b/pyramid/authorization.py
@@ -75,12 +75,12 @@ class ACLAuthorizationPolicy(object):
         acl = '<No ACL found on any object in resource lineage>'
 
         for location in lineage(context):
-            try:
-                acl = location.__acl__
-            except AttributeError:
+            if not hasattr(location, '__acl__'):
                 continue
+            
+            acl = location.__acl__
 
-            if acl and callable(acl):
+            if callable(acl):
                 acl = acl()
 
             for ace in acl:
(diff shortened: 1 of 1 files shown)


mike-perdide:
Ok, so it turns out that hasattr also silently dismisses AttributeErrors.
Would the following be deemed dirty Python ?
`if "__acl__" in dir(location)`

bertjwregeer:
What problem are you trying to solve? If the context has no attr named `__acl__` then it will be ignored.

When are you seeing an `AttributeError` when the attribute does exist?

mmerickel:
This is a duplicate issue to https://github.com/Pylons/pyramid/issues/735 where we chose to allow `__acl__` to be a callable to get around that issue. Did you know about this feature? I'm not sure there's a good way to solve this otherwise.

mike-perdide:
Thanks for your replies:
- @bertjwregeer the issue is if I have defined `__acl__` as a property, and if the code I run raises an AttributeError, then Pyramid considers that it wasn't able to fetch the `__acl__` for the context and carries on, usually to fall back on the default Deny.
- @mmerickel I'll look into the ticket you mentionned.

mike-perdide:
However, it's True that the Pyramid documentation doesn't encourage coders to write `__acl__` properties, so it might just be a RTFM issue :)

mmerickel:
Yes if there's some spots where attributes are used I would suggest we update them to use callables.

mike-perdide:
A grep in the Pyramid project shows no such spot, so I guess we can close this issue.

mmerickel:
Okay cool... we can also add a note somewhere if you find a good spot.

Issue #2609, opened 2016-06-02
Route name isn't in the view deriver options


sontek:

mmerickel:
We basically need to add support for passing the route_name into `derive_view`. https://github.com/Pylons/pyramid/blob/382f93e2bfec5563587e306fda3fd34759314300/pyramid/config/views.py#L886

PR #2616, opened 2016-06-02
Add route name to view deriver options. Closes: #2609

m-martinez:
Passed `route_name` to `_derive_view` method so that it can be added to the `DerivedViewInfo` options.


mmerickel:
LGTM!

PR #2646, opened 2016-06-13
Fix flake8's noqa directive in documentation and scaffolds

jcbohin:
They are a [few (9 as of today)](https://github.com/Pylons/pyramid/search?utf8=✓&q=%22flake8%3A+noqa%22&type=Code) occurrences of the following snippet in docs and scaffolds:

```python
from .a import b  # flake8: noqa
```
*(replace `from .a import b` with any line of code)*

As mentioned in [Flake8 documentation](http://flake8.pycqa.org/en/stable/):

> files that contain this line are skipped:
> `# flake8: noqa`
> lines that contain a `# noqa` comment at the end will not issue warnings.

So the example above should be written
```python
from .a import b  # noqa
```
to only ignore the line instead of the whole file.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -276,3 +276,5 @@ Contributors
 - Marco Martinez, 2016/06/02
 
 - Cris Ewing, 2016/06/03
+
+- Jean-Christophe Bohin, 2016/06/13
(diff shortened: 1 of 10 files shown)


mmerickel:
LGTM

stevepiercy:
@jcbohin thank you!

jcbohin:
My pleasure :)

thanks @mmerickel @stevepiercy and all the team for all the good work on pyramid and the reactivity on GH.

Issue #1646, opened 2015-04-24
An analogue of pyramid.view.view_config which registers an Error View.


zupo:
Much like `notfound_view_config` and `forbidden_view_config` I believe we should have an `error_view_config` which is analogues be the same as registering a catch-all exception view. 

I can start working on this, but I'd like to get a few +1's that it's something that could get merged.

mcdonc:
How would it differ from:

```
   @view_config(context=SomeException)
```

mmerickel:
I guess he's saying it'd be the same thing.

I think this is not worthwhile right now because the `notfound` and `forbidden` views actually do some useful things in their api. For example `append_slash`. This seems like it would be just adding an api to make it more obvious to people that the feature exists.

mmerickel:
This would need to be named `exc_view_config` or `exception_view_config` instead of "error" to make it clear its purpose as well. I think it's potentially interesting to have an `exception_view_config` that had an optional argument `context=` or `exc=` that defaulted to `Exception` but could be overridden with other exception types.

mmerickel:
So far the only reasons I can think of for adding an `exception_view_config(exc=Exception)` is because 1) it is more obvious and 2) it may set a default permission of `NO_PERMISSION_REQUIRED`. However I'm not sure this is compelling enough!

mmerickel:
I think I came up with a reason for this to exist. Currently pyramid has no good way to register a view as an exception-view only. All views are registered as both regular views and optionally as an exception view, but never *just* as an exception view.

As I've built out the view derivers there's certain derivers that I want to only affect exception views and currently the only way to do it is via a per-request check for request.exception. I can't do it at config-time to avoid affecting normal views.

This feature could be a thin wrapper around a new `exception_only` option to `config.add_view()`. This would also be used by notfound and forbidden views to their benefit.

There's some subtleties in pyramid's approach I mentioned in the first paragraph though that may need to be worked out.

mmerickel:
We would want an exc-only and regular+exc view registration to conflict on the same parameters for sure. The part that I'm struggling with is the effect of supporting an exc-only view and regular+exc view on the same MultiView instance. Currently the MultiView doesn't need to think about exc vs regular views, it just thinks about views. It would need to grow some knowledge that some of its views are exc-only and then check request.exception prior to considering them in its match algorithm. Alternately `config.add_view` would need to register 2 separate multiviews but I think that is a non-starter for many reasons.

mmerickel:
[...]

PR #2660, opened 2016-06-30
Exception-only option when registering views

latteier:
I don't think that this work is fully done yet, but I'm creating a PR to start collaboration and discussion.

I think that code could use some re-factoring. I think that there's too much shared between the exception view registration and the not-found view and forbidden view registration. Probably it could be changed to have less repeated code between those three.

I have some time to work on this going forward, but unfortunately a lot less than I'd like.

Discussion that started this work here: https://github.com/Pylons/pyramid/issues/1646

### DIFF
--- a/docs/narr/hooks.rst
+++ b/docs/narr/hooks.rst
@@ -1639,7 +1639,8 @@ the user-defined :term:`view callable`:
   Enforce the ``permission`` defined on the view. This element is a no-op if no
   permission is defined. Note there will always be a permission defined if a
   default permission was assigned via
-  :meth:`pyramid.config.Configurator.set_default_permission`.
+  :meth:`pyramid.config.Configurator.set_default_permission` unless the
+  view is an :term:`exception view`.
 
   This element will also output useful debugging information when
   ``pyramid.debug_authorization`` is enabled.
@@ -1649,7 +1650,8 @@ the user-defined :term:`view callable`:
   Used to check the CSRF token provided in the request. This element is a
   no-op if ``require_csrf`` view option is not ``True``. Note there will
   always be a ``require_csrf`` option if a default value was assigned via
-  :meth:`pyramid.config.Configurator.set_default_csrf_options`.
+  :meth:`pyramid.config.Configurator.set_default_csrf_options` unless
+  the view is an :term:`exception view`.
 
 ``owrapped_view``
 
@@ -1695,6 +1697,8 @@ around monitoring and security. In order to register a custom :term:`view
 deriver`, you should create a callable that conforms to the
 :class:`pyramid.interfaces.IViewDeriver` interface, and then register it with
 your application using :meth:`pyramid.config.Configurator.add_view_deriver`.
+The callable should accept the ``view`` to be wrapped and the ``info`` object
+which is an instance of :class:`pyramid.interfaces.IViewDeriverInfo`.
 For example, below is a callable that can provide timing information for the
 view pipeline:
 
@@ -1744,6 +1748,21 @@ But this view *will* have timing information added to the response headers:
 View derivers are unique in that they have access to most of the options
 passed to :meth:`pyramid.config.Configurator.add_view` in order to decide what
 to do, and they have a chance to affect every view in the application.
+
+.. _exception_view_derivers:
+
+Exception Views and View Derivers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A :term:`view deriver` has the opportunity to wrap any view, including
+an :term:`exception view`. In general this is fine, but certain view derivers
+may wish to avoid doing certain things when handling exceptions. For example,
+the ``csrf_view`` and ``secured_view`` built-in view derivers will not perform
+security checks on exception views unless explicitly told to do so.
+
+You can check for ``info.exception_only`` on the
+:class:`pyramid.interfaces.IViewDeriverInfo` object when wrapping the view
+to determine whether you are wrapping an exception view or a normal view.
 
 Ordering View Derivers
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/narr/viewconfig.rst
+++ b/docs/narr/viewconfig.rst
@@ -34,7 +34,7 @@ determine the set of circumstances which must be true for the view callable to
 be invoked.
 
 A view configuration statement is made about information present in the
-:term:`context` resource and the :term:`request`.
+:term:`context` resource (or exception) and the :term:`request`.
 
 View configuration is performed in one of two ways:
 
@@ -306,8 +306,25 @@ configured view.
   represented class or if the :term:`context` resource provides the represented
   interface; it is otherwise false.
 
+  It is possible to pass an exception class as the context if your context may
+  subclass an exception. In this case **two** views will be registered. One
+  will match normal incoming requests and the other will match as an
+  :term:`exception view` which only occurs when an exception is raised during
+  the normal request processing pipeline.
+
   If ``context`` is not supplied, the value ``None``, which matches any
   resource, is used.
+
+``exception_only``
+
+  When this value is ``True`` the ``context`` argument must be a subclass of
+  ``Exception``. This flag indicates that only an :term:`exception view` should
+  be created and that this view should not match if the traversal
+  :term:`context` matches the ``context`` argument. If the ``context`` is a
+  subclass of ``Exception`` and this value is ``False`` (the default) then a
+  view will be registered to match the traversal :term:`context` as well.
+
+  .. versionadded:: 1.8
 
 ``route_name``
   If ``route_name`` is supplied, the view callable will be invoked only when
--- a/pyramid/config/views.py
+++ b/pyramid/config/views.py
@@ -1613,6 +1613,71 @@ class ViewsConfiguratorMixin(object):
 
     set_notfound_view = add_notfound_view # deprecated sorta-bw-compat alias
 
+    @viewdefaults
+    @action_method
+    def add_exception_view(
+        self,
+        view=None,
+        context=None,
+        attr=None,
+        renderer=None,
+        wrapper=None,
+        route_name=None,
+        request_type=None,
+        request_method=None,
+        request_param=None,
+        containment=None,
+        xhr=None,
+        accept=None,
+        header=None,
+        path_info=None,
+        custom_predicates=(),
+        decorator=None,
+        mapper=None,
+        match_param=None,
+        **view_options
+            ):
+        """ Add a view for an exception to the current configuration state.
+        The view will be called when Pyramid or application code raises an
+        the given exception.
+
+        .. versionadded:: 1.8
+        """
+        for arg in (
+            'name', 'permission', 'for_', 'http_cache',
+            'require_csrf', 'exception_only',
+        ):
+            if arg in view_options:
+                raise ConfigurationError(
+                    '%s may not be used as an argument to add_exception_view'
+                    % arg
+                    )
+        if context is None:
+            raise ConfigurationError('context exception must be specified')
+        settings = dict(
+            view=view,
+            context=context,
+            wrapper=wrapper,
+            renderer=renderer,
+            request_type=request_type,
+            request_method=request_method,
+            request_param=request_param,
+            containment=containment,
+            xhr=xhr,
+            accept=accept,
+            header=header,
+            path_info=path_info,
+            custom_predicates=custom_predicates,
+            decorator=decorator,
+            mapper=mapper,
+            match_param=match_param,
+            route_name=route_name,
+            permission=NO_PERMISSION_REQUIRED,
+            require_csrf=False,
+            exception_only=True,
+            )
+        return self.add_view(**settings)
+
     @action_method
     def set_view_mapper(self, mapper):
         """
--- a/pyramid/config/views.py
+++ b/pyramid/config/views.py
@@ -9,12 +9,11 @@ from zope.interface import (
     implementedBy,
     implementer,
     )
-
 from zope.interface.interfaces import IInterface
 
 from pyramid.interfaces import (
-    IException,
     IExceptionViewClassifier,
+    IException,
     IMultiView,
     IPackageOverrides,
     IRendererFactory,
@@ -503,7 +502,20 @@ class ViewsConfiguratorMixin(object):
           if the :term:`context` provides the represented interface;
           it is otherwise false.  This argument may also be provided
           to ``add_view`` as ``for_`` (an older, still-supported
-          spelling).
+          spelling). If the view should **only** match when handling
+          exceptions then set the ``exception_only`` to ``True``.
+
+        exception_only
+
+          .. versionadded:: 1.8
+
+          When this value is ``True`` the ``context`` argument must be
+          a subclass of ``Exception``. This flag indicates that only an
+          :term:`exception view` should be created and that this view should
+          not match if the traversal :term:`context` matches the ``context``
+          argument. If the ``context`` is a subclass of ``Exception`` and
+          this value is ``False`` (the default) then a view will be
+          registered to match the traversal :term:`context` as well.
 
         route_name
 
@@ -685,7 +697,7 @@ class ViewsConfiguratorMixin(object):
                 obsoletes this argument, but it is kept around for backwards
                 compatibility.
 
-        view_options:
+        view_options
 
           Pass a key/value pair here to use a third-party predicate or set a
           value for a view deriver. See
@@ -701,14 +713,6 @@ class ViewsConfiguratorMixin(object):
 
              Support setting view deriver options. Previously, only custom
              view predicate values could be supplied.
-
-        exception_only
-
-          .. versionadded:: 1.8
-
-          A boolean indicating whether the view is registered only as an
-          exception view. When this argument is true, the view context must
-          be an exception.
 
         """
         if custom_predicates:
@@ -768,13 +772,14 @@ class ViewsConfiguratorMixin(object):
                 raise ConfigurationError(
                     'request_type must be an interface, not %s' % request_type)
 
-        if exception_only and not isexception(context):
-            raise ConfigurationError(
-                'context must be an exception when exception_only is true'
-                )
-
         if context is None:
             context = for_
+
+        isexc = isexception(context)
+        if exception_only and not isexc:
+            raise ConfigurationError(
+                'view "context" must be an exception type when '
+                '"exception_only" is True')
 
         r_context = context
         if r_context is None:
@@ -811,6 +816,7 @@ class ViewsConfiguratorMixin(object):
             # is.  It can't be computed any sooner because thirdparty
             # predicates/view derivers may not yet exist when add_view is
             # called.
+            predlist = self.get_predlist('view')
             valid_predicates = predlist.names()
             pvals = {}
             dvals = {}
@@ -849,6 +855,7 @@ class ViewsConfiguratorMixin(object):
         view_intr.update(dict(
             name=name,
             context=context,
+            exception_only=exception_only,
             containment=containment,
             request_param=request_param,
             request_methods=request_method,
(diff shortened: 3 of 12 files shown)


pyramid/config/views.py
```diff
@@ -1598,6 +1615,71 @@ def notfound(request):
...
+        path_info=None,
+        custom_predicates=(),
+        decorator=None,
+        mapper=None,
+        match_param=None,
+        **view_options
+            ):
+        """ Add a view for an exception to the current configuration state.
+        The view will be called when Pyramid or application code raises an
+        the given exception.
```
stevepiercy: Remove "an", rewrap "the".

pyramid/view.py
```diff
@@ -463,6 +463,58 @@ def callback(context, name, ob):
...
+    .. versionadded:: 1.8
+
+    An analogue of :class:`pyramid.view.view_config` which registers an
+    exception view.
+
+    The exception_view_config constructor requires an exception context, and
+    additionally accepts most of the same arguments as the constructor of
+    :class:`pyramid.view.view_config`.  It can be used in the same places,
+    and behaves in largely the same way, except it always registers an exception
+    view instead of a 'normal' view.
```
stevepiercy: Should use " instead of ' around "normal".

mmerickel:
Nice, this is an excellent start. Thanks @latteier!

I agree that both forbidden and notfound can be rewritten to use `config.add_exception_view` internally.

The remaining work to be done on this is some subtle under-the-hood stuff:

1) Handle conflict resolution / view overriding between exc-only and non-exc-only variants of the same view. A lot of this is to continue to handle `config.add_view(context=SomeExceptionType, ...)` and `config.add_view(context=SomeExceptionType, exception_only=True, ...)` where predicates may or may not be different. This may require some complexity on the `MultiView` to know which is which.

2) Add `ViewDeriverInfo.is_exception_only` so that derivers can optimize their pipeline based on the view they're wrapping. This will allow the secured_view and csrf_view to drop the `if request.exception` checks completely in most cases.

pyramid/view.py
```diff
@@ -463,6 +463,58 @@ def callback(context, name, ob):
...
+        info = self.venusian.attach(wrapped, callback, category='pyramid')
+
+        if info.scope == 'class':
+            # if the decorator was attached to a method in a class, or
+            # otherwise executed at class scope, we need to set an
+            # 'attr' into the settings if one isn't already in there
+            if settings.get('attr') is None:
+                settings['attr'] = wrapped.__name__
+
+        settings['_info'] = info.codeinfo # fbo "action_method"
```
stevepiercy: PEP8: two spaces between code and # for inline comments. Also what is "fbo"? Please avoid acronyms, or whatever it is, and be explicit.

pyramid/view.py
```diff
@@ -463,6 +463,58 @@ def callback(context, name, ob):
...
+        def callback(context, name, ob):
+            config = context.config.with_package(info.module)
+            config.add_exception_view(view=ob, **settings)
+
+        info = self.venusian.attach(wrapped, callback, category='pyramid')
+
+        if info.scope == 'class':
+            # if the decorator was attached to a method in a class, or
+            # otherwise executed at class scope, we need to set an
+            # 'attr' into the settings if one isn't already in there
```
stevepiercy: s/into/in

pyramid/view.py
mmerickel: he's just copying from other pre-existing pyramid code here, steve

stevepiercy: OK, @mmerickel, so it needs to be fixed in two places.

mmerickel: fbo is similar to bbb in that they have standard understandings in software but are not exactly acronyms, at least afaik... the comment is basically saying it's there for the benefit of pyramid's action feature

mmerickel:
I have dropped `exception_only=True/False` and instead added `exception=some_exception_type` which conflicts with `context=some_type`. I think this works to make `context=` accessible via `request.context` or `view(context, request)` analogous to the exception-version with `exception=` accessible via `request.exception` or `view(exception, request)` and we can update the docs similarly.

I'd say at this point this PR is mostly complete except for some subtle complexities in the MultiView and updating the permission/csrf view derivers to no-op themselves if they are wrapping exception-only views.

pyramid/view.py
mmerickel: @stevepiercy all I meant is that this does not need to be fixed as part of this PR... if the acronyms bother you then they can be fixed in a separate PR later but they're consistent with the current conventions and I don't want unrelated things changed in this PR.

stevepiercy:
@mmerickel I'll stand aside so this can be merged, then I'll clean up the docstrings per #2768.

bertjwregeer:
Does this still allow containment to work as before @mmerickel?

Right now I have:

`context=SomeException, containment=MyContext`

in some projects, would that get translated to:

`exception=SomeException, containment=MyContext`

And things will work as before?

mmerickel:
Yeah, the exception is translated to context under the hood. It's just that now in derivers there's an `exception` key alongside `context` which will only be there if `exception=` was used.

bertjwregeer:
Can we add b/w compat by checking if context isexc, and then automatically registering it as an exception view?

mmerickel:
That's already done by the current codebase. We are just adding a path for optimization if you want the view to **only** serve exceptions by enabling view derivers to do config-time structure optimizations instead of runtime checks of `request.exception`. Currently if you add `context=exc_type` it is registered both for normal and exception paths - we haven't changed that in this PR. Here's the rub that I've been working on though when I say "subtle things with MultiView" above. The issue is that we want to maintain bw-compat and support someone adding views via `context=exc_type` **and** `exception=exc_type`. For example:

```python
config.add_view(view, context=Exception, request_method='GET')
config.add_view(view, exception=Exception, request_method='POST')
```

In the old code (`s/exception/context/`), these two views would be part of the same MultiView and they would both be registered for normal requests **and** exception requests. In the new code this is an issue because you are declaring that you don't want the second view to be served for normal requests, however the context is the same and MultiViews must be unique for (context, request, name) triads. This is a corner case, but it's one I want to keep working to ensure no issues upgrading, but it's proving quite complex to work out.

mmerickel:
I actually figured out a pretty decent fix for this which is just to have separate MultiViews for exceptions and normal requests instead of sharing them. The memory impact of this should be almost nil since in most cases we are creating one or the other... only in bw-compat scenarios are we creating two multiviews.

mmerickel:
So this is done. If you look at the view deriver implementations they are slightly hairy due to the new api.

We expose `info['context']` and `info['exception']`. It is easy to tell if the view deriver serves only exceptions by testing for `info['exception'] is not None`, however it is less easy to tell if the deriver serves only non-exceptions which requires testing for `isexception(info['context'])`.

I guess it might be better to come up with something slightly more clear to avoid the `isexception` test.

stevepiercy:
I didn't notice anything that would require a change to the [first diagram in Request Processing](http://docs.pylonsproject.org/projects/pyramid/en/master/narr/router.html), but it would be good to get verification.

bertjwregeer:
Will review this later today!

pyramid/config/views.py
```diff
@@ -762,6 +782,17 @@ def view(context, request):
...
+        if exception is not None:
+            if context is not None:
+                raise ConfigurationError('view "context" and "exception" '
+                                         'arguments are mutually exclusive')
+            context = exception
+
+        isexc = isexception(context)
+        if not isexc and exception is not None:
+            raise ConfigurationError(
+                'view "exception" must be an exception type')
```
bertjwregeer: Should we have docs that are linked to here that explain that it must either be a subclass of Exception or

pyramid/config/views.py
```diff
@@ -908,8 +943,41 @@ def register(permission=permission, renderer=renderer):
...
+                view_classifiers.append(IViewClassifier)
+            if isexc:
+                view_classifiers.append(IExceptionViewClassifier)
+
+            for view_classifier in view_classifiers:
+                register_view(view_classifier, request_iface, derived_view,
+                              order, phash)
+
+            self.registry._clear_view_lookup_cache()
+            renderer_type = getattr(renderer, 'type', None) # gard against None
```
bertjwregeer: s/gard/guard/

pyramid/config/views.py
```diff
@@ -908,8 +943,41 @@ def register(permission=permission, renderer=renderer):
             # see also MultiView.__discriminator__
             view_intr['derived_callable'] = derived_view
 
-            registered = self.registry.adapters.registered
+            # make a new view separately for normal and exception paths
+            view_classifiers = []
+            if exception is None:
+                view_classifiers.append(IViewClassifier)
+            if isexc:
```
bertjwregeer: This was confusing at first. My brain didn't immediately click that `isexc` could potentially be true when the user is registering a context that also happens to be an exception without using `exception=`. I'd recommend a comment of some sort to make it more clear what is going on so that in the future someone doesn't think `isexc` means it is an `exception` so why would `exception` be `None`.

pyramid/config/views.py
```diff
@@ -1545,11 +1588,16 @@ def notfound(request):
         being used, :class:`~pyramid.httpexceptions.HTTPMovedPermanently will
         be used` for the redirect response if a slash-appended route is found.
 
-        .. versionchanged:: 1.6
         .. versionadded:: 1.3
+
+        .. versionchanged:: 1.6
```
bertjwregeer: What exactly changed in 1.6? Wonder if we should add what changed underneath it, otherwise I am not sure that this versionchanged is all that helpful.

pyramid/viewderivers.py
```diff
@@ -286,29 +287,38 @@ def _secured_view(view, info):
     authn_policy = info.registry.queryUtility(IAuthenticationPolicy)
     authz_policy = info.registry.queryUtility(IAuthorizationPolicy)
 
+    # no-op on exception-only views without an explicit permission
```
bertjwregeer: 👍

pyramid/viewderivers.py
```diff
@@ -499,18 +519,23 @@ def csrf_view(view, info):
...
         def csrf_view(context, request):
-            if (
-                request.method not in safe_methods and
-                (
-                    # skip exception views unless value is explicitly defined
-                    getattr(request, 'exception', None) is None or
-                    explicit_val is not None
-                )
-            ):
+            if request.method not in safe_methods:
```
bertjwregeer: Isn't this also missing a check for exceptions? or do you want to enable CSRF checks on exceptions always?

Because down below it is disabled unless it is explicitly set...

pyramid/config/views.py
mmerickel: I will rename `exception is None` to `isexc_only` so the test is more clear.

pyramid/config/views.py
mmerickel: I think that would be a good idea. I considered making `pyramid.exceptions.isexception` a public API as well. What do you think?

bertjwregeer: I think making that public API would probably be a good idea so that people could use that in various other parts of their code as needed.

pyramid/viewderivers.py
mmerickel: Excellent catch, I messed up the logic here. Thank you.

mmerickel:
I don't know what is up with appveyor. It fails immediately after the push, but later it runs the tests and passes successfully.

https://ci.appveyor.com/project/mmerickel/pyramid/build/exception_only.447

mmerickel:
My latest push should address most issues. It fundamentally breaks exception views and normal views apart at the view derivation step, such that `info.exception_only` can be used unequivocally to determine what type of view is being wrapped. I also reverted my original rename of `exception_only` to `exception`. We are back to `context=Exception, exception_only=True`. The reason for this is that basically everything in the view pipeline references `context` and this keeps things consistent. For example, the first parameter passed to view predicates, and to the view callable itself.

Please review this and let me know what you think!

pyramid/config/views.py
mmerickel: I pushed changes to master, 1.6-branch and 1.7-branch to fill in this version changed information. It is not yet merged into this branch.

bertjwregeer:
> I don't know what is up with appveyor. It fails immediately after the push, but later it runs the tests and passes successfully.

There are two AppVeyor tests that are run. The first attempts to merge this PR into master (which instantly fails because there are conflicts) the other just does a test to make sure that this branch successfully builds/tests.

If you were to merge master into this branch, the first appveyor test would work because now this PR could cleanly be merged into master.

bertjwregeer:
[Side note, I wish Github would remove my "approved review" LGTM automatically when new changes are pushed to a PR so that it doesn't look like I give a :+1: when I haven't had a chance to review the latest changes]

docs/narr/hooks.rst
```diff
@@ -1745,6 +1749,21 @@ View derivers are unique in that they have access to most of the options
...
 to do, and they have a chance to affect every view in the application.
 
+.. _exception_view_derivers:
+
+Exception Views and View Derivers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A :term:`view deriver` has the opportunity to wrap any view, including
+an :term:`exception view`. In general this is fine, but certain view derivers
+ay wish to avoid doing certain things when handling exceptions. For example,
```
bertjwregeer: s/ay/may/

docs/narr/viewconfig.rst
```diff
@@ -306,9 +306,26 @@ configured view.
...
+  It is possible to pass an exception class as the context if your context may
+  subclass an exception. In this case **two** views will be registered. One
+  will match normal incoming requests and the other will match as an
+  :term:`exception view` which only occurs when an exception is raised during
+  the normal request processing pipeline.
+
   If ``context`` is not supplied, the value ``None``, which matches any
   resource, is used.
 
+``exception``
```
bertjwregeer: Is this still supported? Or does this text go away?

pyramid/config/views.py
```diff
@@ -502,7 +502,21 @@ def wrapper(context, request):
           if the :term:`context` provides the represented interface;
           it is otherwise false.  This argument may also be provided
           to ``add_view`` as ``for_`` (an older, still-supported
-          spelling).
+          spelling). If the view should **only** match when handling
+          exceptions then use the ``exception`` argument instead of
```
bertjwregeer: there is no `exception` argument anymore.

pyramid/config/views.py
```diff
@@ -762,6 +776,12 @@ def view(context, request):
         if context is None:
             context = for_
 
+        isexc = isexception(context)
+        if exception_only and not isexc:
+            raise ConfigurationError(
+                'view "context" must be an exception type when '
+                '"exception_only" is True')
```
bertjwregeer: Is there a technical reason that the context for an exception view has to be an exception? Just wondering, not that I see any reason why it shouldn't always be an exception...

pyramid/config/views.py
```diff
@@ -1427,27 +1459,30 @@ def forbidden(request):
...
         .. versionadded:: 1.3
+
+        .. versionchanged:: 1.8
+
+           The view is created using ``exception_only=True``.
         """
         for arg in (
-            'name', 'permission', 'context', 'for_', 'http_cache',
-            'require_csrf',
+            'name', 'permission', 'context', 'for_', 'require_csrf',
```
bertjwregeer: Why the removal of `http_cache`?

pyramid/config/views.py
bertjwregeer: 👍

pyramid/tests/test_config/test_views.py
```diff
@@ -1815,6 +1858,124 @@ def configure_view():
...
+                          permission='foo')
+
+    def test_add_exception_view_disallows_for_(self):
+        config = self._makeOne(autocommit=True)
+        self.assertRaises(ConfigurationError,
+                          config.add_exception_view,
+                          context=Exception(),
+                          for_='foo')
+
+    def test_add_exception_view_disallows_http_cache(self):
```
bertjwregeer: This test doesn't seem right. `http_cache` is allowed by an exception view:

https://github.com/Pylons/pyramid/pull/2660/files#diff-8cd5bf5d51c2cecbdf85543126a03927R1663

I wonder if this is raising an `ConfigurationError` because `http_cache` is expecting a number or some other reason, but using `http_cache` should not raise an error at the moment just because it is an exception view.

pyramid/viewderivers.py
```diff
@@ -286,29 +286,27 @@ def _secured_view(view, info):
     authn_policy = info.registry.queryUtility(IAuthenticationPolicy)
     authz_policy = info.registry.queryUtility(IAuthorizationPolicy)
 
+    # no-op on exception-only views without an explicit permission
+    if explicit_val is None and info.exception_only:
```
bertjwregeer: Registering an exception view with `exception_view_config` we can't set permission at all... if `info.exception_only` is set we should just return the view always... unless I am missing a way in that `explicit_val` could still be true.

docs/narr/viewconfig.rst
bertjwregeer: This text should go away.

pyramid/config/views.py
mmerickel: I couldn't think of a reason why it was prevented versus any other view deriver option... There's nothing that enables http caching automatically and so the option is only there if people really want to opt-in to use it... and if they want to, I don't see a reason to stop them.

pyramid/config/views.py
mmerickel: There is nothing that actually requires it to be an exception but it will only be matched for a caught exception so preventing it is a good sanity check.

pyramid/viewderivers.py
mmerickel: It's possible to hit this by using `view_config` directly.

mmerickel:
Thanks for the quick review! I pushed several changes to address concerns.

Is there any way to get appveyor to stop trying to merge master? I don't see it as a goal of the CI testing.

bertjwregeer:
There is the last remaining typo for `s/ay/may/` in `hooks.rst` that I don't see fixed.

Also, maybe I missed it, but was there an answer for:

- What happens if I register an exception view and then register a normal view with the same context? Do they conflict? I assume they do... is there a test for this that I missed?

----

> Is there any way to get appveyor to stop trying to merge master? I don't see it as a goal of the CI testing.

AFAIK, no. I also thought Travis used to attempt to merge too, but I don't see that behavior on this PR.

https://docs.travis-ci.com/user/pull-requests#My-Pull-Request-isn’t-being-built

Says that it should be attempting to build the "merged" product, but only if Github actually has it available. With a merge conflict I assume it just isn't finding that hidden commit with the merge, and thus won't even try building it, unlike AppVeyor which tries to do it locally. 

See this issue: https://github.com/appveyor/ci/issues/882 for possible solutions?

mmerickel:
AFAICT the appveyor solution doesn't directly address this issue. It would reverse things such that the PR was built but not the feature branch which is fine, but it'd still be trying to merge master.

mmerickel:
The discriminator is the same between exception-only and normal views registered for the same context, thus they will conflict. If you override them between commits then it depends on the order but it will do what you asked it to do meaning: If you add the dual-view first, then override with an exception view then you will still have a normal view registered from the first step (we could fix this, but that's how it is now). If you add the excview first then add the dual-view then you will have only the dual-views registered.

mmerickel:
Now that we talked about it I think I might want to fix that such that if you register a dual-view and then override with an exception-only view that it will unregister the normal view. This would line up better with the fact that they use the same discriminator and conflict.

mmerickel:
I decided not to change the existing behavior as it's pretty hairy and requires changing the multiview a lot so that it can remove views instead of just replacing them. Maybe we can address this in a separate PR. I have rebased my changes into a single commit and merged master so everything should be good to go.

bertjwregeer:
And travis did the test for the merge now that it is merge-able again ;-)

mmerickel:
travis is smart

PR #2672, opened 2016-07-04
Show testsetup code in rendered docs.

### LINKED ISSUE #2670: reify docstring doesn't render properly
The testsetup section is omitted from the rendered docs at http://pyramid.readthedocs.io/en/latest/api/decorator.html#pyramid.decorator.reify

Unfortunately this section is essential for understanding the example.

stevepiercy:
For `testsetup`, Sphinx does not show code in the output: http://www.sphinx-doc.org/en/stable/ext/doctest.html#directive-testsetup This commit fixes that issue, although duplicates code. I'll submit a feature request to https://github.com/sphinx-doc/sphinx/ Fixes #2670
(cherry picked from commit 2612de2)

### DIFF
--- a/pyramid/decorator.py
+++ b/pyramid/decorator.py
@@ -8,6 +8,16 @@ class reify(object):
     replacing the function it decorates with an instance variable.  It is, in
     Python parlance, a non-data descriptor.  An example:
 
+    .. code-block:: python
+
+        from pyramid.decorator import reify
+
+        class Foo(object):
+            @reify
+            def jammy(self):
+                print('jammy called')
+                return 1
+
     .. testsetup::
 
         from pyramid.decorator import reify


tseaver:
The whole point of `testsetup` is to hide *irrelevant* details from the reader.  If it is your judgement that the reader needs to see the setup, It would be better to inline the `class Foo(object)` bits into the doctest example, and delete the `testsetup` block altogether.

stevepiercy:
@tseaver I tried that, too, but couldn't get it to work inside docstrings, so I gave up. But thanks to your remark, sensei, I gave it another shot.

I found that a simple copy-pasta was not enough, and that I needed to precede every line from the `testsetup` stanza with the Python prompt or ellipses, and proper indentation. See PR #2676

PR #2681, opened 2016-07-05
Remove cachebust_match

bertjwregeer:
It was leftover from when this feature was removed in the 1.6 release cycle


mmerickel:
I'm not sure yet whether we should remove this or document it and expose it via `add_static_view`. It is a possible escape hatch for someone to supply their own resolver to deal with cachebusted urls.

bertjwregeer:
I'm a -0 on that. I don't think it's a good idea.

PR #2684, opened 2016-07-06
updated old link to Alembic naming convention

enkidulan:
All docs now live on alembic.zzzcomputing.com, but for some reason link http://alembic.readthedocs.org/en/latest/naming.html redirects to wrong domain.


stevepiercy:
Thanks for catching this. There are about a dozen instances of links to Alembic throughout the docs, too, but I'll handle those, replacing them with intersphinx links instead so we only need to update one instance going forward.

PR #2687, opened 2016-07-06
This shows the correct view when a class and `attr` is involved.

sontek:

PR #2693, opened 2016-07-11
Fixed bug in `pcreate` such that it now shows the correct view when a class and `attr` is involved

stevepiercy:
Backport from master
See: https://github.com/Pylons/pyramid/pull/2687

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,11 @@
+unreleased
+==========
+
+- Fixed bug in `pcreate` such that it now shows the correct view when a class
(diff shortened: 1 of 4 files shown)


CHANGES.txt
```diff
@@ -1,3 +1,11 @@
+unreleased
+==========
+
+- Fixed bug in `pcreate` such that it now shows the correct view when a class
```
bertjwregeer: This should be `proutes` not `pcreate`

bertjwregeer:
This should be proutes not pcreate.

Issue #2590, opened 2016-05-20
wiki2 tutorial - Failed building wheel for bcrypt


viniciusban:
Unable to install ``bcrypt`` in Ubuntu 12.04 with Python 3.5.1.

Maybe it would be useful to write some note about this possible issue in Ubuntu 12.04.

Steps to reproduce the problem
-----------------------------------------

- Follow wiki2's steps until installing ``bcrypt``.
- pip will show long error message.

Solution
-----------

Set dependency to ``py_bcrypt`` instead of ``bcrypt`` in ``setup.py``.

mmerickel:
I hesitated to add bcrypt for reasons like this. I would sooner change the library than recommend a different library per platform.

How are you installing python 3.5.1 on ubuntu 12.04?

viniciusban:
From the deadsnakes ppa:

    $ sudo add-apt-repository ppa:fkrull/deadsnakes
    $ sudo apt-get update
    $ sudo apt-get install python3.5

Then, to create my virtualenv:

    $ python3.5 -m venv $VENV --without-pip
    $ source $VENV/bin/activate
    (venv) $ curl https://bootstrap.pypa.io/get-pip.py | python


BTW, I don't know if this should affect many people.

mmerickel:
You should ask @dstufft to release a manylinux1 wheel on PyPI for bcrypt. However it may still not work without having libffi installed on the system. I was able to get this to install via the following:

```
$ docker run --rm --it ubuntu:12.04 /bin/bash
$ add-apt-repository ppa:fkrull/deadsnakes
$ apt-get update
$ apt-get install libffi-dev python3.5-dev build-essential
$ python3.5 -m venv --without-pip env
$ curl https://bootstrap.pypa.io/get-pip.py | env/bin/python
$ env/bin/pip install bcrypt
```

bertjwregeer:
16.04 was released last month...

mmerickel:
I tried on 16.04 btw and it's a pretty similar story. You still need to install libffi-dev and python3.5-dev in order to build bcrypt.

dstufft:
IIRC ``py-bcrypt`` has some security bug now that renders it a bad idea, I could be misremembering though.

dstufft:
Also: I believe that ``py-bcrypt`` requires ``python3-dev`` too, so it's really just the ``libffi-dev`` dependency via CFFI.

mmerickel:
[...]

Issue #2204, opened 2015-12-30
Update official tutorials to use one-way encryption for authentication steps


stevepiercy:
The updated tutorial [Pyramid Blogr](http://pyramid-blogr.readthedocs.org/en/latest/registration.html) uses `passlib` and has a [warning about best practices](http://pyramid-blogr.readthedocs.org/en/latest/basic_models.html?highlight=encryption#update-database-initialization-script). The following tutorials should be modified to implement this best practice.
- [x] http://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/wiki2/authorization.html
- [x] http://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/wiki/authorization.html
- [x] http://docs.pylonsproject.org/projects/pyramid/en/master/quick_tutorial/authentication.html

RaHus:
The authorization tutorial and the authentication quick tutorials are using hard-coded values for a user database, i guess because the intent is to showcase pyramid features and not webdev best practices, so password encryption is not exactly applicable IMO.

Is the intent of this ticket to switch the tutorials to use permanent storage? and if so, which would be a good choice given the requirements of the tutorials?

stevepiercy:
@RaHus all these tutorials use persistent storage at some point. Currently these tutorials lack sufficient notice about encryption and omit examples of best practices, which gives a very misleading message about security: "unencrypted passwords are OK". We should do a lot better.

I think what we will end up doing is creating a single page, either in the main documentation or in the Pyramid Community Cookbook, that gives example for security best practices.

RaHus:
@stevepiercy I agree, the tutorial code should stick to best practices, i am just saying that the way the auth tutorials are written you'll have to make extra changes for them to use persistent storage in addition to the encrypted password changes.
in http://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/index.html 20 and 21 are based on code from 09, also 18 and 19 are based on 09. One solution would be to base 20 and 21 on 19.

stevepiercy:
I'd love to review something concrete: PR, gist, whatever. Contributing to docs has some challenges, and I'm happy to help out.

The grant-worthy label is to identify those issues or PRs that are considered worthy to be part of a grant application. We're looking into possibilities of funding sources to help make certain items a higher priority.

mmerickel:
[...]

Issue #2623, opened 2016-06-03
Change password_hash to LargeBinary from Text


stevepiercy:
- [ ] Change the column `password_hash` to type `LargeBinary` from `Text`.
- [ ] Remove `.encode('utf8')` so that it no longer encodes the value.
- Update across the following files:
    - [ ] docs/tutorials/wiki2/src/models/tutorial/models/user.py
    - [ ] docs/tutorials/wiki2/src/views/tutorial/models/user.py
    - [ ] docs/tutorials/wiki2/src/authentication/tutorial/models/user.py
    - [ ] docs/tutorials/wiki2/src/authorization/tutorial/models/user.py
    - [ ] docs/tutorials/wiki2/src/tests/tutorial/models/user.py
- Update tutorial docs to align with resulting changes.
    - [ ] docs/tutorials/wiki2/definingmodels
    - [ ] docs/tutorials/wiki2/definingviews
    - [ ] docs/tutorials/wiki2/authentication
    - [ ] docs/tutorials/wiki2/authorization
    - [ ] docs/tutorials/wiki2/tests

See #2576 for explanation.

Closes #2605.

stevepiercy:
@mmerickel is working on fix for `set_password` that will resolve this.

PR #2699, opened 2016-07-16
Add one-way password hash to security example in Quick Tutorial.

keitheis:
This pull-request is aiding #2204.
The usage of hashlib for one-way hash is from the discussion of #2590.

### DIFF
--- a/docs/quick_tutorial/authentication.rst
+++ b/docs/quick_tutorial/authentication.rst
@@ -50,7 +50,9 @@ Steps
     :linenos:
 
 #. Create an ``authentication/tutorial/security.py`` module that can find our
-   user information by providing an *authentication policy callback*:
+   user information by providing an *authentication policy callback*.
+   One-way hash function ``hash_password_with_salt`` is used to avoid user's
+   password stored in plain text:
 
    .. literalinclude:: authentication/tutorial/security.py
     :linenos:
--- a/docs/quick_tutorial/authentication/tutorial/security.py
+++ b/docs/quick_tutorial/authentication/tutorial/security.py
@@ -1,5 +1,21 @@
-USERS = {'editor': 'editor',
-         'viewer': 'viewer'}
+import os
+import hashlib
+import binascii
+
+SALT_LENGTH = 16
+
+def hash_password_with_salt(password, salt=None):
+    salt = salt or os.urandom(SALT_LENGTH)
+    dk = hashlib.pbkdf2_hmac('sha512', password.encode('utf-8'), salt, 100000)
+    return binascii.hexlify(dk + salt).decode('utf-8')
+
+def check_password(password, salted_password):
+    salt = binascii.unhexlify(salted_password[-2 * SALT_LENGTH:])
+    return hash_password_with_salt(password, salt) == salted_password
+
+
+USERS = {'editor': hash_password_with_salt('editor'),
+         'viewer': hash_password_with_salt('viewer')}
 GROUPS = {'editor': ['group:editors']}
 
 
--- a/docs/quick_tutorial/authentication/tutorial/views.py
+++ b/docs/quick_tutorial/authentication/tutorial/views.py
@@ -9,7 +9,10 @@ from pyramid.view import (
     view_defaults
     )
 
-from .security import USERS
+from .security import (
+    USERS,
+    check_password
+)
 
 
 @view_defaults(renderer='home.pt')
@@ -40,7 +43,7 @@ class TutorialViews:
         if 'form.submitted' in request.params:
             login = request.params['login']
             password = request.params['password']
-            if USERS.get(login) == password:
+            if check_password(password, USERS.get(login)):
                 headers = remember(request, login)
                 return HTTPFound(location=came_from,
                                  headers=headers)
--- a/docs/quick_tutorial/authorization/tutorial/security.py
+++ b/docs/quick_tutorial/authorization/tutorial/security.py
@@ -1,3 +1,19 @@
+import os
+import hashlib
+import binascii
+
+SALT_LENGTH = 16
+
+def hash_password_with_salt(password, salt=None):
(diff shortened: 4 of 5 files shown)


keitheis:
The travis-ci docs build failed is because of that http://docs.zope.org/zope.component/objects.inv' is unreachable.

docs/quick_tutorial/authentication.rst
```diff
@@ -50,7 +50,9 @@ Steps
     :linenos:
 
 #. Create an ``authentication/tutorial/security.py`` module that can find our
-   user information by providing an *authentication policy callback*:
+   user information by providing an *authentication policy callback*.
```
stevepiercy: Undo changes to lines 53-55, and move them into the "Analysis" section where the code is explained in detail. At this point, the user should focus on code replication, then later analyze what they have written.

In the Analysis or See Also sections (as appropriate), include a link to the relevant Python docs for each call. For example, in reST:

```rest
:py:func:`hashlib.pbkdf2_hmac`
```

...will generate HTML that links to https://docs.python.org/3/library/hashlib.html#hashlib.pbkdf2_hmac

docs/quick_tutorial/authentication.rst
```diff
@@ -50,7 +50,9 @@ Steps
...
+   One-way hash function ``hash_password_with_salt`` is used to avoid user's
+   password stored in plain text:
```
stevepiercy: It is preferable to explain what something does, instead of what it avoids. Suggest:

"The function ``hash_password_with_salt`` hashes the concatenation of the user's password and a randomly generated salt."

docs/quick_tutorial/authentication/tutorial/security.py
```diff
@@ -1,5 +1,21 @@
...
+import os
+import hashlib
+import binascii
+
+SALT_LENGTH = 16
+
+def hash_password_with_salt(password, salt=None):
+    salt = salt or os.urandom(SALT_LENGTH)
+    dk = hashlib.pbkdf2_hmac('sha512', password.encode('utf-8'), salt, 100000)
+    return binascii.hexlify(dk + salt).decode('utf-8')
```
stevepiercy: @bertjwregeer can you look this over and provide feedback?

This does not look right to me, because the function `pbkdf2_hmac` in the preceding line already hashes the concatenation of the password and salt behind the scenes.

Also 100000 iterations was OK in 2013 for SHA-256, but we're now in 2016. I don't know whether it needs to be adjusted for time as well as for SHA-512. I'd suggest using a variable of `iterations` to be consistent with the function's argument name.

Also we want to avoid the en/decode utf-8 stuff, as referred in issue #2623.

This goes for authorization as well.

stevepiercy:
Looking good so far. See line comments, and await code review from @bertjwregeer.

Regarding the build error, either we'll have to wait for docs.zope.org to come back online then try another build, or one of the maintainers can comment out that line in `conf.py` and run tox locally on the final PR.

docs/quick_tutorial/authentication/tutorial/security.py
ztane: No, this is not adjustable. hashlib.pbkdf2_hmac returns the bare hash, with neither salt nor iterations stored into the resulting hash. You cannot adjust the parameters with this or otherwise you cannot match the passwords later on!

I believe *if* we're wanting to show the proper way, then this example needs to install the passlib module and use the primitives from there instead.

keitheis:
@stevepiercy Thanks for the feedback. I'll wait for docs.zope.org to come back to make next update.

tseaver:
@keitheis, @stevepiercy  we should probably switch the intersphinx dependency to point to http://zopecomponent.readthedocs.io/en/stable/objects.inv

ztane:
As I noted in the comments, the proper thing to do in an application is to use a format such as the modular crypt:

     https://pythonhosted.org/passlib/lib/passlib.hash.pbkdf2_digest.html#passlib.hash.pbkdf2_sha256

which will make clearer which part is salt and how many rounds were used for the hash; it also makes it possible to *upgrade* the difficulty by increasing the number of rounds later on; that was what @stevepiercy was referring to by "This does not look right to me, because the function pbkdf2_hmac in the preceding line already hashes the concatenation of the password and salt behind the scenes."; but this is not true, this is just the bare primitive which makes up the `98jZicV16ODfEsEZeYPGHU3kbrUrvUEXOPimVSQDD44` part in `$pbkdf2-sha256$6400$.6UI/S.nXIk8jcbdHx3Fhg$98jZicV16ODfEsEZeYPGHU3kbrUrvUEXOPimVSQDD44`; the modular crypt format used by passlib prepends the *algorithm*, *number of rounds* **and** *the salt*, to the value, separated by `$`.

stevepiercy:
Thanks @tseaver.

I pushed a commit to my branch to make a PR, but GitHub has not updated yet, so I'm stuck. It's probably the same issue as [the one on July 12](https://status.github.com/messages). I filled out their [contact form](https://github.com/contact) and sent email to support@github.com, so feel free to ping them to make more noise.

@keitheis in addition to pushing a commit to your branch, you can restart the build at anytime by following the Details link for the failed build.

stevepiercy:
Updates are showing up again. @keitheis please pull in the commit from #2702.

docs/quick_tutorial/authentication/tutorial/security.py
stevepiercy: @ztane my intent was to show that the values could be changed according to the current recommendation by crypto folks, and suggest not to accept our tutorial as gospel, if that makes sense. IOW, set it and forget it, but set it for what is recommended at that point in time. That might be too much to expect from tutorial users.

stevepiercy:
Closing this because we hashed out (ha!) a fix for wiki2 in #2704. Please base your work off the example using bcrypt. Thank you!

PR #2704, opened 2016-07-16
fix the wiki2 tutorial to set the password as unicode

mmerickel:
Something really weird is happening but this fixes it. SQLAlchemy is
returning the "password_hash" from queries as the type that it was
inserted as. Not consistently unicode or bytes. If I insert bytes, then
I get bytes back out. If I insert unicode then I get unicode back out.
It's unclear why, as the type is Text, the data we're storing is
unambiguously US-ASCII and the connection is using a consistent
text_factory for unicode conversions of "str" on Python 3.

Here, we ensure that we always insert the value as unicode which appears
to fix downstream issues like those mentioned in #2605. I was able to
reproduce that bug and confirm this fixes it if the original database is
initialized using this fix.

Obsoletes #2623.


bertjwregeer:
:+1:

PR #2706, opened 2016-07-17
Fix renderers: JSON is not text and has no charset

### LINKED ISSUE #2691: JSON is not text and has no charset (upcoming WebOb changes)
Due to changes in WebOb (https://github.com/Pylons/webob/pull/261) to fix content type/charset handling, various tests within Pyramid are failing because the JSON renderer returns `str` objects and then the `RenderHelper` attempts to set `Response.text` which fails if there is no `charset`.

All none-text types (so anything that does not match mime type `text/*`, `application/xml` or `*/*+xml`) no longer automatically have a charset set, this means they are treated as bytes objects only.

https://github.com/Pylons/pyramid/blob/master/pyramid/renderers.py#L275
https://github.com/Pylons/pyramid/blob/master/pyramid/renderers.py#L382

Should both have a `.encode('UTF-8')` added, to correctly encode it's rendering into a bytes object.

Extra protection could also be added here:

https://github.com/Pylons/pyramid/blob/master/pyramid/renderers.py#L468-476

It should verify that a `charset` is set, before attempting to set the `response.text`.

With the addition of the call to `encode()` in the JSON renderers, the test suite would have to be appropriately updated to use `b''` wherever there are string literals that are being compared, currently there are 12 tests that fail with the `encode()` fix in place.

bertjwregeer:
This closes #2691, and means there has to be no new `default_encoding` in WebOb and removes the backwards incompatible change that would bring with it.

WebOb is still backwards incompatible in that it removed the default charset for `application/json`, but setting it in the first place was wrong.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -19,10 +19,15 @@ Features
 
 Bug Fixes
 ---------
+
 - Fixed bug in `proutes` such that it now shows the correct view when a class
   and `attr` is involved.
   See: https://github.com/Pylons/pyramid/pull/2687
 
+- The JSON renderers now encode their result as UTF-8. The renderer helper
+  will now warn the user and encode the result as UTF-8 if a renderer returns a
+  text type and the response does not have a valid character set. See
+  https://github.com/Pylons/pyramid/pull/2706
 
 Deprecations
 ------------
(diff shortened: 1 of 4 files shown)


mmerickel:
LGTM. We should backport this to 1.6 and 1.7.

bertjwregeer:
I'll get back ports completed later today.

PR #2711, opened 2016-07-18
Change gendered language in examples

inklesspen:
Some examples in documentation use "dude" and "bro" -- for example, "Not
found, bro". While playful, this language can make some people
uncomfortable. I have changed the wording to something equally playful
that doesn't make assumptions about the reader's gender.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -278,3 +278,5 @@ Contributors
 - Cris Ewing, 2016/06/03
 
 - Jean-Christophe Bohin, 2016/06/13
+
+- Jon Davidson, 2016/07/18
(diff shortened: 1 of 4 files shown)


tseaver:
I'd rather just get rid of the cutesy, semantically-null messages, e.g. just the following, everywhere:

```python
    return HTTPNotFound()
```

stevepiercy:
@inklesspen thanks for bringing these instances forward and offering a change for the better. I prefer narrative docs free of annoyances, yet I think @tseaver offers the best option.

...unless of course someone has a really good mummy reference.

inklesspen:
@tseaver @stevepiercy Does the change I made look good? If so, I'll squash it down and apply the same change to the secondary PRs targeting the version branches.

mmerickel:
LGTM

PR #2714, opened 2016-07-19
re.split() requires non-empty match

bertjwregeer:
Change from a * to a +, so long as there is something to split on, it
will split.

Fixes this warning:

```
FutureWarning: split() requires a non-empty pattern match.
  return _compile(pattern, flags).split(string, maxsplit)
```

### DIFF
--- a/pyramid/scripts/proutes.py
+++ b/pyramid/scripts/proutes.py
@@ -296,7 +296,7 @@ class PRoutesCommand(object):
             items = config.items('proutes')
             for k, v in items:
                 if 'format' == k:
-                    cols = re.split(r'[,|\s|\n]*', v)
+                    cols = re.split(r'[,|\s\n]+', v)
                     self.column_format = [x.strip() for x in cols]
 
         except configparser.NoSectionError:


bertjwregeer:
I didn't verify that this doesn't modify or change the output, just stops the warning, and tests all pass (locally).

mmerickel:
I think this is fine.
It is related to the `format` key documented here: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/commandline.html?highlight=proutes#displaying-all-application-routes

bertjwregeer:
Then it is merge ready. Unless the tests were testing the wrong thing :P

mmerickel:
@sontek Can you please give this a quick once-over to be sure the change is fine?

ztane:
Yes, it is correct. The current split logic will find all matching positions, but ignore those matches that are zero characters in width, thus this would still match the current behaviour, as it only excludes zero-width matches. Python 3.6 would then instead split between each character.

ztane:
however, now that I look into it, another question is *why* does it have `|` in a character class?

ztane:
Thus `[\n\s,]+`

bertjwregeer:
I assume someone was attempting to use the pipe as an "or". Either way leaving the pipe shouldn't make a difference, removing one pipe may be useful.

mmerickel:
@bertjwregeer can you fix that and then we can merge

bertjwregeer:
Done.

PR #2739, opened 2016-08-15
update output of pcreate --list for Quick Tour

stevepiercy:


mmerickel:
Nice find @stevepiercy.

PR #2746, opened 2016-08-17
Fix #2744 by reverting #2706 on 1.6-branch.

### LINKED ISSUE #2744: 1.7.1 changed the json renderer to return bytes instead of unicode
```
result = pyramid.renderers.render('json', {}, request=request)
```

``result`` is now ``bytes`` instead of ``unicode``. This is a regression in 1.6.2 and 1.7.1.

This is not a problem with `render_to_response` or `view_config(..., renderer=json)` because they are using the response object.

mmerickel:
This reverts commit 9db23a0a0b071ff68867c647d325b71c93e9893c.
This reverts commit 6cad66d3c8ab8a9bae04d0b08f5130f9462d7851.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,15 @@
+unreleased
+==========
+
+Bug Fixes
+---------
+
+- Revert changes from #2706 released in Pyramid 1.6.2. JSON renderers will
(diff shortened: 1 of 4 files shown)


tseaver:
LGTM

PR #2747, opened 2016-08-17
Fix #2744 by reverting #2706 on 1.7-branch.

### LINKED ISSUE #2744: 1.7.1 changed the json renderer to return bytes instead of unicode

mmerickel:
This reverts commit e00f8b0e3c17e86e21e6ae52509840c69c264d10.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,15 @@
+unreleased
+==========
+
+Bug Fixes
+---------
+
+- Revert changes from #2706 released in Pyramid 1.7.1. JSON renderers will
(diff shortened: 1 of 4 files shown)


tseaver:
LGTM

mmerickel:
thanks @tseaver !

PR #2750, opened 2016-08-17
Clarify that headers are ADDED to the response

mark0978:
Make it clear that headers are added to the response instead of overwriting existing headers.  Explain how to set the content_type instead of just assuming they will know.

### DIFF
--- a/pyramid/httpexceptions.py
+++ b/pyramid/httpexceptions.py
@@ -98,7 +98,9 @@ be forwarded to its :class:`~pyramid.response.Response` superclass:
      a plain-text override of the default ``detail``
 
    ``headers``
-     a list of (k,v) header pairs
+     a list of (k,v) header pairs or a dict to be added to the 
+     response, use content_type='application/json' kwarg to change 
+     the content type of the response.
 
    ``comment``
      a plain-text additional information which is


mmerickel:
LGTM.

I'd also ask whether you know also that the HTTPExceptions automatically support json responses as of 1.7, so you may not need to be doing whatever you're doing if the client claims it accepts json responses.

pyramid/httpexceptions.py
```diff
@@ -98,7 +98,9 @@
      a plain-text override of the default ``detail``
 
    ``headers``
-     a list of (k,v) header pairs
+     a list of (k,v) header pairs or a dict to be added to the 
+     response, use content_type='application/json' kwarg to change 
+     the content type of the response.
```
stevepiercy: This needs punctuation to clarify. Previously it seemed that only a list was acceptable. Now it seems that either a list or a dict is acceptable but that only the dict will be added to the response.

Additionally the passage is a fragment and a run-on, and should be broken into a statement and a sentence.

Suggest something like this:

```
A list of (k,v) header pairs or a dict. The headers will be added to the response.
Use the kwarg ``content_type='application/json'`` to change
the content type of the response.
```

bertjwregeer:
I believe that `Set-Cookie` is actually the ONLY header that may be duplicated in an HTTP Response.

Wonder if I should add a fix in WebOb so that when this merge happens it raises an error if you have duplicates other than `Set-Cookie`.

mmerickel:
@bertjwregeer Yeah I'm not sure raising an error is the right answer but perhaps the `extend` logic would collapse certain blacklisted headers that are known to not allow duplicates (or a conjugate approach with a whitelist). That sounds like a can of worms to me though.

mmerickel:
I tweaked this a bit and merged it. Thanks again @mark0978 .

Issue #2748, opened 2016-08-17
Add testing for Python 3.6


stevepiercy:
Another [alpha of Python 3.6 was released](https://www.python.org/downloads/release/python-360a4/), and the next pre-release of Python 3.6 will be 3.6.0b1, currently scheduled for 2016-09-12.

Should we start testing the 3.6 alpha or betas? If so, what do we need to do in tox?

We'll also need to update docs.

tseaver:
Travis has 3.6 support in its [`nightly` Python](https://docs.travis-ci.com/user/languages/python/#Nightly-build-support).

In `tox.ini`, just add `py36` as an environment, and add [`skip-missing-interpreters = True`](http://tox.readthedocs.io/en/latest/config.html#confval-skip_missing_interpreters=BOOL).

mmerickel:
I would also want `allow_failures` set on the nightly build.

https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

stevepiercy:
I gave this a few attempts in #2754, and I think I got it right on the third try. Would one of you verify?

Locally I installed `3.6-dev` into `~/.pyenv`, but at first I could not figure out how to get tox to pick it up:

```bash
InvocationError: Failed to get version_info for python3.6: b"pyenv: python3.6: command not found\n\nThe `python3.6' command exists in these Python versions:\n  3.6-dev\n\n"
```

This blog post helped:
https://anil.io/blog/python/pyenv/using-pyenv-to-install-multiple-python-versions-tox/#switching-versions-globally

This command:

```bash
pyenv global 3.5.2 2.7.12 3.4.5 3.3.6 3.6-dev 2.6.9
```

...allowed `tox -e py36` to run and pass all tests locally.

stevepiercy:
@mmerickel @tseaver @bertjwregeer does PR #2754 look good? I'll backport it to 1.7, too, if y'all approve.

mmerickel:
Closed by #2754. I don't see any need to backport this to 1.7.

PR #2754, opened 2016-08-19
add testing for py36

stevepiercy:
Hopefully I got the configuration correct in this attempt. See #2748

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
           env: TOXENV=docs
         - python: 3.5
           env: TOXENV=pep8
+    allow_failures:
+        - env: TOXENV=py36
 
 install:
   - travis_retry pip install tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
           env: TOXENV=docs
         - python: 3.5
           env: TOXENV=pep8
+        - python: nightly
+          env: TOXENV=py36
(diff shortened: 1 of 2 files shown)


bertjwregeer:
This looks good, but I am hesitant to add it while Py 3.6 is still in the dev cycle. I saw a lot of random failures on Py 3.5 (with WebOb development) when it was in its dev cycle that led to me chasing my tail in a way that was not meaningful or useful. Maybe when it is beta?

I am a big -1 on back-porting this to 1.7, until Py 3.6 has been released, and even then, I don't think anything will break, but I feel like it increases the maintenance burden for something already released.

stevepiercy:
@bertjwregeer on master, per @mmerickel's request, Travis will allow the build for just 3.6 to fail yet allow the entire build to pass, so it won't actually hold up anything. Example (although not due to 3.6 bugs, but due to my fumbling around with configuration, it still demonstrates the concept):
https://travis-ci.org/Pylons/pyramid/builds/153506166

We would be forewarned of issues with 3.6, and we can ignore any issues with it until it really matters. As a side benefit, ambitious developers on the bleeding edge can try it out in tox, and submit bug fixes. The first beta comes out on 2016-09-12.

I'll go ahead and merge this PR.

I'll leave backporting to 1.7 for later until Python 3.6 has an official release, or maybe by then we'll have Pyramid 1.9 and 2.0. 😜

PR #2760, opened 2016-09-01
allow prepare/bootstrap to be used as a context manager

### LINKED ISSUE #1822: make pyramid.paster.bootstrap into a context manager

mmerickel:
closes #1822

### DIFF
--- a/pyramid/tests/test_scripting.py
+++ b/pyramid/tests/test_scripting.py
@@ -134,6 +134,27 @@ class Test_prepare(unittest.TestCase):
         root, closer = info['root'], info['closer']
         closer()
 
+    def test_it_is_a_context_manager(self):
+        request = DummyRequest({})
+        registry = request.registry = self._makeRegistry()
+        closer_called = [False]
+        with self._callFUT(request=request) as info:
+            root, request = info['root'], info['request']
+            pushed = self.manager.get()
+            self.assertEqual(pushed['request'], request)
+            self.assertEqual(pushed['registry'], registry)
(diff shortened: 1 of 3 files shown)


pyramid/tests/test_scripting.py
```diff
@@ -134,6 +134,26 @@ def test_it_with_extensions(self):
...
+        registry = request.registry = self._makeRegistry()
+        closer_called = [False]
+        with self._callFUT(request=request) as info:
+            root, request = info['root'], info['request']
+            pushed = self.manager.get()
+            self.assertEqual(pushed['request'], request)
+            self.assertEqual(pushed['registry'], registry)
+            self.assertEqual(pushed['request'].registry, registry)
+            self.assertEqual(root.a, (request,))
+            info['closer']()
```
bertjwregeer: Possibly add a comment that you are calling closer here manually to clean up, before replacing it with a testable closer?

Also, why not just have your new closer call the original closer instead?

bertjwregeer:
LGTM.

pyramid/tests/test_scripting.py
mmerickel: Sure, I will fix this to be more clear.

PR #2762, opened 2016-09-06
convert response callbacks into a tween

### LINKED ISSUE #2622: move response callbacks to a tween
- [x] Move response callbacks to a tween `pyramid.tweens.response_callbacks`.
- [x] Adjust `config.add_tween` to, by default, always add tweens with `under='response callbacks', over=EXCVIEW`.
- [x] Document issues when adding tweens under EXCVIEW where cyclic dependencies are possible unless `over='MAIN'` is defined. This is similar to how `config.add_view_deriver` works with its defaults.
- [ ] Add changelog about bw-incompatibility where if a user is doing `add_tween(under=EXCVIEW)` then they should update their code to do `add_tween(under=EXCVIEW, over=MAIN)`.

This work will help the TM wrap the response callbacks in #2606.

mmerickel:
There is a backward incompatibility here in how tweens are defined. This
is consistent with the new view deriver API but may present some
frustration for anyone currently defining a tween **under** the EXCVIEW
as all tweens now default to **over** the EXCVIEW.

This incompatibility is not strictly required to implement this feature,
but if we don't then:

1) Tween ordering is almost completely dependent on the user's
   environment and sorting behavior. If they add another tween then the
   addon tween (if left unspecified) may be shifted to a bad place.
   This is an issue unrelated to this feature... just a lingering issue with
   tweens.

2) Another incompatibility may surface in which a user's tween would
   show up OVER the response callbacks tween, which would lead to
   inconsistent behavior versus the previous pipeline where callbacks
   were always executed after tweens.

This patch shifts tweens to be more fully specified which should be
generally considered a good thing.

Side note, the response callbacks and excview tweens are defined
with fallbacks incase one of them is overridden or unavailable for some
reason by using a list of dependencies. We should apply this pattern to
the view derivers as well to make them easier to override.

Closes #2622.

### DIFF
--- a/docs/api/tweens.rst
+++ b/docs/api/tweens.rst
@@ -7,6 +7,8 @@
 
    .. autofunction:: excview_tween_factory
 
+   .. autofunction:: response_callbacks_tween_factory
+
    .. attribute:: MAIN
 
       Constant representing the main Pyramid handling function, for use in
@@ -21,5 +23,9 @@
    .. attribute:: EXCVIEW
 
       Constant representing the exception view tween, for use in ``under``
-      and ``over`` arguments to
-      :meth:`pyramid.config.Configurator.add_tween`.
+      and ``over`` arguments to :meth:`pyramid.config.Configurator.add_tween`.
+
+   .. attribute:: RESPONSE_CALLBACKS
+
+      Constant representing the response callbacks tween for use in ``under``
+      and ``over`` arguments to :meth:`pyramid.config.Configurator.add_tween`.
--- a/docs/narr/hooks.rst
+++ b/docs/narr/hooks.rst
@@ -1249,6 +1249,7 @@ very last tween factory added) as its request handler function.  For example:
 The above example will generate an implicit tween chain that looks like this::
 
     INGRESS (implicit)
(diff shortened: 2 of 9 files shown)


bertjwregeer:
LGTM

mmerickel:
I'm working on docs for this now.

mmerickel:
There's another issue with this feature that I just noticed.

The default for ``request.invoke_subrequest`` is ``use_tweens=False``. Normally the response callbacks would have been executed here... this changes things so that callbacks are not executed. This incompatibility concerns me.

mmerickel:
Updated the docs but I'm worried about the response callbacks situation. It feels like a fairly large incompatibility. I argued when the feature was introduced to keep `use_tweens=True` and honestly `use_tweens` as an option seems like a bad idea in general (but it's there for things like the transaction manager that cannot be nested when using the threadlocal manager... which we're starting to move away from) but it's there now so we have to deal with it.

stevepiercy:
I'll review docs when the implementation is finalized. Please ping me at that time, and I'll dive in.

mmerickel:
I'm going to close this and port over only the changes from config.add_tween that explicitly add tweens OVER excview unless otherwise specified.

PR #2775, opened 2016-09-29
make --interactive the default for pcreate

### LINKED ISSUE #1225: scaffolds can be run more than once
```
env/bin/pcreate -s starter myproj
env/bin/pcreate -s alchemy myproj
cd myproj
../env/bin/python setup.py develop
# horrible errors because the dependencies required for the alchemy scaffold
# were not installed because the second pcreate did not overwrite the setup.py file
```

We need to either make the pcreate output more interactive and obvious that you are running a scaffold on top of another scaffold or prevent running pcreate on a directory that already exists.

mmerickel:
The current default is to just skip conflicting files. I think this is a better default, and if you want the old behavior you have to type it in at the prompt (`all n`). Instead of skipping I think most people either want interactive or ``--overwrite``.

fixes #1225

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -16,6 +16,12 @@ Backward Incompatibilities
 
    See https://github.com/Pylons/pyramid/pull/2615
 
+- ``pcreate`` is now interactive by default. You will be prompted if it
+  a file already exists with different content. Previously if there were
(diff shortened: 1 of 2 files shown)


bertjwregeer:
LGTM and the change makes sense.

PR #2778, opened 2016-09-30
add a callback hook to set_default_csrf_options for disabling checks …

### LINKED ISSUE #2596: Support hooks for csrf_view/default_csrf_options to allow custom business logic
I have a use case where by default i use `set_default_csrf_options` to protect all my views,
however I have a specific case when someone is not using default cookie based AuthTkt policy but instead uses command line/automation scripts with Auth Tokens to perform operations.

In this scenario it would be great if there would be an easy way to disable csrf checks on the views based on custom needs.

mmerickel:
…per-request

fixes #2596

### DIFF
--- a/pyramid/tests/test_config/test_security.py
+++ b/pyramid/tests/test_config/test_security.py
@@ -108,14 +108,18 @@ class ConfiguratorSecurityMethodsTests(unittest.TestCase):
         self.assertEqual(result.header, 'X-CSRF-Token')
         self.assertEqual(list(sorted(result.safe_methods)),
                          ['GET', 'HEAD', 'OPTIONS', 'TRACE'])
+        self.assertTrue(result.callback is None)
 
     def test_changing_set_default_csrf_options(self):
         from pyramid.interfaces import IDefaultCSRFOptions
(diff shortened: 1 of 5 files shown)


pyramid/tests/test_config/test_security.py
```diff
@@ -108,14 +108,18 @@ def test_set_default_csrf_options(self):
         self.assertEqual(result.header, 'X-CSRF-Token')
         self.assertEqual(list(sorted(result.safe_methods)),
                          ['GET', 'HEAD', 'OPTIONS', 'TRACE'])
+        self.assertTrue(result.callback is None)
 
     def test_changing_set_default_csrf_options(self):
         from pyramid.interfaces import IDefaultCSRFOptions
         config = self._makeOne(autocommit=True)
+        def callback(request): return True
```
tseaver: Why the single-line function here?

bertjwregeer: `callback = lambda x: True`

gives pep8 warnings, since you aren't supposed to assign a lambda to a variable...

bertjwregeer:
LGTM.

PR #2783, opened 2016-10-03
Add --package-name option to pcreate. 

goodwillcoding:
This solves the problem of scaffold creating an existing directory where the package created should not be named after the base folder. For example if I am in I run pcreate in ~/code/trypyramid.com and
would like to create package called tpc this is currently impossible. This solves the issues by allowing me to specify the package name on the command line.

Note: I'll add changelog changes post review

### DIFF
--- a/pyramid/scripts/pcreate.py
+++ b/pyramid/scripts/pcreate.py
@@ -49,8 +49,9 @@ class PCreateCommand(object):
                       dest='package_name',
                       action='store',
                       type='string',
-                      help='Package name to use. Has to be a valid python '
-                           'package name. (By default package name is derived '
+                      help='Package name to use. Named provided is assumed to '
+                           'be a valid python package name and will not be '
+                           'validated. (By default package name is derived '
                            'from output_directory base folder name)')
     parser.add_option('--simulate',
                       dest='simulate',
--- a/pyramid/scripts/pcreate.py
+++ b/pyramid/scripts/pcreate.py
@@ -45,6 +45,13 @@ class PCreateCommand(object):
                       action='store_true',
                       help=("A backwards compatibility alias for -l/--list.  "
                             "List all available scaffold names."))
+    parser.add_option('--package-name',
+                      dest='package_name',
+                      action='store',
+                      type='string',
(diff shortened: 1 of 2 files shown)


pyramid/scripts/pcreate.py
```diff
@@ -99,9 +106,13 @@ def output_path(self):
     def project_vars(self):
         output_dir = self.output_path
         project_name = os.path.basename(os.path.split(output_dir)[1])
-        pkg_name = _bad_chars_re.sub(
-            '', project_name.lower().replace('-', '_'))
-        safe_name = pkg_resources.safe_name(project_name)
+        if self.options.package_name is None:
+            pkg_name = _bad_chars_re.sub(
+                '', project_name.lower().replace('-', '_'))
```
mmerickel: Is this code not run on the project name if `--package-name` is not specified? It seems the `pkg_name` should be normalized in one spot.

goodwillcoding: That's existing code though, unmodified.

goodwillcoding: The only change is if package name is given it is assumed to be safe and hence assigned in the else block. The assumption of safety is documented in help for pcreate

mmerickel: Sorry I read that test wrong, I thought it was `is not None`. So are we not validating the package name if you specify `--package-name`?

goodwillcoding: We are not validating it. Rationale being: if user provide the package name then it's a case where they want a specific one and hence need to know if it is safe. I can add validation of course if you think it is needed.

mmerickel: Where is the assumption of safety documented? I don't get that from reading the help. Maybe rephrase that sentence slightly because I think `has to be a valid python package name` implies that it will be validated. Perhaps `Assumed to be a valid python package name`.

goodwillcoding: Fixed verbiage, and pushed.

bertjwregeer:
Thanks!

PR #2784, opened 2016-10-03
All of the tests for prequest assume that there is a Content-Type set

bertjwregeer:
But the application in this case doesn't actually send a Content-Type,
and thus there is no charset, and thus the tests should be assuming the
output is binary not text.

Add a Content-Type to be sent back from the app.

This is required to pass all tests on WebOb >1.7 where
Request.get_response(app) will no longer add the default_content_type if
a headerlist is passed in to Response.__init__, this allows
the Response to match what was provided by the app.


bertjwregeer:
This should get back-ported to the two latest releases as well so that when WebOb 1.7 comes out they don't start failing.

mmerickel:
That's fine, I can take care of that.

Issue #1125, opened 2013-09-21
Incorrect threadlocal behaviour when using gunicorn and eventlet


lukecyca:
When using [gunicorn](http://gunicorn.org) with `worker_class=eventlet`, it [monkey patches the threading library](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/geventlet.py#L43) (among other things). However, it does so after `pyramid.threadlocal` has been imported, so `pyramid.threadlocal.ThreadLocalManager` has already inherited the unpatched version of `threading.local`. This effectively means that pyramid's threadlocal functions are unreliable. For example, `get_current_request()` may return a request from a different thread!

To illustrate this, I made a [small test case](https://github.com/lukecyca/pyramid-eventlet-testcase).

As [recommended in the eventlet docs](http://eventlet.net/doc/patching.html#eventlet.patcher.monkey_patch), I've tried monkey patching as early as I can (in the module holding my app's `main()` function), but I still can't get it patched before `pyramid.threadlocal` imports it. I even instrumented `pyramid.threadlocal` to check whether `threading` has been patched, and here is the stack when it first gets imported:

```
Traceback (most recent call last):
  File "/Users/luke/.virtualenvs/eventlet-test/bin/pserve", line 8, in <module>
    load_entry_point('pyramid==1.4.2', 'console_scripts', 'pserve')()
  File "/Users/luke/.virtualenvs/eventlet-test/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 318, in load_entry_point
  File "/Users/luke/.virtualenvs/eventlet-test/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 2221, in load_entry_point
  File "/Users/luke/.virtualenvs/eventlet-test/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 1954, in load
  File "/Users/luke/.virtualenvs/eventlet-test/lib/python2.7/site-packages/pyramid-1.4.2-py2.7.egg/pyramid/scripts/pserve.py", line 30, in <module>
    from pyramid.paster import setup_logging
  File "/Users/luke/.virtualenvs/eventlet-test/lib/python2.7/site-packages/pyramid-1.4.2-py2.7.egg/pyramid/paster.py", line 10, in <module>
    from pyramid.scripting import prepare
[...]

PR #2797, opened 2016-10-14
fix pserve to work with gevent workers

mmerickel:
- refactored code to remove ``pyramid.threadlocal`` from pserve's import
  tree. It was coming from the ``pyramid.paster`` module which imports
  ``pyramid.scripting`` which imports ``pyramid.threadlocal``.

- dropped ``pyramid.scripts.common.logging_file_config`` as it appears
  to be unused and almost identical to ``setup_logging`` except
  without ``global_conf`` support.

It's super finicky but basically when using a `[server:main]` with `use = egg:gunicorn` with a worker of gevent, it's necessary to monkeypatch threading modules, etc. Before this patch, the import tree caused `pyramid.threadlocal` to be imported into other modules which triggers configuring the threadlocal manager using the current threading module **before** the server/app have an opportunity to patch the threading module.

Without this patch people who have been using greenlet-based workers from `[server:main]` have probably been broken in any code that uses `get_current_request` or `get_current_registry`.

related https://github.com/Pylons/pyramid/issues/1125 https://github.com/benoitc/gunicorn/issues/737

### DIFF
--- a/pyramid/paster.py
+++ b/pyramid/paster.py
@@ -5,9 +5,8 @@ from paste.deploy import (
     appconfig,
     )
 
-from pyramid.compat import configparser
-from logging.config import fileConfig
 from pyramid.scripting import prepare
+from pyramid.scripts.common import setup_logging  # noqa, api
 
 def get_app(config_uri, name=None, options=None, loadapp=loadapp):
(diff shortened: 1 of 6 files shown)


bertjwregeer:
LGTM.

PR #2801, opened 2016-10-30
Fix typo in docstring for check_csrf_token

dairiki:
This same typo appears to exist in all branches from 1.4 onward

### DIFF
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -215,7 +215,7 @@ def check_csrf_token(request,
     supplied by ``request.session.get_csrf_token()``, and ``raises`` is
     ``True``, this function will raise an
     :exc:`pyramid.exceptions.BadCSRFToken` exception.
-    If the check does succeed and ``raises`` is ``False``, this
+    If the check does not succeed and ``raises`` is ``False``, this
     function will return ``False``.  If the CSRF check is successful, this
     function will return ``True`` unconditionally.
 


pyramid/session.py
```diff
@@ -215,7 +215,7 @@ def check_csrf_token(request,
     supplied by ``request.session.get_csrf_token()``, and ``raises`` is
     ``True``, this function will raise an
     :exc:`pyramid.exceptions.BadCSRFToken` exception.
-    If the check does succeed and ``raises`` is ``False``, this
+    If the check does not succeed and ``raises`` is ``False``, this
```
stevepiercy: I think the term "does succeed" could be more clear. I had to read this statement several times and look at the code closely. In this context I think "does succeed" means `strings_differ(bytes_(expected_token), bytes_(supplied_token))` is `True`.

Anyway, I think that a better phrase would be the following:

```rst
If the values differ and ``raises`` is ``False``, this
function will return ``False``.
```

I'd like one of the other core developers to double-check this.

mmerickel: The behavior in this PR and your text are both correct. The current docs are wrong.

stevepiercy: @mmerickel which one: mine or @dairiki's?

I think using "succeed" is difficult to parse due to double negatives and ambiguity of the context. Does "succeed" mean that the function returns `True` or that the check for matching tokens is `True`?

mmerickel: I read both of your suggestions the same. I'd prefer your suggestion but they are both better than the current docs which are wrong.

mmerickel:
Ok this stayed open too long. Fixed it myself in b13b1c8de94a55ac99865c06598743899806c689 and backported to 1.6/1.7 branches.

Issue #2732, opened 2016-08-04
support extending pserve to monitor arbitrary files via a custom ini section


mmerickel:
Currently the only way to extend `pserve` to monitor extra files for restarts is via monkey patching the code at runtime.

Similar to `pshell`'s `[pshell]` section I propose we add a new `[pserve]` ini section with a `watch_files` key supporting new-line separated patterns.

```
[pserve]
watch_files =
    foo.ini 
    something else with spaces in the name.png
    a glob pattern/**/*.png
```

mmerickel:
Glob support requires updating hupper. See https://github.com/Pylons/hupper/issues/5.

Issue #2250, opened 2016-01-19
fix pserve --reload to avoid crashing the monitor


mmerickel:
In Pyramid 1.6 we deprecated the daemon functionality in pserve to simplify the codebase. As a result it should be easier to rewrite the reloading functionality into a master-slave such that the monitor keeps running and forks the app, restarting it when it crashes.
- [x] Remove daemon functionality.
- [x] Rewrite loader to use multiprocessing.
- [x] Update reloader to avoid crashing the monitor.
- [x] Remove syntax error checking in favor of crashing.
- [x] Get all of this working on windows (https://github.com/Pylons/hupper/issues/4).
- [x] Merge the `pserve-reloader-revamp` [1] branch (#2805).

Removing syntax error checking is debatable but I'm in favor of some more user-facing error reporting instead of the current message in the console without actually crashing (#2249).

[1] https://github.com/mmerickel/pyramid/tree/pserve-reloader-revamp

mmerickel:
So there is some precedent for this in wsgiwatcher [1] and reloadwsgi [2].

Carl Meyer was nice enough to basically give me full control of wsgiwatcher to do with what we'd like. If nothing else it shows an excellent approach to using multiprocessing alongside watchdog to support inotify-based reloading. I don't really have any issue with adding a dependency on watchdog if we wish as it appears to fallback gracefully on alternate platforms.

It feels a little heavy to add a new dependency to pyramid just for handling `pserve --reload` (as production uses of `pserve` do not require this library, but it'd also be better to just work than to require users to install a new dependency in development via `pip install pyramid[debug]` or something.

There are 2 approaches available to us that I think are both viable:

1. Copy wsgiwatcher into pserve and make pserve work!

2. Update wsgiwatcher to support some hooks that pserve can use to pass it a callable that loads the wsgi app and leave all the reloading logic inside wsgiwatcher. Then release a version of wsgiwatcher that pyramid depends on.

I'd like to hear some ideas here because I've gone back and forth about 10 times in the last week deciding whether the new dependency is worthwhile (I think pip makes extra deps mostly a non-issue so right now I'm leaning toward option 2).

[1] https://github.com/carljm/wsgiwatcher/blob/master/wsgiwatcher/watcher.py

[2] https://bitbucket.org/dholth/reloadwsgi/src/7a251ecd21b517a1a3bffd98a2e5ae090e7d6640/reloadwsgi.py?at=default&fileviewer=file-view-default

[...]

Issue #2081, opened 2015-11-05
pserve --reload should use inotify or equivalent when available


JonnoFTW:
After reading the pserve code it appears that using the `--reload` flag makes a loop that runs every second (with adjustable times) which runs stat against all relevant files. This is a waste of resources, since linux provides the inotify system.

The offending code is in this function:

https://github.com/Pylons/pyramid/blob/master/pyramid/scripts/pserve.py#L863

A potential fix might require a FS monitoring library like [watchdog](https://pythonhosted.org/watchdog/) which handles these invents in an OS independent manner.

stevepiercy:
Please consider the following points.

* The behavior is not broken.
* `--reload` is only useful during development. It's merely not efficient as you would like in development where inefficiency is not a high priority. If the current behavior impedes development work, then that would be a good argument.
* I don't think more dependencies is a good idea, and Pyramid has been shedding them in order to be efficient where it matters most.

I would be interested in seeing how you would implement your suggestion, preferably through a PR on a new branch.

I think this might be a candidate for the [Pyramid Cookbook](http://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/), perhaps with benchmarking comparisons.

bertjwregeer:
To add another point to Steve's list, Pyramid is heavily developed on not just Linux but also Mac OS X where `inotify` does not exist.

Windows is another target that we have to keep in mind, thus making changes that are tied to a single operating system makes things more difficult to debug.

That being said, `pserve` is a simple script that you are free to take, and modify as you see fit. It is a runner that ultimately just starts a WSGI server. If you would like you could provide `pyramid_linux_pserve` as a package that implemented your changes.

mmerickel:
Adding an optional dependency on watchdog isn't a bad idea. We could add it as an `extras` on pyramid install and pserve could use it if available. Watchdog itself is cross platform.

JonnoFTW:
@bertjwregeer I mentioned watchdog because it is cross platform. I'll make a package and see how it goes.

domenkozar:
It depends on `PyYAML >=3.10', 'argh >=0.24.1', 'pathtools >=0.1.1`

mmerickel:
related #2250

bertjwregeer:
This issue has been resolved with #2805 which uses `hupper` which can use `watchdog` to look for changes.

PR #2805, opened 2016-11-04
pserve reloader revamp

### LINKED ISSUE #2223: pserve --reload in 1.6 is less reliable than in 1.5
Somehow in 1.6 the `pserve --reload` is not as reliable as it used to be in pre-1.6.

Bug: 
Pserve can get into a mode, where it forgets about files, and no matter if the file has been saved and resaved, it just doesn't want to reload it. I cannot reproduce it yet, but it happened after I had a NameError in a try-except block. Pserve didn't give any warning in that case, it just 'forgot' about that particular file's updates. It's also possible that one of the savings of the file was during a reload process or something timing related.

I had to manually stop the process with Ctrl + C and start it again, to get pserve back into normal mode.

mmerickel:
This PR replaces the `--reload` support in `pserve` with the new `hupper` package [1].

**Features**
- The monitor is now a separate process that will not crash and starts up before any of your code.
- The monitor will not restart the process after a crash until a file is saved.
- The monitor works on windows (~~currently only on py35 but I'm working on it~~... the current pserve has lots of issues on windows, so it's ~~still~~ an improvement).
- You can now trigger a reload manually from a pyramid view or any other code via `hupper.get_reloader().trigger_reload()`. Kind of neat.
- You can trigger a reload by issuing a `SIGHUP` to the monitor process.
- If the `watchdog` package [2] is installed then monitoring will be done using inotify instead of cpu-intensive polling.
- Fixes #2223,  #2250 and #2081.

**Incompatibilities**
- Drops hidden APIs like `pyramid.scripts.pserve.watch_file`, and `pyramid.scripts.pserve.add_file_callback`. #2732 will be a path to a fully supported version of these features. `hupper` also has its own public API to add files to monitor via `hupper.get_reloader().watch_files(...)`.

~~I want to avoid merging this until windows support is improved in hupper but we're close enough to get some review.~~

[1] http://docs.pylonsproject.org/projects/hupper/en/latest/
[2] https://pypi.org/project/watchdog/

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -93,6 +93,29 @@ Features
   using the ``PYRAMID_CSRF_TRUSTED_ORIGINS`` environment variable similar to
   other settings. See https://github.com/Pylons/pyramid/pull/2823
 
+- ``pserve --reload`` now uses the
+  `hupper <http://docs.pylonsproject.org/projects/hupper/en/latest/>`
+  library to monitor file changes. This comes with many improvements:
+
+  - If the `watchdog <http://pythonhosted.org/watchdog/>`_ package is
+    installed then monitoring will be done using inotify instead of
+    cpu and disk-intensive polling.
+
+  - The monitor is now a separate process that will not crash and starts up
+    before any of your code.
+
+  - The monitor will not restart the process after a crash until a file is
+    saved.
+
+  - The monitor works on windows.
+
+  - You can now trigger a reload manually from a pyramid view or any other
+    code via ``hupper.get_reloader().trigger_reload()``. Kind of neat.
+
+  - You can trigger a reload by issuing a ``SIGHUP`` to the monitor process.
+
+  See https://github.com/Pylons/pyramid/pull/2805
+
 Bug Fixes
 ---------
 
--- a/docs/narr/project.rst
+++ b/docs/narr/project.rst
@@ -1045,3 +1045,23 @@ Another good production alternative is :term:`Green Unicorn` (aka
 mod_wsgi, although it depends, in its default configuration, on having a
 buffering HTTP proxy in front of it.  It does not, as of this writing, work on
 Windows.
+
+Automatically Reloading Your Code
+---------------------------------
+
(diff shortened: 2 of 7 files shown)


mmerickel:
This is ready for review.

stevepiercy:
Is there a need to update the Pyramid docs?

If so, is there a separate issue open?

mmerickel:
There's no change to the docs unless we want to add some mentions to the pserve docs like... To make this work better you should install watchdog into your development environment. See http://docs.pylonsproject.org/projects/hupper/en/latest/#watchdog-support for more info on what I mean.

stevepiercy:
I made a new issue #2806 for making mention.

mmerickel:
New commits should handle #2806.

bertjwregeer:
Will review this in a little bit.

mmerickel:
I just rebased this branch on top of the latest master, should be mergeable again.

bertjwregeer:
Would it make sense to rebase and squash the first three commits down to "use hupper"?

mmerickel:
Sure if you want me to. In general Pyramid hasn't been very picky about commit history.

bertjwregeer:
I tend to not want to have WIP commits in the final history... but if it doesn't matter, I don't care. I've been using this branch + hupper + watchdog for the last 3 hours for a project I am working on. Haven't seen it crash yet!!!

Also, it seems to be much faster at reloading than the old version. With watchdog it is fast enough that doing a `pip install -U webob==1.7.0rc1` will cause it to restart three times as pip installs the newly updated files ;-)

mmerickel:
Is it restarting more than once per second?

The behaviors are a little bit more complex than the old reloader because the file monitor is always running so some race conditions can definitely exist that I've tried to account for.

mmerickel:
Okay I made the history pretty.

bertjwregeer:
It might be more than once per second (I have watchdog installed) but it's just really fast... I haven't seen it do something weird or get wedged.

PR #2811, opened 2016-11-12
Mark more characters as safe in escaping generated path components

moriyoshi:
According to [RFC3986](https://tools.ietf.org/html/rfc3986) [1], the following characters are allowed to occur bare (that is, don't need percent encoding) in addition to alphanumeric characters:

`-` / `.` / `_` / `~` / `!` / `$` / `&` / `'` / `(` / `)` / `*` / `+` / `,` / `;` / `=` / `:` / `@`

But `route_url()` escapes all the occurrences of the following characters in the path components given as keywords.

`~` / `!` / `$` / `&` / `'` / `(` / `)` / `*` / `+` / `,` / `;` / `=` / `:` / `@`

```python
config.add_route('someroute', '/{x}')
...
assert request.route_path('someroute', x='@') == '/@' # '/%40' != '/@'
```

This patch tries to address the issue by marking the following characters as safe in escaping path components in `IRoute.generate()`.

[1] An excerpt from RFC3986:
```
      path          = path-abempty    ; begins with "/" or is empty
                    / path-absolute   ; begins with "/" but not "//"
                    / path-noscheme   ; begins with a non-colon segment
                    / path-rootless   ; begins with a segment
                    / path-empty      ; zero characters

      path-abempty  = *( "/" segment )
      path-absolute = "/" [ segment-nz *( "/" segment ) ]
      path-noscheme = segment-nz-nc *( "/" segment )
      path-rootless = segment-nz *( "/" segment )
      path-empty    = 0<pchar>

      segment       = *pchar
      segment-nz    = 1*pchar
      segment-nz-nc = 1*( unreserved / pct-encoded / sub-delims / "@" )
                    ; non-zero-length segment without any colon ":"

      unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

      pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -284,3 +284,5 @@ Contributors
 - Jon Davidson, 2016/07/18
 
 - Keith Yang, 2016/07/22
+
+- Moriyoshi Koizumi, 2016/11/20
--- a/pyramid/tests/test_traversal.py
+++ b/pyramid/tests/test_traversal.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import unittest
 import warnings
 
@@ -839,7 +840,7 @@ class QuotePathSegmentTests(unittest.TestCase):
     def test_string(self):
         s = '/ hello!'
         result = self._callFUT(s)
-        self.assertEqual(result, '%2F%20hello%21')
+        self.assertEqual(result, '%2F%20hello!')
 
     def test_int(self):
         s = 12345
@@ -1298,6 +1299,15 @@ class Test__join_path_tuple(unittest.TestCase):
     def test_nonempty_tuple(self):
         result = self._callFUT(('x',))
         self.assertEqual(result, 'x')
+
+    def test_segments_with_unsafes(self):
+        safe_segments = tuple(u"0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-._~!$&'()*+,;=:@")
+        result = self._callFUT(safe_segments)
+        self.assertEqual(result, u'/'.join(safe_segments))
+        unsafe_segments = tuple(chr(i) for i in range(0x20, 0x80) if not chr(i) in safe_segments) + (u'あ',)
+        result = self._callFUT(unsafe_segments)
+        self.assertEqual(result, u'/'.join(''.join('%%%02X' % (ord(c) if isinstance(c, str) else c) for c in unsafe_segment.encode('utf-8')) for unsafe_segment in unsafe_segments))
+
 
 def make_traverser(result):
     class DummyTraverser(object):
--- a/pyramid/tests/test_urldispatch.py
+++ b/pyramid/tests/test_urldispatch.py
@@ -485,11 +485,15 @@ class TestCompileRouteFunctional(unittest.TestCase):
     def test_generator_functional_newstyle(self):
         self.generates('/{x}', {'x':''}, '/')
         self.generates('/{x}', {'x':'a'}, '/a')
+        self.generates('/{x}', {'x':'a/b/c'}, '/a/b/c')
+        self.generates('/{x}', {'x':':@&+$,'}, '/:@&+$,')
         self.generates('zzz/{x}', {'x':'abc'}, '/zzz/abc')
         self.generates('zzz/{x}*traverse', {'x':'abc', 'traverse':''},
                        '/zzz/abc')
         self.generates('zzz/{x}*traverse', {'x':'abc', 'traverse':'/def/g'},
                        '/zzz/abc/def/g')
+        self.generates('zzz/{x}*traverse', {'x':':@&+$,', 'traverse':'/:@&+$,'},
+                       '/zzz/:@&+$,/:@&+$,')
         self.generates('/{x}', {'x':text_(b'/La Pe\xc3\xb1a', 'utf-8')},
                        '//La%20Pe%C3%B1a')
         self.generates('/{x}*y', {'x':text_(b'/La Pe\xc3\xb1a', 'utf-8'),
--- a/pyramid/traversal.py
+++ b/pyramid/traversal.py
@@ -35,6 +35,9 @@ with warnings.catch_warnings():
     warnings.filterwarnings('ignore')
     from pyramid.interfaces import IContextURL
 
+PATH_SEGMENT_SAFE = "~!$&'()*+,;=:@" # from webob
+PATH_SAFE = PATH_SEGMENT_SAFE + "/"
+
 empty = text_('')
 
 def find_root(resource):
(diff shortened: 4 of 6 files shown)


moriyoshi:
> 1) Fix webob's PATH_SAFE and ensure that everything is okay there.

This definitely needs to be taken account of, but I suspect there were consequences behind the reduced character set and I'd like to address it separately (by the issue for WebOb).

> 2) Fix the rest of request.route_url and request.resource_url to support extra characters (pyramid.url._join_elements).

Agreed.

> 4) Fix resource paths using pyramid.traversal._join_path_tuple similarly.
> 3) Try to use the same pyramid.url.PATH_SAFE in all pyramid locations.

Wondering if webob.request:PATH_SAFE is intended for public use, I guess it would be better to import the constant from the module instead of duplicating it in pyramid.url.  Any opinions?

mmerickel:
> but I suspect there were consequences behind the reduced character set and I'd like to address it separately (by the issue for WebOb).

Yes, that's why I would want to see it changed there as well. If the reasons for its current state can be found then we may not want to change things in Pyramid.

> Wondering if webob.request:PATH_SAFE is intended for public use, I guess it would be better to import the constant from the module instead of duplicating it in pyramid.url. Any opinions?

I think that this will not be possible due to their decoupled release cycles. Right now it's not public, and also doesn't do what you want. If you fix it there and make it public then Pyramid will need to depend on a new version of webob. That's not unreasonable, but can only really be considered if a fix does go into webob and not before then.

I'm definitely curious what @bertjwregeer will have to say about the special characters and why webob uses a reduced set.

bertjwregeer:
Just quit my job... I'll go spelunking over the next day or two to find an answer as to why, and provide my feedback on what is the best way to move forward.

bertjwregeer:
RFC3986 says that the following characters are reserved:

      reserved    = gen-delims / sub-delims
      gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="

Then the following are unreserved (these absolutely don't need to be percent encoded)

    unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

For the path segments the following is allowed:

    pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"

That gives us:

    ALPHA / DIGIT / "-" / "." / "_" / "~" / "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "=" / ":" /"@"

All of these are path segment safe.

Then there is the case of the query part of the URL, which has yet another different set of requirements:

    query       = *( pchar / "/" / "?" )

The following is valid:

    http://example.com/~mypath@something/myfile!?q=?/funstuff

And the `q` parameter will be set to `?/funstuff`. Although using key/value pairs like that is not required. People have also used `;` for doing versioning for example.

    http://example.com/~mypath@something/myfile!?q;v=1.1=?/funstuff


This could refer to using the q parameter as version 1.1. I've seen some URL's in my time man (insert acid flashbacks ;-))

Does Python correctly handle the case currently of the query parameters containing the slash and the question mark?

Either way, the fix here should be to just set PATH_SAFE to the above full string of characters. Let's not expand it only a little.

From what I can find in the WebOb history, it was added because there was an issue with double percent encoding with certain Zope testing components. 

https://bitbucket.org/ianb/webob/issues/5

I don't see any issues with the list being expanded to match all of the actual characters in RFC3986.

bertjwregeer:
Do note that there is a fix in Pyramid that is waiting on WebOb 1.7 as it stands, so if we get this into WebOb before then, then Pyramid can depend on it without issues.

ztane:
`;` used to have a separate meaning in the paths; one could pass parameters to path segments (http://stackoverflow.com/a/2163937/918959). I don't think this is used anywhere any more for HTTP URLs. And in query strings, `;` is separator, alternative to `&`, thus the `q;v=1.1` would be parsed by parse_qsl as

    >>> urllib.parse.parse_qsl('q;v=1.1&foo=bar')
    [('v', '1.1'), ('foo', 'bar')]

ztane:
Also:

```
>>> urllib.parse.urlsplit('http://example.com/~mypath@something/myfile!?q;v=1.1=?/funstuff')
SplitResult(scheme='http', netloc='example.com', path='/~mypath@something/myfile!', query='q;v=1.1=?/funstuff', fragment='')
>>> urllib.parse.urlunsplit(_)
'http://example.com/~mypath@something/myfile!?q;v=1.1=?/funstuff'
>>> urllib.parse.parse_qsl('q;v=1.1=?/funstuff')
[('v', '1.1=?/funstuff')]
```

ztane:
And the `;` should be considered as an alternative to `&` in query strings as per https://www.w3.org/TR/1999/REC-html401-19991224/appendix/notes.html#h-B.2.2

bertjwregeer:
@ztane: That's the older RFC2396 which was obsoleted by RFC3986.

Pythons url parsing is still stuck in the RFC2396 days, which is why the newer characters are not considered path safe.


>   Aside from dot-segments in hierarchical paths, a path segment is
>   considered opaque by the generic syntax.  URI producing applications
>   often use the reserved characters allowed in a segment to delimit
>   scheme-specific or dereference-handler-specific subcomponents.  For
>   example, the semicolon (";") and equals ("=") reserved characters are
>   often used to delimit parameters and parameter values applicable to
>   that segment.  The comma (",") reserved character is often used for
>   similar purposes.  For example, one URI producer might use a segment
>   such as "name;v=1.1" to indicate a reference to version 1.1 of
>   "name", whereas another might use a segment such as "name,1.1" to
>   indicate the same.  Parameter types may be defined by scheme-specific
>   semantics, but in most cases the syntax of a parameter is specific to
>   the implementation of the URI's dereferencing algorithm.

It's not in the generic URI specification, and encoding it is not strictly necessary.

I don't think encoding `;` is something that should be done based upon RFC3896 3.4: https://tools.ietf.org/html/rfc3986#section-3.4 otherwise we should also percent encode `=`.

That being said, regarding your HTML4 TR link... in HTML5 using the ampersand in a href is allowed in most cases, and using ; is not something that really has taken off.

moriyoshi:
Now that the patch contains the full whitelist, I've fixed the PR's description accordingly.

mmerickel:
This looks good to me. I will wait to see if @bertjwregeer has anything else to say prior to merging. Thanks @moriyoshi!

mmerickel:
@moriyoshi please add your name to CONTRIBUTORS.txt as well, thank you.

moriyoshi:
Done, thank you for all the guidance!

bertjwregeer:
I'm good with these changes :-). Thanks @moriyoshi!

PR #2816, opened 2016-11-14
Fix flake8 failures

tseaver:
Flake 3.1, released today, caused previously-overlooked E305 and E306 errors to trip.  This PR annotates the excluded codes, and adds those two explicitly.


mmerickel:
Great, thanks for adding those comments!

PR #2818, opened 2016-11-14
Strip allegedly-smart quotes in comments.

tseaver:
They break installing on a `LANG=C` system.


tseaver:
@mmerickel I just saw Jenkins fail after merge of #2816:  I had copied the comments from the `flake8` docs for the errors, and overlooked the fact that they used non-ASCII single quotes (instead of apostrophes).

PR #2822, opened 2016-11-19
remove the deprecated IContextURL

mmerickel:
This PR drops support for the IContextURL interface that was deprecated in 1.3. The only part that was pretty confusing for me was the reimplementation of `pyramid.traversal.virtual_root` which seems somewhat different than before. It no longer consults the registered `IResourceURL` object for the given resource and thus the root is not configurable. I really have no idea if this is okay.

It'd be really great if @mcdonc could review this considering he originally deprecated this interface and probably understands the purpose of the API much better than I.


mmerickel:
Updated this with an attempt at using `virtual_path` and `physical_path` from the custom `IResourceURL` for an object in order to find the virtual root. I think this probably works.

bertjwregeer:
I am not as familiar with this part of the code base. From just a discerning eye and looking over the logic, it looks good to me.

PR #2827, opened 2016-11-21
support a [pserve] config section with a list of files to watch

### LINKED ISSUE #2732: support extending pserve to monitor arbitrary files via a custom ini section

mmerickel:
fixes #2732


mmerickel:
I'm thinking that I will update this to support asset specs such that if there's a `:` in the name it will resolve it. For example,

```
[pserve]
watch_files =
    myapp:static/foo.png         # relative to myapp package
    extra_files/bar.png          # relative to ini file ??? cwd?
    %(here)s/extra_files/baz.png # relative to ini file
```

Currently everything is relative to the ini file and the `%(here)s` syntax is not supported. Does anyone have some thoughts on whether I should support the `%(here)s` syntax?

I think ideally `extra_files/bar.png` should be relative to the app but it's not really possible because this is loaded by pserve, not by the app, so we don't really know where the app's package is or what it is. The only sane thing seems to be relative to the ini file.

ztane:
Well, I wouldn't support cwd-relativeness at all, usually it is just PITA because it is difficult to tell what the cwd happens to be at which time. I'd support full variable interpolations though, not only here.

And how about globbing. I wouldn't want to change my development.ini all the time. Just pass the final "filename" into `glob.glob()` at *initialization time* (let's not consider new files yet), to get a list of resulting paths. (And too bad that recursive globbing doesn't exist < 3.5)

ztane:
hmm though... `glob.glob('foo.bar')` where foo.bar doesn't exist returns empty list; so if it should watch new named files that don't exist yet...

mmerickel:
Globbing is coming, I just need to think about it a bit more. It will be handled by hupper and so it won't require any changes to this PR aside from a doc improvement.

Originally I wanted to support a glob that would actually watch for new files that matched... but it's somewhat sucky to implement so I might just re-evaluate the glob after each restart.

ztane:
well, if it is possible to register a new matcher, then someone could write a version that uses [inotify](https://pypi.python.org/pypi/inotify) say

mmerickel:
> well, if it is possible to register a new matcher, then someone could write a version that uses inotify say

me thinks you haven't read anything about hupper

ztane:
Obviously no :D so perhaps just as simple as rewrite glob patterns into a regex for WatchdogFileMonitor à la fnmatch.translate (but better), trivial to support ** for recursive-glob on Python <3.5 too. Don't know how the polling should work, though

mmerickel:
Okay I added support for some variable interpolation as well as asset specs. I'm looking at adding at least some basic glob support to hupper now.

mmerickel:
This is ready for review, I added some super basic globbing support to hupper by passing the patterns into `glob.glob` directly.

bertjwregeer:
Code looks good. Testing.

bertjwregeer:
LGTM and works!

PR #2832, opened 2016-11-23
sanitize the url printed on the 404 Not Found to fix a XSS flaw

bagerard:
A recent vulnerability scan (qualys) detected that the default 404 page returned by pyramid when a request is done on an unknown route is printing the url provided in input in the html.

E.g: calling /<script>alert('hello')</script>
returns:
```
404 Not Found
The resource could not be found.
/<script>alert('hello')</script>
```
This represents an xss flaw, my suggestion is to sanitize the url before returning it, by using urllib.quote to sanitize the url

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -258,3 +258,5 @@ Contributors
 - Rami Chousein, 2015/10/28
 
 - Sri Sanketh Uppalapati, 2015/12/12
+
+- Bastien Gérard, 2016/11/23
--- a/pyramid/router.py
+++ b/pyramid/router.py
@@ -22,6 +22,7 @@ from pyramid.events import (
     NewResponse,
     )
 
+from pyramid.encode import url_quote
 from pyramid.httpexceptions import HTTPNotFound
 from pyramid.request import Request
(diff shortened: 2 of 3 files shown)


mmerickel:
Thanks for this report. I do not think this fix is appropriate however.

As you can see, there's already some escaping going on in the `HTTPException` objects being used here: https://github.com/Pylons/pyramid/blob/732c748a392a28294769d5413ef383501c3c40c9/pyramid/httpexceptions.py#L248-L316

Escaping is turned off for plain/json responses but for html it should be working. Can you figure out why this code is breaking and re-evaluate your fix?

bagerard:
Thanks for the quick feedback. I debugged the code you pointed and it is working fine. The url is escaped with webob.html_escape. Since it is rendered in the browser as html it is printed as /<script>alert('hello')</script> but in the actual response it is  /&lt;script&gt;alert(\'hello\')&lt;/script&gt;

Qualys recognized this was a false positive finding even if it recommends to avoid printing the url (even encoded) in the response. I believe the current behavior is fine and I don't see how it could be used for xss so let's discard the PR :)

PR #2834, opened 2016-11-24
Warning: view derivers only take effect for view configured afterwards

nhoening:
I ran into this because I had assumed Pyramid maintains a chain of view derivers in its config that is applied whenever a view is called. Actually, the chain is made and configured to be used by a view at the time that this view is added to the config - with whatever view derivers are known at that time.

### DIFF
--- a/docs/narr/hooks.rst
+++ b/docs/narr/hooks.rst
@@ -1772,6 +1772,11 @@ View derivers are unique in that they have access to most of the options
 passed to :meth:`pyramid.config.Configurator.add_view` in order to decide what
 to do, and they have a chance to affect every view in the application.
 
+.. warning::
+
+Only views which are added to the configuration *after* a view deriver has been
+added will be derived by it.
+
 .. _exception_view_derivers:
 
 Exception Views and View Derivers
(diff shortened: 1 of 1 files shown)


mmerickel:
This is only true if you do a commit() between the add_view and the add_view_deriver. If it is different within a single commit cycle then that's a bug.

nhoening:
I'm not calling commit(). Maybe I can find the place in the pyramid code that convinced me to move all  my `add_view` calls after the last `add_view_deriver` call (which fixed my problem) and/or I can provide a minimal pyramid example to see if it might actually be a bug.

stevepiercy:
As an aside to the actual issue, the reST syntax requires that, after a `.. warning::` directive, the content either be indented on subsequent lines (preferred) or on the same line. This is why the build fails in Travis.

nhoening:
Thanks, @stevepiercy. I fixed that. I could not confirm this issue in a simple example and actually did find a `config.commit()` in our code that I had overlooked earlier. Thanks to @mmerickel for that hint, this helps me to understand our code better. Unless a warning about being careful about `config.commit()` is useful i this regard, this PR can be closed.

mmerickel:
@nhoening great to hear. Thanks for following up so quickly!

The warning about `config.commit()` applies to the entire concept of the configurator and so I don't think there's a good place related to `add_view` and `add_view_deriver` that would be worthwhile.

PR #2842, opened 2016-11-27
Scripts optparse to argparse

stevepiercy:


pyramid/scripts/pcreate.py
```diff
@@ -23,71 +23,73 @@ def main(argv=sys.argv, quiet=False):
...
+                             'created.')
 
     pyramid_dist = pkg_resources.get_distribution("pyramid")
 
     def __init__(self, argv, quiet=False):
         self.quiet = quiet
-        self.options, self.args = self.parser.parse_args(argv[1:])
-        if not self.options.interactive and not self.options.overwrite:
-            self.options.interactive = True
+        self.args = self.parser.parse_args()
```
mmerickel: Do not throw away `argv`. Pass it to `parse_args`.

pyramid/scripts/pdistreport.py
```diff
@@ -15,7 +15,7 @@ def main(argv=sys.argv, pkg_resources=pkg_resources, platform=platform.platform,
     # all args except argv are for unit testing purposes only
     description = "Show Python distribution versions and locations in use"
     usage = "usage: %prog"
```
mmerickel: `%(prog)s`

docs/pscripts/pcreate.rst
```diff
@@ -6,8 +6,7 @@
 ``pcreate``
 -----------
 
-.. program-output:: pcreate --help
-   :prompt:
-   :shell:
+.. autoprogram:: pyramid.scripts.pcreate:PCreateCommand
```
mmerickel: This either needs to be `pyramid.scripts.pcreate:PCreateCommand.parser` or if the autoprogram package isn't smart enough to parse that then after the class definition you need to pull the parser out into a module-scope variable that autoprogram can find. This means `pyramid.scripts.pcreate:_parser` and `_parser = PCreateCommand.parser`.

docs/pscripts/pcreate.rst
```diff
@@ -6,8 +6,7 @@
...
+   :prog: pcreate.py
```
mmerickel: Depending on the rendering output we probably want this to just say `pcreate` as that's the name of the program.

pyramid/scripts/pdistreport.py
```diff
@@ -10,12 +10,16 @@ def out(*args): # pragma: no cover
         sys.stdout.write(' ')
     sys.stdout.write('\n')
 
+def get_parser():
+    _parser = argparse.ArgumentParser(
+        usage="%(prog)s",
+        description="Show Python distribution versions and locations in use")
+    return _parser
```
mmerickel: local variable should just be `parser`, not `_parser`.

stevepiercy: fixed in d8216e6

stevepiercy:
@mmerickel tests fail for pretty much everything related to p* scripts, but at least docs build. What is next?

mmerickel:
Well next is fixing the tests I suppose.

stevepiercy:
I meant I have no idea where to start to fix these tests. Take this one test. I don't know why there is an unrecognized argument.

```bash
$ env35/bin/py.test --strict pyramid/tests/test_scripts/test_pcreate.py::TestPCreateCommand::test_unknown_scaffold_name
========================================================================= test session starts ==========================================================================
platform darwin -- Python 3.5.2, pytest-3.0.3, py-1.4.31, pluggy-0.4.0
rootdir: /Users/stevepiercy/projects/hack-on-pyramid, inifile: 
collected 21 items 

pyramid/tests/test_scripts/test_pcreate.py F

=============================================================================== FAILURES ===============================================================================
____________________________________________________________ TestPCreateCommand.test_unknown_scaffold_name _____________________________________________________________

self = <pyramid.tests.test_scripts.test_pcreate.TestPCreateCommand testMethod=test_unknown_scaffold_name>

    def test_unknown_scaffold_name(self):
>       cmd = self._makeOne('-s', 'dummyXX', 'distro')

pyramid/tests/test_scripts/test_pcreate.py:60: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pyramid/tests/test_scripts/test_pcreate.py:20: in _makeOne
    cmd = tgt_class(effargs, **kw)
pyramid/scripts/pcreate.py:86: in __init__
    self.args = self.parser.parse_args(argv)
../../.pyenv/versions/3.5.2/lib/python3.5/argparse.py:1729: in parse_args
    self.error(msg % ' '.join(argv))
../../.pyenv/versions/3.5.2/lib/python3.5/argparse.py:2385: in error
    self.exit(2, _('%(prog)s: error: %(message)s\n') % args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = ArgumentParser(prog='py.test', usage='%(prog)s [options] -s <scaffold> output_directory', description='Render Pyramid ...ing to an output directory', formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
status = 2, message = 'py.test: error: unrecognized arguments: distro\n'

    def exit(self, status=0, message=None):
        if message:
            self._print_message(message, _sys.stderr)
>       _sys.exit(status)
E       SystemExit: 2

../../.pyenv/versions/3.5.2/lib/python3.5/argparse.py:2372: SystemExit
------------------------------------------------------------------------- Captured stderr call -------------------------------------------------------------------------
usage: py.test [options] -s <scaffold> output_directory
py.test: error: unrecognized arguments: distro
======================================================================= 1 failed in 0.27 seconds =======================================================================
```

mmerickel:
It's because you are passing in `argv` instead of `argv[1:]` to `parse_args()`. As such, is treating `argv[0]` as the positional argument and `argv[-1]` is detected as an extra unrecognized arg.

pyramid/scripts/pcreate.py
```diff
@@ -21,73 +22,76 @@ def main(argv=sys.argv, quiet=False):
...
+                             '(this is the default, but you may specify it to '
+                             'override --overwrite)')
+    parser.add_argument('--ignore-conflicting-name',
+                        dest='force_bad_name',
+                        action='store_true',
+                        default=False,
+                        help='Do create a project even if the chosen name '
+                             'is the name of an already existing / importable '
+                             'package.')
+    parser.add_argument('output_directory',
```
bertjwregeer: You'll need to make this optional if you want the program to behave the same as before... otherwise argparse will helpfully sys.exit() and let the user know they need to specify  an output directory...

pyramid/scripts/pcreate.py
```diff
@@ -21,73 +22,76 @@ def main(argv=sys.argv, quiet=False):
...
+        if not self.args.interactive and not self.args.overwrite:
+            self.args.interactive = True
         self.scaffolds = self.all_scaffolds()
 
     def run(self):
-        if self.options.list:
+        if self.args.list:
             return self.show_scaffolds()
-        if not self.options.scaffold_name and not self.args:
+        if not self.args.scaffold_name and not self.args:
```
bertjwregeer: `self.args` is not what it used to be, you really want to test for `self.args.output_directory`. Also, this should never be falsy because `argparse` requires the user to provide input unless you make it optional.

In `optparse`:

- `self.options`: Contained named options/flags
- `self.args`: Contained whatever arguments were left over (positional arguments)

in `argparse` optional arguments are now rolled into `self.args` as well, and providing more arguments than expected will cause `argparse` to call sys.exit while providing the help text.

bertjwregeer:
As mentioned on IRC the big difference between `optparse` and `argparse` is that `optparse` would happily let the program continue on it's merry way if you didn't provide the right flags/input. This is not the case for `argparse` which will happily call `sys.exit` for you when the wrong input is created.

This is causing issues for you because this will call `sys.exit` inside of your running py-test. You need to find a way to make ArgumentParser not exit see [PEP-0389][1] for how to accomplish that, or change the tests to run the script in a sub-process and capture the output.

[1]: https://www.python.org/dev/peps/pep-0389/#discussion-sys-stderr-and-sys-exit

CHANGES.txt
```diff
@@ -42,6 +42,9 @@ Backward Incompatibilities
 Features
 --------
 
+- Python 3.6 compatibility.
```
bertjwregeer: I feel like this doesn't belong in this PR...

stevepiercy: Yeah, if https://github.com/Pylons/pyramid/pull/2840 is merged, I can rebase on master and push, and these 3.6 items should go away.

pyramid/scripts/pcreate.py
stevepiercy: I think leaving `output_directory` as required is a more helpful behavior from the user perspective. The behavior would not change, except in the messages returned. argparse's messages seem more user friendly and informative, too.

```bash
# master
bash-3.2$ ../env35/bin/pcreate -s starter
usage: pcreate [options] -s <scaffold> output_directory
pcreate: error: the following arguments are required: output_directory

# 1.7-branch
bash-3.2$ ../env35/bin/pcreate -s starter
You must provide a project name
```

If it were optional, then that would imply that the program could run without an `output_directory` option. That in turn implies either the current working directory would be the default, or that the program be interactive and require an input. This has less appeal to me.

stevepiercy:
> You need to find a way to make ArgumentParser not exit see [PEP-0389](https://www.python.org/dev/peps/pep-0389/#discussion-sys-stderr-and-sys-exit) for how to accomplish that, or change the tests to run the script in a sub-process and capture the output.

Which direction do you think is better? I'm leaning toward `subprocess` in the tests, because I prefer the messages of `argparse` as they are.

I hunted for an example for subclassing `argparse.ArgumentParser` and overriding `exit`. PEP-0389 had none, but I did find a [Python issue](http://bugs.python.org/issue9938) which had a link to an [example in IPython](https://github.com/ipython/ipython/blob/master/IPython/core/magic_arguments.py#L105), but that example looks funky. The [docs on argparse](https://docs.python.org/3/library/argparse.html#exiting-methods) are barren of examples, too.

bertjwregeer:
>  I'm leaning toward `subprocess` in the tests, because I prefer the messages of `argparse` as they are.

`subprocess` is probably a good idea.

> I hunted for an example for subclassing `argparse.ArgumentParser` and overriding `exit`.

I don't think there is a solid example, just explaining what the process is and what to do.

mmerickel:
1)  here's an example of subclassing the parser to raise an alternate error. This could easily be done with a little refactoring to allow the test harness to patch the class used.

https://github.com/mmerickel/subparse/blob/1526ae890f3b4736bd579ad010cd1a15d444bc66/subparse/__init__.py#L9-L15

2) A `sys.exit()` does not actually kill the process immediately. It simply raises a `SystemExit` exception. This can be caught in the test. I do not think using subprocess should be necessary here. This is by far the easiest solution for testing the arguments.

bertjwregeer:
> A sys.exit() does not actually kill the process immediately. It simply raises a SystemExit exception. This can be caught in the test. I do not think using subprocess should be necessary here. This is by far the easiest solution for testing the arguments.

Ohhh, I was not aware of this. This makes sense.

stevepiercy:
> allow the test harness to patch the class used.

Sorry, I don't understand. I don't know what is a test harness or how to patch the used class.

Is the test harness `_makeOne` and `setUp` methods, or something else?

Is the used class `TestPCreateCommand`, `PCreateCommand`, `argparse.ArgumentParser`, or something else?

mmerickel:
Just go with option 2. I don't really want to explain how to write testable code in an issue and if I were implementing it on this branch right now I'd just use option 2 myself.

CHANGES.txt
stevepiercy: @bertjwregeer these did not go away after pushing a few commits. Can you help?

mmerickel: It means you never rebased this on the version of master with the commits.

pyramid/scripts/pcreate.py
```diff
@@ -21,73 +22,78 @@ def main(argv=sys.argv, quiet=False):
...
-    parser.add_option('--ignore-conflicting-name',
-                      dest='force_bad_name',
-                      action='store_true',
-                      default=False,
-                      help='Do create a project even if the chosen name '
-                           'is the name of an already existing / importable '
-                           'package.')
+    verbosity = 1  # required
+    parser = argparse.ArgumentParser(
+        prog="pcreate",
```
mmerickel: `prog` is unnecessary... usually we'd leave this unspecified. For the docs you're already forcefully setting it via `:prog:` so that they will render with the selected name.

pyramid/scripts/pcreate.py
mmerickel: As @bertjwregeer said, testing `self.args` as a boolean makes no sense here and it should be `self.args.output_directory`.

pyramid/scripts/prequest.py
```diff
@@ -139,10 +149,10 @@ def run(self):
                 name, value = item.split(':', 1)
                 headers[name] = value.strip()
 
-        app = self.get_app(app_spec, self.options.app_name,
-                options=parse_vars(self.args[2:]))
+        app = self.get_app(app_spec, self.args.app_name,
+                options=self.args)
```
mmerickel: nope, this makes no sense

pyramid/scripts/pserve.py
```diff
@@ -48,92 +47,106 @@ class PServeCommand(object):
...
         dest='verbose',
         help="Suppress verbose output")
+    parser.add_argument(
+        'config_uri',
+        nargs='?',
+        default=None,
+        help='The URI to the configuration file.',
+        )
+    parser.add_argument(
+        'config_args',
```
mmerickel: `config_vars` not `config_args`

pyramid/scripts/pserve.py
```diff
@@ -48,92 +47,106 @@ class PServeCommand(object):
     """
     default_verbosity = 1
 
-    parser = optparse.OptionParser(
-        usage,
+    parser = argparse.ArgumentParser(
+        usage="%(prog)s config_uri [var=value]",
```
mmerickel: usage is generated by argparse... was it not acceptable? Why is it being overridden?

pyramid/scripts/pserve.py
```diff
@@ -48,92 +47,106 @@ class PServeCommand(object):
...
+        'config_uri',
+        nargs='?',
+        default=None,
+        help='The URI to the configuration file.',
+        )
+    parser.add_argument(
+        'config_args',
+        nargs='*',
+        default=(),
+        help='Arbitrary options to override those in the [app:main] section '
```
mmerickel: Nope... not `app:main` section. Should say something more like `Variables required by the config file. For example, http_port=%%(http_port)s would expect "http_port=8080" to be passed here.`.

pyramid/scripts/pserve.py
```diff
@@ -48,92 +47,106 @@ class PServeCommand(object):
...
+            self.args.verbose = 0
 
     def out(self, msg): # pragma: no cover
-        if self.options.verbose > 0:
+        if self.args.verbose > 0:
             print(msg)
 
-    def get_options(self):
-        restvars = self.args[1:]
+    def get_config_args(self):
```
mmerickel: `config_vars`

pyramid/tests/test_scripts/test_pcreate.py
```diff
@@ -50,11 +50,15 @@ def test_run_no_scaffold_name(self):
...
-        self.assertTrue(out.startswith('You must provide a project name'))
+        try:
+            cmd = self._makeOne('-s', 'dummy')
+            result = cmd.run()
+            self.assertEqual(result, 2)
+            out = self.out_.getvalue()
+            self.assertTrue(out.startswith('You must provide a project name'))
+        except SystemExit as e:
+            code = e.args[0]
+            self.assertEqual(code, 2)
```
mmerickel: This test shouldn't have needed to be changed - and if it did then it's not properly asserting when the except is not called.

pyramid/tests/test_scripts/test_proutes.py
```diff
@@ -20,7 +20,7 @@ def _getTargetClass(self):
     def _makeOne(self):
         cmd = self._getTargetClass()([])
         cmd.bootstrap = (dummy.DummyBootstrap(),)
-        cmd.args = ('/foo/bar/myapp.ini#myapp',)
+        cmd.args.config_uri = ('/foo/bar/myapp.ini#myapp',)
```
mmerickel: This is not right. Why is `config_uri` a tuple?

pyramid/scripts/pcreate.py
stevepiercy: @mmerickel Removed all the progs. Ready for review.

pyramid/scripts/pserve.py
stevepiercy: refactored this and other instances

pyramid/scripts/pserve.py
stevepiercy: Updated.

However is there supposed to be one or two `%`?

pyramid/scripts/pserve.py
stevepiercy: refactored

pyramid/tests/test_scripts/test_pcreate.py
stevepiercy: reverted to original

stevepiercy:
I thought the proper way was to create a branch on my fork, and create a branch in pyramid with the same name, so that others could easily push to the branch in pyramid.

I can delete this branch in pyramid, then submit the PR from the branch in my fork to master in Pyramid.

pyramid/tests/test_scripts/test_proutes.py
stevepiercy: failure to adjust original. updated.

pyramid/scripts/pcreate.py
stevepiercy: I restore this a while ago to:

```python
        self.args = self.parser.parse_args(argv[1:])
```

docs/pscripts/pcreate.rst
stevepiercy: I've updated all instances to use the proper syntax for autoprogram.

docs/pscripts/pcreate.rst
stevepiercy: I dropped .py from all instances.

pyramid/scripts/pcreate.py
stevepiercy: I've fixed this to test for `self.args.output_directory`.

mmerickel:
You can put the branch wherever, but you shouldn't make us review a branch that is not "X to master". In its current state it's not possible to compare what things were in master to what you want them to be.

stevepiercy:
Closing in favor of #2864

PR #2846, opened 2016-12-05
A helper for quick starting a simple Pyramid app

mfrlin:
When comparing the Flask and Pyramid quick start example,
it is obvious that we could significantly simplify Pyramid’s one
by having a method such as this.

Consider the following examples:

BEFORE
```python
from pyramid.view import view_config

@view_config(route_name='hello', renderer='string')
def hello_world(request):
    return {'content': 'Hello World!'}

if __name__ == '__main__':
    config = Configurator()
    config.add_route('hello', '/')
    config.scan()
    app = config.make_wsgi_app()
    server.serve_forever()
```

AFTER
```python
from pyramid.view import view_config

@view_config(route_name='hello', renderer='string')
def hello_world(request):
    return {'content': 'Hello World!'}

if __name__ == '__main__':
    config = Configurator()
    config.add_route('hello', '/')
    config.scan()
    config.run_app()
```

Such a helper makes writing simple tutorials more, well, simple.

### DIFF
--- a/pyramid/config/__init__.py
+++ b/pyramid/config/__init__.py
@@ -329,6 +329,13 @@ class Configurator(
                 exceptionresponse_view=exceptionresponse_view,
                 )
 
+    def run_app(self, host='localhost', port=8080, **kwargs):
+        from wsgiref.simple_server import make_server
+        app = self.make_wsgi_app()
+        server = make_server(host, port, app, **kwargs)
+        print('Running on http://{}:{} (stop with Ctrl-C)'.format(host, port))
+        server.serve_forever()
+
     def setup_registry(self,
                        settings=None,
                        root_factory=None,
--- a/pyramid/config/__init__.py
+++ b/pyramid/config/__init__.py
@@ -329,11 +329,15 @@ class Configurator(
                 exceptionresponse_view=exceptionresponse_view,
                 )
 
-    def run_app(self, host='localhost', port=8080, **kwargs):
+    def run_app(self, host='localhost', port=8080):
+        """ Helper method to quickly run a development server listening on ``host`` and ``port``."""
         from wsgiref.simple_server import make_server
         app = self.make_wsgi_app()
-        server = make_server(host, port, app, **kwargs)
+        server = make_server(host, port, app)
         print('Running on http://{}:{} (stop with Ctrl-C)'.format(host, port))
+        docs_link = 'http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/project.html#using-an-alternate-wsgi-server'
+        print('This is running on wsgiref server and is not appropriate for production' \
+              'please see: {}'.format(docs_link))
         server.serve_forever()
 
     def setup_registry(self,
--- a/pyramid/tests/test_config/test_init.py
+++ b/pyramid/tests/test_config/test_init.py
@@ -1996,3 +1996,23 @@ class DummyIntrospectable(object):
         
 class DummyPredicate(object):
     pass
+
+class TestRunAppHelper(unittest.TestCase):
+
+    def test_run_app(self):
+        from pyramid.config import Configurator
+        from pyramid.response import Response
+
+        self.running = False
+        # Mocks wsgiref's serve_forever method
+        def serve_forever_mock(*args, **kwargs):
+            self.running = True
+
+        from wsgiref.simple_server import WSGIServer
+        WSGIServer.serve_forever = serve_forever_mock
+
+        config = Configurator()
+
+        config.run_app()
+        self.assertTrue(self.running)
+


zupo:
Docstring and tests please.

stevepiercy:
This appears to be a change for documentation. Why did you put this in Pyramid core? Please see [docs/narr/helloworld.py](https://github.com/Pylons/pyramid/blob/master/docs/narr/helloworld.py)

mfrlin:
This is here because it implements run_app helper method in Configurator class.

mmerickel:
The idea of integrating a `run_app` or whatever into the configurator has come up before and I'm a strong -1 on it. I think it's much better to show the user that the WSGI server and the WSGI app are separate entities. It would require making pyramid depend on a particular implementation because anything shipped with the stdlib is not suitable for production. Flask gets away with this because their core library (werkzeug) happens to have a wsgi server in it.

mfrlin:
I've added the warning about production usage. I've also added docstring and test.

Also, not only Flask uses .run(), but also Bottle, CherryPy, Django, ... All have development, not production ready webservers.

mmerickel:
We add webservers to the scaffolds which are out-of-the-box ready to run similar to this setup. The core configurator is not where this belongs for many reasons.

1) The `config` object will never be garbage collected, despite its purpose being to configure the app, not to run the app.

2) A wsgi app should have no opinions what wsgi server it is run with.

3) Many people need different wsgi servers for different reasons (green threads, websockets, asyncio).

Once this support is in a core API inside of Pyramid, it is now considered the defacto standard and I will start receiving issues about why it can't just also run waitress, or gevent, or gunicorn, or or or.. if I pass this one other option to `run_app`, or don't set some debug option..

Fundamentally a wsgi app should not care what wsgi server it is hosted on and this conflates that issue under the guise of "usability for newcomers" which is the entire purpose of the scaffolds / cookiecutters which offer a production-ready setup.

I'm really sorry, I do appreciate you trying to come up with ways to make things easier, but I'm not going to accept this feature unless other maintainers decide to override me.

bertjwregeer:
I'm all for making things simpler for users, I am a huge -1 on this. I don't think it is a good idea to integrate `wsgiref`, and I am not sure we want to make a decision for the user on what WSGI server we ship by default if we need to pick something other than `wsgiref`.

`wsgiref` has led to various issues in the past too, because it doesn't actually implement WSGI correctly, and we've had bugs filed against Pyramid/WebOb because of it.

This PR allows a single example to remove 3 lines of code, I don't think 3 lines of code are that difficult to grok.

blaflamme:
While I understand and like the benefits of this idea I agree with @mmerickel and @bertjwregeer on this. I don't think the benefits worth the shortcomings from the Pyramid architecture POV.

zupo:
I Googled around for python web (micro)frameworks and I have found that the great majority have a runnable example on their front page. 

I understand that putting something into core for marketing purposes might not be pure or ideal. But  can we at least consider what our options are? To the screenshots:


Hug does it like so:
screen shot 2016-12-07 at 10 08 14

Probably the shortest quick example, by Bottle, uses the ``run`` method (similar to the one proposed in this PR):
screen shot 2016-12-07 at 10 08 35

Falcon defaults to using gunicorn in example, can *we* do this and why not?
screen shot 2016-12-07 at 10 09 04

I like CherryPy's approach the most, ``quickstart`` clearly conveys the information that this method of running the server is for quick hacks and not for production means:
screen shot 2016-12-07 at 10 09 47

And lastly, Flask:
screen shot 2016-12-07 at 10 09 40


For comparison, a newcomer coming to TryPyramid.com must first click the correct link to get to http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html, then find the tiny ``next`` link in the footer to then go over this blob of text to get to the same point as with the frameworks above:

screen shot 2016-12-07 at 10 19 42

stevepiercy:
@zupo this issue is about adding code and docs to Pyramid core. It's also in the Pyramid repo, not the trypyramid.com repo, so I'm confused about what you are advocating and where.

zupo:
I'm advocating adding a helper to Pyramid core that allows starting a Pyramid app with a single line.

mmerickel:
The only thing I can picture right now that doesn't ruffle my feathers too much is something along the lines of the following:

```
from pyramid.view import view_config

@view_config(route_name='hello', renderer='string')
def hello_world(request):
    return {'content': 'Hello World!'}

if __name__ == '__main__':
    config = Configurator()
    config.add_route('hello', '/')
    config.scan()
    app = config.make_wsgi_app()
    app.run_super_unsafe_server_that_should_never_be_used_in_production()
```

Alternative names include `run_quickstart_server`, `run_demo_server`, `run_development_server`, or `run_crappy_singlethreaded_server`.

On the whole I appreciate the concern this PR is attempting to solve, but as someone maintaining this software and envisioning the future ways people may want to expand this feature to work better for them in production I'm simply frightened about the costs associated with maintaining it in terms of even simply having to say no to any new hooks.

zupo:
``run_crappy_singlethreaded_server`` :)

We do need to think hard to minimize maintenance impact, indeed. There are several things that might help:
* have a descriptive method (already presented)
* have a big honking print to console saying this is only meant for demo purposes
* print URL pointing to documentation on how to set it up for production
* write up a straightforward document on how to migrate from crappy singlethreaded server to one server we choose in advance to be Good Enough TM (waitress?) with references to configuration for other potential production servers (I am willing to take this on)
* Inject a warning sign into HTML (quite aggressive, but then again ...)
* anything else?

zupo:
Related: https://github.com/Pylons/trypyramid.com/pull/211

On can do this:
```python
from pyramid.scripts.pserve import wsgiref_server_runner
...
    wsgiref_server_runner(config.make_wsgi_app(), None)
```

mmerickel:
Do not use `wsgiref_server_runner`. It's a special function conforming to the server runner api and should not be used by user code directly. We could add a different API that was intended as such, but wsgiref is in the stdlib so I'm just amazed at the hoops we're going through here to save 1 line (original is 3 - import, make server and run server, this one is 2 - import and run server).

bertjwregeer:
I'm not a big fan of importing the wsgiref_server_runner that is meant to be used in paste .ini scripts only. Especially since it doesn't print a warning that it should not be used in production. I am really worried about that.

zupo:
My point mostly was: Pyramid already ships with a function that starts a server. We either need to
* fix it: make its API nicer, make it run the "correct" server the "correct" way, print out warnings, ..
* remove it, because if I found it, someone else will too and will use it

mfrlin:
Is there really a big difference if the quick runner lives in `pyramid.wsgi` or is on `Configurator`? For me module wsgi implies that pyramid offers some wsgi functionality which it does not.

Here is a list of requirements that implementation should have. Feel free to add more:

- Warning in docstring
- Warning printed out with a link to documentation (use warnings module?)
- Documentation explaining why you should not use this in production and better ways of running your app. Here is one possible location for this documentation: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/project.html#using-an-alternate-wsgi-server

mmerickel:
> Is there really a big difference if the quick runner lives in pyramid.wsgi or is on Configurator?

Yes, I put a technical issue in https://github.com/Pylons/pyramid/pull/2846#issuecomment-264916349 above.

bertjwregeer:
> My point mostly was: Pyramid already ships with a function that starts a server.

We provide `pserve` that uses paste scripts to run a WSGI server.

That function inside the scripts is NOT public API other than using it from a paste ini config. If someone else "finds" it and uses it, they bought it, they own it.

PR #2849, opened 2016-12-06
Changed wiki tutorial to showcase passwrd hashing with bcrypt.

mfrlin:
Relates to #2204

Since this tutorial does not store user data in a database I have just showcased how hashing works and explained more strongly that in real applications you are probably gong to store users in a database.

Would appreciate if someone would have read through and suggest corrections if something needs more explaining or stronger warnings about clear text dangers or something.

Disregard branch name.

### DIFF
--- a/docs/tutorials/wiki/authorization.rst
+++ b/docs/tutorials/wiki/authorization.rst
@@ -43,7 +43,7 @@ Add dependencies
 ~~~~~~~~~~~~~~~~
 
 Just like in :ref:`wiki_defining_views` we need a new dependency.
-We need to add the ``bcrypt`` package, to our tutorial package's
+We need to add the ``bcrypt`` [1]_ package, to our tutorial package's
 ``setup.py`` file by assigning this dependency to the ``requires`` parameter
 in the ``setup()`` function.
 
@@ -81,15 +81,15 @@ request)`` returns ``None``.  We will use ``groupfinder()`` as an
 :term:`authentication policy` "callback" that will provide the
 :term:`principal` or principals for a user.
 
-There are two helper methods that will help us later when loging-in users.
+There are two helper methods that will help us later to authenticate users.
 The first is ``hash_password`` which takes a raw password and transforms it using
-bcrypt into an irreversible representation, a process known as "hashing". The
+bcrypt_ into an irreversible representation, a process known as "hashing". The
 second method, ``check_password``, will allow us to compare the hashed value of the
 submitted password against the hashed value of the password stored in the user's
 record. If the two hashed values match, then the submitted
 password is valid, and we can authenticate the user.
 
-We hash passwords so that it is impossible to decrypt them and use them to
+We hash passwords so that it is impossible to decrypt and use them to
 authenticate in the application. If we stored passwords foolishly in clear text,
 then anyone with access to the database could retrieve any password to authenticate
 as any user.
@@ -403,3 +403,11 @@ following URLs, checking that the result is as expected:
   the login form with the ``editor`` credentials), we'll see a Logout link in
   the upper right hand corner.  When we click it, we're logged out, and
   redirected back to the front page.
+
+
+.. _bcrypt: https://pypi.python.org/pypi/bcrypt
+
+.. [1] We are using the bcrypt_ package from PyPI to hash our passwords
+       securely. There are other one-way hash algorithms for passwords if
+       bcrypt is an issue on your system. Just make sure that it's an
+       algorithm approved for storing passwords versus a generic one-way hash.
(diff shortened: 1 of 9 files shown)


docs/tutorials/wiki/authorization.rst
```diff
@@ -61,7 +81,20 @@ request)`` returns ``None``.  We will use ``groupfinder()`` as an
 :term:`authentication policy` "callback" that will provide the
 :term:`principal` or principals for a user.
 
-In a production system, user and group data will most often come from a
+There are two helper methods that will help us later when loging-in users.
```
mmerickel: `s/loging-in/authenticating` perhaps? I think loging-in is quite difficult to parse if it is even a real word.

docs/tutorials/wiki/src/authorization/tutorial/security.py
```diff
@@ -1,5 +1,16 @@
-USERS = {'editor':'editor',
-          'viewer':'viewer'}
+import bcrypt
+
+
+def hash_password(pw):
+    return bcrypt.hashpw(pw.encode('utf-8'), bcrypt.gensalt())
```
mmerickel: Just to clarify, we are aware we are returning bytes here, and happy with that? This is slightly different from the wiki2 [1] tutorial which keeps things as unicode strings at all times.

[1] https://github.com/Pylons/pyramid/blob/7ab666e3e2539662ff3324a4bdedbca3c8f8040a/docs/tutorials/wiki2/src/authentication/tutorial/models/user.py#L20-L28

docs/tutorials/wiki/src/authorization/tutorial/security.py
```diff
@@ -1,5 +1,16 @@
...
+
+def check_password(expected_hash, pw):
+    if expected_hash is not None:
```
mmerickel: Similarly the `expected_hash` here is expected to always be bytes, right?

docs/tutorials/wiki/authorization.rst
```diff
@@ -61,7 +81,20 @@ request)`` returns ``None``.  We will use ``groupfinder()`` as an
...
+The first is ``hash_password`` which takes a raw password and transforms it using
+bcrypt into an irreversible representation, a process known as "hashing". The
+second method, ``check_password``, will allow us to compare the hashed value of the
+submitted password against the hashed value of the password stored in the user's
+record. If the two hashed values match, then the submitted
+password is valid, and we can authenticate the user.
+
+We hash passwords so that it is impossible to decrypt them and use them to
```
stevepiercy: Drop extra "them" after decrypt:
We hash passwords so that it is impossible to decrypt and use them to

docs/tutorials/wiki/src/authorization/tutorial/security.py
stevepiercy: I would like wiki and wiki2 to have consistent implementations, whichever is preferred.

stevepiercy:
@mfrlin thanks for taking this on! Please let me know when it is ready for a human to run through manually, and I'll do that task to check your work, making sure docs align with code and other details.

docs/tutorials/wiki/authorization.rst
stevepiercy: Drop extra "them" after decrypt:
We hash passwords so that it is impossible to decrypt and use them to

docs/tutorials/wiki/src/authorization/tutorial/security.py
mmerickel: I'm not strictly against it since zodb can just handle the bytes, and the data is ascii, but I think unicode is slightly more accurate.

mfrlin:
Ok, I've made implementation of helper functions consistent with wiki2 example. My reasoning was to make things simpler because there was no reason to convert hashes back to unicode.

I've also fixed wording with changes you suggested.

Added footnote from wiki2 example and a link to bcrypt.

Line numbers and highlighting are all good from what I can tell.

I've also had concern about bcrypt because it did not install on my system until I installed libffi-dev package. But I think installation is out of scope of this tutorial and footnote kinda covers it?

mmerickel:
Thank you @mfrlin!

mmerickel:
@mfrlin What system are you having issues installing bcrypt on? It has a manylinux1 wheel so it should work fine on almost all systems unless you are not using a recent version of pip. For example buildout or `setup.py develop`.

mfrlin:
@mmerickel I'm using Ubuntu 15.10, so maybe I am somewhat behind the times. Can't test on other systems currently.

ztane:
@mfrlin note that ubuntu 15.10 was eoled long ago already.

Issue #2548, opened 2016-05-07
use integrity attribute for javascripts from CDN


stevepiercy:
Our scaffolds and possibly other items reference jQuery without an integrity attribute. From http://jquery.com/download/#jquery-39-s-cdn-provided-by-maxcdn

> The jQuery CDN supports https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity (SRI) which allows the browser to verify that the files being delivered have not been modified. This https://www.w3.org/TR/SRI/ is currently being implemented by browsers. Adding the new integrity attribute will ensure your application gains this security improvement as browsers support it.

``` html
<script
   src="https://code.jquery.com/jquery-1.12.3.min.js"
   integrity="sha256-aaODHAgvwQW1bFOGXMeX+pC4PZIPsvn2h1sArYOhgXQ="
   crossorigin="anonymous"></script>
```

ztane:
... and the important thing is to remember to change these integrity stamps as you update versions.

It happened to me that I copied these integrity stamps verbatim from getbootstrap.com, worked for me (on Firefox) and failed spectacularly then on customer's Chrome :D

ikolar:
Working on it ...

fangpenlin:
I can handle this

stevepiercy:
So... I hate to say this and I apologize for not stating this up front, but things have changed since this issue was created and I did not provide a complete description:

* scaffolds are deprecated in favor of cookiecutters (the submitted changes to the scaffolds are still acceptable)
* the three supported cookiecutters should also be updated:
  * https://github.com/Pylons/pyramid-cookiecutter-starter
  * https://github.com/Pylons/pyramid-cookiecutter-alchemy
  * https://github.com/Pylons/pyramid-cookiecutter-zodb
* an update to scaffolds/cookiecutters requires an update to documentation that uses source files, specifically every tutorial step. For example, a quick search for `<script src="//oss.maxcdn.com/libs/html5shiv/3.7.0/html5shiv.js"></script>` found 25 instances in 25 files

If you could also update all those places, then I'd prefer to merge all the PRs in one swoop. Thank you!

fangpenlin:
it's all good, let me try to fix them all at once

PR #2850, opened 2016-12-06
Fix for issue #2548 - use integrity attribute for javascripts from CDN in fixtures

ikolar:
Used fgrep to find css and js includes from CDNs in fixtures.. Calculated their SHA hashes with:
 $ curl -s "$url" | openssl dgst -sha384 -binary | openssl base64 -A

Inserted the hashes into three fixture templates.

Verified by opening them in Chrome (there were no errors in the console). Garbled the
hash by hand to double-check.

Hope all is ok.

(PS. There are some more includes inside the docs/ html files. I can fix those as well.)

### DIFF
--- a/pyramid/scaffolds/alchemy/+package+/templates/layout.jinja2_tmpl
+++ b/pyramid/scaffolds/alchemy/+package+/templates/layout.jinja2_tmpl
@@ -11,15 +11,15 @@
     <title>Alchemy Scaffold for The Pyramid Web Framework</title>
 
     <!-- Bootstrap core CSS -->
-    <link href="//oss.maxcdn.com/libs/twitter-bootstrap/3.0.3/css/bootstrap.min.css" rel="stylesheet">
+    <link href="//oss.maxcdn.com/libs/twitter-bootstrap/3.0.3/css/bootstrap.min.css" integrity="sha384-IS73LIqjtYesmURkDE9MXKbXqYA8rvKEp/ghicjem7Vc3mGRdQRptJSz60tvrB6+" crossorigin="anonymous" rel="stylesheet">
 
     <!-- Custom styles for this scaffold -->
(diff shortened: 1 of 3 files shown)


mmerickel:
Can you explain the steps necessary to maintain this?

Also can you please contribute patches to the cookiecutters, as these scaffolds are going to be deprecated in the next release?

stevepiercy:
Suggest using http://www.jsdelivr.com/ to search for:

* jQuery
* Bootstrap CSS and JS
* HTML5Shiv
* respond.js

and selecting appropriate versions. It would be nice to have a singular URL to save the search step, although it appears that you can create a collection which gets stored in a cookie.

mmerickel:
@stevepiercy What?

stevepiercy:
@mmerickel What did you mean about steps to maintain? This is one step in that process.

mmerickel:
Did you just ask me what I meant and then claim your comment was a step in the thing you didn't know about? :-)

I asked above what adding the integrity attributes means with respect to maintenance of the tutorials? Hopefully it just means that if we update bootstrap then we need to update the integrity hashes, but I want to know before this is merged what it means for the maintenance work load. I don't see what jsdelivr and cookies has to do with this at all.

stevepiercy:
How do you get the hashes for the 5 (or possibly more) CDN served assets?

mmerickel:
I'll ask again just incase this got lost in the noise. Please update the cookiecutter repositories with these changes. The scaffolds themselves are deprecated and so I'm going to avoid merging this change without similar updates to the cookiecutters. It doesn't have to be done by you, but until it's done I'm leaving this PR open as a reminder.

mmerickel:
> (PS. There are some more includes inside the docs/ html files. I can fix those as well.)

Please update them as well.

I'm concerned about the fact that this PR is now from an "unknown repository" so I'm closing this. If you still want to work on this please comment and we'll re-open it, or make a new PR.

Issue #1573, opened 2015-02-10
Improve CSRF Support for Safety, REST, and Decoupling


dstufft:
The current CSRF support in Pyramid suffers from a few problems:
- It's needlessly coupled to sessions, if you wish to use a CSRF token without a session you have to make a fake `ISession` which has exception raising session methods.
- Since it uses the predicate system, failing a CSRF check is signaled back to the user as a 404 instead of something more appropriate like a 403 error. This makes users feel like they got the wrong URL not the wrong CSRF token.
- Views which use the CSRF token need a Vary: header based on the kind of session that it's using (`Vary: Cookie` for example or `Vary: Authorization`), however the current CSRF system doesn't do anything about this.
- The CSRF protection is opt in, which means that it's extremely easy for developers to accidentally leave it off. If they leave it off then their views will be vulnerable to CSRF but that is a non-obvious situation that, like many security sensitive things, is non obvious unless you're looking for it since everything will appear to work.

I extracted some CSRF stuff from an application I'm writing in Pyramid and sketched out some basic ideas for how I think CSRF should be implemented. That's available (without tests or documentation) at https://github.com/dstufft/pyramid_csrf.

The basic idea there is:
1. Instead of making CSRF coupled to `ISession`, it adds it's own interface `ICSRF` (and `ICSRFFactory`), the `ICSRF` instance is made available as `request.csrf`.
   - Splitting this off of session means that CSRF can be implemented without requiring a session to be around. My pyramid_csrf app includes 3 different implementations, one is `CookieCSRF` which just stores the tokens it generates inside of a dedicated cookie, another is `SessionCSRF` which just stores the tokens it generates inside of the session, and the final is `LegacySessionCSRF` which doesn't store or generate tokens itself at all, and instead it just dispatches to the legacy methods on `ISession`.
2. Instead of using the predicate system, it relies on being part of the request -> response cycle of calling a view, this means that instead of getting a 404 error it'll have a 403 error raised while attempting to call the view.
3. Since (again) this is part of the request -> response cycle, it's able to smartly add `Vary` headers to the response to ensure that caches don't serve a page with a CSRF token cached inside of them.
4. It adds code in the default response path that ensure CSRF checking is done, in pyramid_csrf this is done in a hacky way by creating a dynamic subclass of the current view mapper that wraps every view with a decorator, however with support in Pyramid for some sort of on by default mechanism or just baking CSRF directly into it this hack wouldn't need to exist.
[...]

PR #2854, opened 2016-12-07
Decouple CSRF protection from the session machinery

MatthewWilkes:
In #1573 @dstufft makes a number of feature requests about the CSRF implementation of Pyramid, most of which were added in 1.7.

This patch attempts to move the CSRF generation logic out of the session, so new CSRF implementations that do not depend on sessions can be added. It currently does not add any such implementations, the only one shipped is the default which falls back to an implementation of CSRF on the session.

It also introduces new utility functions for `get_csrf_token(request)` and `new_csrf_token(request)` in Python, as well as a `get_csrf_token()` in the template global scope, and repoints documentation to use these rather than explicitly calling the session machinery.

This was written at the DragonSprint by me and @jcerjak, a new contributor to Pyramid.


pyramid/config/security.py
```diff
@@ -218,9 +227,14 @@ def register():
         intr['header'] = header
         intr['safe_methods'] = as_sorted_tuple(safe_methods)
         intr['callback'] = callback
+
+        from pyramid.csrf import csrf_token_template_global
+        from pyramid.events import BeforeRender
```
jcerjak: Could we move these imports to the top?

MatthewWilkes: Done.

pyramid/csrf.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def csrf_token_template_global(event):
+    request = event.get('request', None)
+    try:
+        registry = request.registry
+        csrf = registry.queryUtility(ICSRF)
+        if csrf is not None:
+            event['get_csrf_token'] = partial(csrf.get_csrf_token, request)
+    except AttributeError:
```
MatthewWilkes: This also catches the case when request is None or some weird mock object, like in tests. I've moved the `except` handler only to cover request.registry and the rest in in an `else` handler.

pyramid/csrf.py
```diff
@@ -0,0 +1,225 @@
...
+    urlparse,
+    )
+
+from pyramid.exceptions import (
+    BadCSRFOrigin,
+    BadCSRFToken,
+)
+from pyramid.interfaces import ICSRF
+from pyramid.settings import aslist
+from zope.interface import implementer
```
tseaver: Third-party imports should be prior to "local" imports.

pyramid/csrf.py
tseaver: The utility promises to implement ICSRF, which guarantees that it has a `get_csrf_token` method.  We don't need to defend against borked implementations.

docs/narr/security.rst
```diff
@@ -765,3 +765,192 @@ which would allow the attacker to control the content of the payload.  Re-using
...
+load a URL because it is linked from, or embedded in, an attacker's website.
+If the URL is one that may modify or delete data, the consequences can be dire.
+
+You can avoid most of these attacks by issuing a unique token to the browser
+and then requiring that it be present in all potentially unsafe requests.
+:app:`Pyramid` sessions provide facilities to create and check CSRF tokens.
+
+To use CSRF tokens, you must first enable a :term:`session factory` as
+described in :ref:`using_the_default_session_factory` or
+:ref:`using_alternate_session_factories`.
```
tseaver: There should be language here which points up the fact that enabling sessions is a requirement of the default `ICSRF` implementation, rather than a general requirement.

docs/narr/security.rst
```diff
@@ -765,3 +765,192 @@ which would allow the attacker to control the content of the payload.  Re-using
...
+To get the current CSRF token, use the
+:data:`pyramid.csrf.get_csrf_token` method.
+
+.. code-block:: python
+
+   from pyramid.csrf import get_csrf_token
+   token = get_csrf_token(request)
+
+The ``get_csrf_token()`` method accepts a single argument, of the request. It
+returns a CSRF *token* string. If ``get_csrf_token()`` or ``new_csrf_token()``
```
tseaver: > accepts a single argument, of the request.

The "of" should not be there.

docs/narr/security.rst
```diff
@@ -765,3 +765,192 @@ which would allow the attacker to control the content of the payload.  Re-using
...
+    <form method="post" action="/myview">
+      <input type="hidden" name="csrf_token" value="${get_csrf_token()}">
+      <input type="submit" value="Delete Everything">
+    </form>
+
+Or include it as a header in a jQuery AJAX request:
+
+.. code-block:: javascript
+
+    var csrfToken = ${get_csrf_token()};
```
tseaver: My JS-fu may be weak, but ISTM that the token value needs to be wrapped in quotes here.

docs/narr/security.rst
```diff
@@ -765,3 +765,192 @@ which would allow the attacker to control the content of the payload.  Re-using
...
+method.  This differs only from ``csrf.get_csrf_token()`` inasmuch as it
+clears any existing CSRF token, creates a new CSRF token, sets the token into
+the user, and returns the token.
+
+.. code-block:: python
+
+   from pyramid.csrf import get_csrf_token
+   token = new_csrf_token()
+
+Note, it is not possible to force a new CSRF token from a template. If you want
```
tseaver: Use `.. note::`.

pyramid/csrf.py
```diff
@@ -0,0 +1,225 @@
...
+    """ Get the currently active CSRF token for the request passed, generating
+    a new one using ``new_csrf_token(request)`` if one does not exist. This
+    calls the equivalent method in the chosen CSRF protection implementation.
+
+    .. versionadded :: 1.8a1
+    """
+    registry = request.registry
+    csrf = registry.queryUtility(ICSRF)
+    if csrf is not None:
+        return csrf.get_csrf_token(request)
```
tseaver: Shouldn't this helper raise if no utility is registered?

pyramid/csrf.py
```diff
@@ -0,0 +1,225 @@
...
+    """ Generate a new CSRF token for the request passed and persist it in an
+    implementation defined manner. This calls the equivalent method in the
+    chosen CSRF protection implementation.
+
+    .. versionadded :: 1.8a1
+    """
+    registry = request.registry
+    csrf = registry.queryUtility(ICSRF)
+    if csrf is not None:
+        return csrf.new_csrf_token(request)
```
tseaver: Shouldn't this helper raise if no utility is registered?

pyramid/csrf.py
```diff
@@ -0,0 +1,225 @@
...
+    # encoded data over anything but a request.POST.
+    if token is not None:
+        supplied_token = request.POST.get(token, "")
+
+    # If we were unable to locate a CSRF token in a request body, then we'll
+    # check to see if there are any headers that have a value for us.
+    if supplied_token == "" and header is not None:
+        supplied_token = request.headers.get(header, "")
+
+    impl = request.registry.getUtility(ICSRF)
```
tseaver: Shouldn't this helper raise if no utility is registered?

pyramid/csrf.py
MatthewWilkes: I'm not sure, I think you can make arguments either way. If you've written third part code that's intended to be reusable it means you have to special case when users haven't got CSRF enabled.

Either way, we should document the behavior.

docs/narr/security.rst
MatthewWilkes: I believe @jcerjak is looking at rewording this, is that the case?

docs/narr/security.rst
MatthewWilkes: Agreed, the same is true of the current documentation, I didn't pick up on it when I changed `request.session.get_csrf_token` to `get_csrf_token`.

docs/narr/security.rst
jcerjak: I've fixed it in the code docstrings, but missed it here, fixing.

pyramid/csrf.py
MatthewWilkes: As this one is `getUtility` it will raise a `ComponentLookupError` if it's missing, were you thinking of something a bit more descriptive?

tseaver: Ah, my eyes read that as `queryUtility()` like the ones above.

docs/whatsnew-1.8.rst
```diff
@@ -0,0 +1,40 @@
+What's New in Pyramid 1.8
```
mmerickel: Please don't add a whatsnew document. We will do it prior to release. Your changes are good I just think it's easier to create that file from scratch just before a major release.

pyramid/csrf.py
mmerickel: > If third-party code registers a view with require_csrf, AFAIK there is no way straightforward way to disable CSRF protection.

This is intentional... if someone defines a view that explicitly requires csrf then it will be inaccessible until someone registers a csrf policy for the app. Turning off csrf protection for those views is a non-goal and can already be done just by re-defining it after `config.commit()` to override the old view.

I'd expect this code to raise if there is no policy. Simply using `getUtility` should be enough.

pyramid/csrf.py
mmerickel: Yes, it should raise. Using `getUtility` should be enough.

pyramid/csrf.py
```diff
@@ -0,0 +1,226 @@
...
+    """ Get the currently active CSRF token for the request passed, generating
+    a new one using ``new_csrf_token(request)`` if one does not exist. This
+    calls the equivalent method in the chosen CSRF protection implementation.
+
+    .. versionadded :: 1.8a1
+    """
+    registry = request.registry
+    csrf = registry.queryUtility(ICSRF)
+    if csrf is not None:
+        return csrf.get_csrf_token(request)
```
mmerickel: Should these be functions or should we add them to the request as methods, similar to `request.has_permission`, `request.session.get_csrf_token` etc? `request.get_csrf_token()`? `request.csrf.get_token()`?

bertjwregeer:
The CSRF cookie doesn't need to be signed.

pyramid/csrf.py
```diff
@@ -0,0 +1,226 @@
...
+    .. versionchanged:: 1.8a1
+       Moved from pyramid.session to pyramid.csrf
+    """
+    supplied_token = ""
+    # If this is a POST/PUT/etc request, then we'll check the body to see if it
+    # has a token. We explicitly use request.POST here because CSRF tokens
+    # should never appear in an URL as doing so is a security issue. We also
+    # explicitly check for request.POST here as we do not support sending form
+    # encoded data over anything but a request.POST.
+    if token is not None:
```
bertjwregeer: I would recommend swapping the check for the token in the headers/body. `request.POST` is more expensive than checking `request.headers`.

docs/whatsnew-1.8.rst
MatthewWilkes: Got it, thanks!

MatthewWilkes:
@mmerickel I've pondered a bit about the change to be specifically about storage, and think that it might be worthwhile to have some of the checking in the CSRF implementation too. Specifically, currently it assumes that constant-time string comparison is sufficient, but an implementation might want to do funky stuff like scope tokens (as mentioned in the original ticket).

If we have a hook there then we can allow this to be supported without having to also supply a `callback`.

pyramid/csrf.py
MatthewWilkes: I think that `new_csrf_token` and `get_csrf_token` should be available in the same place, but I'm a little nervous about having `new_csrf_token` available in templates, personally. Maybe I'm worrying over nothing, but it feels to me like something that should only be called in Python code.

pyramid/csrf.py
MatthewWilkes: This is unchanged from the previous implementation, but that makes sense. Switching.

MatthewWilkes:
@mmerickel @jcerjak and I worked on this more today and incorporated changes based on your feedback. We have added the checking function I referenced in my previous comment, as I think it gives us a bit more flexibility.

pyramid/csrf.py
bertjwregeer: Thanks @MatthewWilkes!

pyramid/csrf.py
```diff
@@ -0,0 +1,285 @@
...
+
+def csrf_token_template_global(event):
+    request = event.get('request', None)
+    try:
+        registry = request.registry
+    except AttributeError:
+        return
+    else:
+        csrf = registry.getUtility(ICSRFPolicy)
+        event['get_csrf_token'] = partial(csrf.get_csrf_token, request)
```
mmerickel: This is code from my own apps that I'll at least paste here incase you like it better.

```python
def includeme(config):
    # other session setup usually
    # ...
    config.add_subscriber(renderer_globals, 'pyramid.interfaces.IBeforeRender')

class CSRFToken(object):
    def __init__(self, request):
        self.request = request

    def __json__(self, request):
        return self.request.session.get_csrf_token()

    def __html__(self):
        return self.request.session.get_csrf_token()

def renderer_globals(event):
    request = event['request']
    event['csrf_token'] = CSRFToken(request)  # allows {{ csrf_token }}
```

bertjwregeer: :+1: I like this idea!

MatthewWilkes:
Will reply to the rest tomorrow, but if you can give us steer on naming that would help. Do you prefer ICSRFStorage?

mmerickel:
> Will reply to the rest tomorrow, but if you can give us steer on naming that would help. Do you prefer ICSRFStorage?

Yeah sorry, I know naming is the biggest derailment. Let's go with `ICSRFStoragePolicy` to disambiguate it from the rest of the csrf settings.

MatthewWilkes:
@mmerickel With regards to scoped tokens, this isn't included in core (and I personally don't think it should be), but it's nice and easy to implement in userspace with these changes. The implementation would be something like:

```python
from functools import wraps
import hmac

from zope.interface import implementer

from pyramid.csrf import ICSRFStoragePolicy, SessionCSRF
from pyramid.compat import bytes_

def csrf_scope(scope):
    def csrf_scope_decorator(func):
        @wraps(func)
        def with_scope(request, *args, **kwargs):
            request.csrf_scope = scope
            return func(request, *args, **kwargs)
        return with_scope
    return csrf_scope_decorator

@implementer(ICSRFStoragePolicy)
class ScopedSessionCSRF(SessionCSRF):

    def _get_scope(self, request):
        return getattr(request, 'csrf_scope', '')

    def _scope_token(self, scope, base_token):
        token = hmac.new(bytes_(base_token, 'ascii'), scope).hexdigest()
        print "Scoping {} with {} = {}".format(base_token, scope, token)
        return token

    def new_csrf_token(self, request):
        base_token = super(ScopedSessionCSRF, self).new_csrf_token(request)
        return self._scope_token(self._get_scope(request), base_token)

    def get_csrf_token(self, request):
        base_token = super(ScopedSessionCSRF, self).get_csrf_token(request)
        return self._scope_token(self._get_scope(request), base_token)
```

and would be used like:

```python
@view_config(route_name='home', renderer='templates/mytemplate.pt')
@csrf_scope('foobar')
def my_view(request):
    return {'project': 'hacking'}
```

MatthewWilkes:
@mmerickel WRT user facing changes, that's a good point. I've added a fallback to the `check_csrf_token` method that ensures that if a user does not configure CSRF the old behaviour of assuming the session will manage it is preserved.

That means that if the user explicitly marks a view as being CSRF checked without calling `config.set_default_csrf_options(require_csrf=True)` they will get SessionCSRF if there is a session present or an error about no session being configured if not.

pyramid/csrf.py
jcerjak: I like the simplicity, but maybe it's better to be more explicit here. Don't have a strong opinion about it, but I think it would be better if it's a separate PR.

docs/narr/security.rst
```diff
@@ -765,3 +765,210 @@ which would allow the attacker to control the content of the payload.  Re-using
...
+:meth:`pyramid.config.Configurator.set_default_csrf_options` directive.
+
+For example:
+
+.. code-block:: python
+
+   from pyramid.config import Configurator
+
+   config = Configurator()
+   config.set_default_csrf_options(implementation=MyCustomCSRFPolicy())
```
mmerickel: I'd prefer if this were a separate method for `config.set_csrf_storage_policy` as CSRF is entirely unusable without it and it would be good if it could be replaced without changing the other options.

pyramid/csrf.py
```diff
@@ -0,0 +1,291 @@
...
+from pyramid.interfaces import ICSRFStoragePolicy
+from pyramid.settings import aslist
+from pyramid.util import (
+    is_same_domain,
+    strings_differ
+)
+
+
+@implementer(ICSRFStoragePolicy)
+class SessionCSRF(object):
```
mmerickel: `SessionCSRFStoragePolicy`

pyramid/csrf.py
```diff
@@ -0,0 +1,291 @@
...
+        """ Returns True if supplied_token is the same value as get_csrf_token
+        returns for this request. """
+        expected = self.get_csrf_token(request)
+        return not strings_differ(
+            bytes_(expected, 'ascii'),
+            bytes_(supplied_token, 'ascii'),
+        )
+
+@implementer(ICSRFStoragePolicy)
+class CookieCSRF(object):
```
mmerickel: `CookieCSRFStoragePolicy`

pyramid/config/security.py
```diff
@@ -218,9 +231,12 @@ def register():
         intr['header'] = header
         intr['safe_methods'] = as_sorted_tuple(safe_methods)
         intr['callback'] = callback
+
+        self.add_subscriber(csrf_token_template_global, [BeforeRender])
```
mmerickel: Currently Pyramid has no precedent for automatically registering a `BeforeRender` subscriber and I think we should keep it that way. If you really want to inject it we could a) do it in the current renderer bindings (pyramid_jinja2, pyramid_mako, pyramid_chameleon) or b) do it in the `pyramid.renderer.RendererHelper` as a `system` value.

pyramid/csrf.py
```diff
@@ -0,0 +1,291 @@
...
+    .. versionadded :: 1.8a1
+    """
+
+    def __init__(self, cookie_name='csrf_token', secure=False, httponly=False,
+                 domain=None, path='/'):
+        self.cookie_name = cookie_name
+        self.secure = secure
+        self.httponly = httponly
+        self.domain = domain
+        self.path = path
```
mmerickel: This implementation should be using the `webob.cookies.CookieProfile` from webob in order to set the cookies on potentially multiple domains similar to how `SignedCookieSessionFactory` and `AuthTktAuthenticationPolicy` operate.

MatthewWilkes:
@mmerickel I've looked at this again today, as I now have a client who wants CSRF protection without sessions. I've pushed changes to address your comments, let me know when you've had a chance to take a look?

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    application automatically against CSRF attacks.
+
+    .. versionadded:: 1.4a2
+
+    .. versionchanged:: 1.7a1
+       A CSRF token passed in the query string of the request is no longer
+       considered valid. It must be passed in either the request body or
+       a header.
+
+    .. versionchanged:: 1.8a1
```
bertjwregeer: We are going to need to change these to Pyramid 1.9

pyramid/config/security.py
```diff
@@ -218,9 +225,25 @@ def register():
...
+        Set the CSRF storage policy used by subsequent view registrations.
+
+        ``policy`` is a class that implements the
+        :meth:`pyramid.interfaces.ICSRFStoragePolicy` interface that will be used for all
+        CSRF functionality.
+        """
+        def register():
+            self.registry.registerUtility(policy, ICSRFStoragePolicy)
+
+        self.action(ICSRFStoragePolicy, register, order=PHASE1_CONFIG)
```
mmerickel: I'm pretty sure the `PHASE1_CONFIG` is unnecessary here. It's used in `set_default_csrf_options` because it needs to be configured prior to `add_view` running the `csrf_view` view deriver. I think the storage policy is only used at request-time though so can be left at the default config phase.

If you think it should be `PHASE1_CONFIG` then it should be added to http://docs.pylonsproject.org/projects/pyramid/en/master/narr/extconfig.html#ordering-actions

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    if supplied_token == "" and token is not None:
+        supplied_token = request.POST.get(token, "")
+
+    policy = request.registry.queryUtility(ICSRFStoragePolicy)
+    if policy is None:
+        # There is no policy set, but we are trying to validate a CSRF token
+        # This means explicit validation has been asked for without configuring
+        # the CSRF implementation. Fall back to SessionCSRFStoragePolicy as that is the
+        # default
+        policy = SessionCSRFStoragePolicy()
```
mmerickel: This is probably unnecessary now. Just use `getUtility` instead an exception if there's no policy. Thoughts?

pyramid/config/views.py
```diff
@@ -641,16 +641,14 @@ def wrapper(context, request):
...
+          the value of ``policy.get_csrf_token()`` (where ``policy`` is an
+          implementation of :meth:`pyramid.interfaces.ICSRFStoragePolicy`), and the
+          check will pass if these two values are the same. If the check
+          passes, the associated view will be permitted to execute. If the
           check fails, the associated view will not be permitted to execute.
 
-          Note that using this feature requires a :term:`session factory` to
-          have been configured.
-
           .. versionadded:: 1.4a2
```
stevepiercy: possibly add a `.. versionchanged 1.x` along with a brief description of the change. I'm not sure into which version these changes will be implemented. @bertjwregeer indicates 1.9 in another comment.

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+        """ Sets a new CSRF token into the session and returns it. """
+        return request.session.new_csrf_token()
+
+    def get_csrf_token(self, request):
+        """ Returns the currently active CSRF token from the session, generating
+        a new one if needed."""
+        return request.session.get_csrf_token()
+
+    def check_csrf_token(self, request, supplied_token):
+        """ Returns True if supplied_token is the same value as get_csrf_token
```
stevepiercy: """Returns ``True`` if ``supplied_token`` is the same value as ``get_csrf_token``

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+        return not strings_differ(
+            bytes_(expected, 'ascii'),
+            bytes_(supplied_token, 'ascii'),
+        )
+
+@implementer(ICSRFStoragePolicy)
+class CookieCSRFStoragePolicy(object):
+    """ An alternative CSRF implementation that stores its information in
+    unauthenticated cookies, known as the 'Double Submit Cookie' method in the
+    OWASP CSRF guidelines. This gives some additional flexibility with regards
```
stevepiercy: Please add link to OWASP resource.

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+        """ Returns the currently active CSRF token by checking the cookies
+        sent with the current request."""
+        bound_cookies = self.cookie_profile.bind(request)
+        token = bound_cookies.get_value()
+        if not token:
+            token = self.new_csrf_token(request)
+        return token
+
+    def check_csrf_token(self, request, supplied_token):
+        """ Returns True if supplied_token is the same value as get_csrf_token
```
stevepiercy: """Returns ``True`` if ``supplied_token`` is the same value as ``get_csrf_token``

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+
+    .. versionadded:: 1.4a2
+
+    .. versionchanged:: 1.7a1
+       A CSRF token passed in the query string of the request is no longer
+       considered valid. It must be passed in either the request body or
+       a header.
+
+    .. versionchanged:: 1.8a1
+       Moved from pyramid.session to pyramid.csrf
```
stevepiercy: Moved from :mod:`pyramid.session` to :mod:`pyramid.csrf`.

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    if not policy.check_csrf_token(request, supplied_token):
+        if raises:
+            raise BadCSRFToken('check_csrf_token(): Invalid token')
+        return False
+    return True
+
+
+def check_csrf_origin(request, trusted_origins=None, raises=True):
+    """
+    Check the Origin of the request to see if it is a cross site request or
```
stevepiercy: Not sure why "Origin" is capitalized. Should it be lowercased or have reStructuredText syntax applied?

Also "cross-site".

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    not.
+
+    If the value supplied by the Origin or Referer header isn't one of the
```
stevepiercy: ibid. So perhaps ``Origin`` and ``Referer`` (also ``Referrer``). Although the former misspelling seems to be inherited, I think the correct spelling is valid as well.

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    trusted origins and ``raises`` is ``True``, this function will raise a
+    :exc:`pyramid.exceptions.BadCSRFOrigin` exception but if ``raises`` is
```
stevepiercy: exception, but

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    ``False`` this function will return ``False`` instead. If the CSRF origin
```
stevepiercy: ``False``, this

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    checks are successful this function will return ``True`` unconditionally.
+
+    Additional trusted origins may be added by passing a list of domain (and
+    ports if nonstandard like `['example.com', 'dev.example.com:8080']`) in
```
stevepiercy: Let's standardize our "non-standard"s.

Double up the ticks:

    ``['example.com', 'dev.example.com:8080']``

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+    with the ``trusted_origins`` parameter. If ``trusted_origins`` is ``None``
+    (the default) this list of additional domains will be pulled from the
+    ``pyramid.csrf_trusted_origins`` setting.
+
+    Note that this function will do nothing if request.scheme is not https.
```
stevepiercy: ``request.scheme`` is not ``https``.

pyramid/csrf.py
```diff
@@ -0,0 +1,283 @@
...
+
+    .. versionadded:: 1.7
+
+    .. versionchanged:: 1.8a1
+       Moved from pyramid.session to pyramid.csrf
```
stevepiercy: Moved from :mod:`pyramid.session` to :mod:`pyramid.csrf`.

pyramid/interfaces.py
```diff
@@ -981,19 +981,30 @@ def peek_flash(queue=''):
...
         ``new_csrf_token``, that token will be returned.  If no CSRF token
-        was previously set into the session, ``new_csrf_token`` will be
-        called, which will create and set a token, and this token will be
-        returned.
+        was previously set, ``new_csrf_token`` will be called, which will 
+        create and set a token, and this token will be returned.
         """
 
+    def check_csrf_token(request, supplied_token):
+        """ Returns a boolean that represents if supplied_token is a valid CSRF
```
stevepiercy: ``supplied_token``

pyramid/interfaces.py
```diff
@@ -981,19 +981,30 @@ def peek_flash(queue=''):
...
+        token for this request. Comparing strings for equality must be done
+        using :func:`pyramid.utils.strings_differ` to avoid timing attacks.
+        """
+
+
 class IIntrospector(Interface):
     def get(category_name, discriminator, default=None):
         """ Get the IIntrospectable related to the category_name and the
```
stevepiercy: """ Get the ``IIntrospectable`` related to the ``category_name`` and the

pyramid/csrf.py
bertjwregeer: `Origin` is the HTTP Header. All HTTP headers start with an uppercase.

pyramid/csrf.py
bertjwregeer: `Referer` being misspelled is unfortunately in the spec. You will not find that `Referrer` anywhere because it is not valid for an HTTP header.

pyramid/csrf.py
stevepiercy: OK, in that case, it needs some markup:

    ``Origin``

pyramid/csrf.py
stevepiercy: ``Origin`` or ``Referer``

pyramid/interfaces.py
mmerickel: @stevepiercy You're commenting on a lot of things that he didn't write in this code. I'm not sure what the right answer is here but I'd rather have him fix the parts he actually wrote and not ask him to do spring cleaning on pre-existing things to get this feature complete.

stevepiercy: I can circle back afterward to do spring cleaning if it doesn't get changed in this round.

mmerickel:
Just FYI I want to get an alpha of pyramid out soon and this is close enough that I'd rather it didn't miss the cut again. @MatthewWilkes if you don't have time to wrap this up I'll probably do it myself this weekend.

MatthewWilkes:
@mmerickel I'll address these issues tomorrow during the work day, thanks for the prod.

pyramid/config/security.py
MatthewWilkes: I had added it to extconfig, but you're right. Removing it from here and extconfig.

pyramid/csrf.py
MatthewWilkes: Yep, thanks. I was missing a call in the testing setup that this was hiding. Fixed.

CHANGES.txt
```diff
@@ -24,6 +24,14 @@ Features
   can be alleviated by invoking ``config.begin()`` and ``config.end()``
   appropriately. See https://github.com/Pylons/pyramid/pull/2989
 
+- A new CSRF implementation, :class:`pyramid.csrf.SessionCSRF` has been added,
+  which deleagates all CSRF generation to the current session, following the
```
MatthewWilkes: :+1:

docs/narr/security.rst
```diff
@@ -765,3 +765,210 @@ which would allow the attacker to control the content of the payload.  Re-using
...
+unless otherwise overridden by ``set_default_csrf_options``. The token is
+checked against the value in ``request.POST`` which is the submitted form body.
+If this value is not present, then the header will be checked.
+
+In addition to token based CSRF checks, if the request is using HTTPS then the
+automatic CSRF checking will also check the referrer of the request to ensure
+that it matches one of the trusted origins. By default the only trusted origin
+is the current host, however additional origins may be configured by setting
+``pyramid.csrf_trusted_origins`` to a list of domain names (and ports if they
+are non standard). If a host in the list of domains starts with a ``.`` then
```
MatthewWilkes: Not done as it's old text that's moved, and I don't want to change the corresponding text in the changelog as part of this.

pyramid/config/views.py
MatthewWilkes: :+1: 

The change is based on the removed text, and adds a glossary term for the CSRF storage policy in the explanation.

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/csrf.py
MatthewWilkes: :+1:

pyramid/interfaces.py
MatthewWilkes: :+1:

pyramid/interfaces.py
MatthewWilkes: I really don't mind spring cleaning lines I've touched, even if they're just moving. I'm applying almost all of @stevepiercy's changed (and, yes, I did see this comment before I started applying them!)

MatthewWilkes:
@mmerickel I've pushed changes to the things brought up in this thread, with comments against each either confirming I've made the change (:+1:) or explaining why I haven't.

docs/narr/sessions.rst
```diff
@@ -320,179 +319,3 @@ flash storage.
 .. index::
```
MatthewWilkes: :+1:

bertjwregeer:
Thanks @MatthewWilkes!

pyramid/interfaces.py
stevepiercy: :hugs:

pyramid/csrf.py
```diff
@@ -96,8 +96,8 @@ def get_csrf_token(self, request):
         return token
 
     def check_csrf_token(self, request, supplied_token):
-        """ Returns True if supplied_token is the same value as get_csrf_token
-        returns for this request. """
+        """ Returns ``True`` if ``supplied_token is`` the same value as
```
stevepiercy: backticks goof:

``supplied_token is``
s/b
``supplied_token`` is

stevepiercy:
Thanks @MatthewWilkes. Good stuff. I found only one typo, and then it's good to go from my perspective.

mmerickel:
This PR will closed and show as merged automatically when #3019 is merged.

MatthewWilkes:
@mmerickel Please don't remove the check_csrf_token method, it is very useful to have it pluggable as that allows for keyring based or fully-stateless CSRF policies.

For example, consider this CSRF policy:

    implementer(ICSRFStoragePolicy)
    class HMACCSRFStoragePolicy(object):
        """ An alternative CSRF implementation that uses short-lived HMAC keyed on
        the user.
        """

        def __init__(self, key):
            self.key = key

        def new_csrf_token(self, request):
            user = request.authenticated_userid
            generated = int(time.time())
            digest = hmac.new(self.key, 'csrf-{}-{}'.format(user, generated), hashlib.sha1).hexdigest()
            return '{}:{}:{}'.format(user, generated, digest)

        def get_csrf_token(self, request):
            return self.new_csrf_token(request)

        def check_csrf_token(self, request, supplied_token):
            user, generated, digest = supplied_token.split(':')
            if user != request.authenticated_userid:
                return False
            if int(generated) < (time.time() - 300):
                return False
            expected = hmac.new(self.key, 'csrf-{}-{}'.format(user, generated), hashlib.sha1).hexdigest()
            return hmac.compare_digest(expected, digest)
            

For non-security experts reading this, I'm not claiming that the above is a good CSRF implementation, however I do think that it demonstrates that there is a value in allowing the CSRF implementation to decide what's valid and what isn't.

If you're concerned about boilerplate (although, personally, I think it's minimal), maybe we could make a mixin available?

bertjwregeer:
@MatthewWilkes `get_csrf_token` has access to the `request` object. You will have to manually pull the existing token out of the appropriate header/POST data or wherever, but it is still possible to do what you want to do.

MatthewWilkes:
This is true, good point.

mmerickel:
@MatthewWilkes thank you very much for the alternative example. I definitely had not considered this stateless use case. It appears that the main downside of supporting that use case without a pluggable `check_csrf_token` is that the policy is now responsible for parsing the header/body itself to find the supplied token. This feels vaguely similar to `unauthenticated_userid` versus `authenticated_userid` in authentication policies. Perhaps I can add another API to get the "unauthenticated" csrf token which can be used if necessary. I'm thinking `find_csrf_token` or `get_unverified_csrf_token`.

MatthewWilkes:
Thanks for coming to that decision, it occured to me this morning that Bert's workaround has a fatal flaw. Any submission that includes a CSRF check would cause get_card_token to return the supplied token, making it very easy to hit the short timeout. Imagine a form that takes 3 minutes to fill in, once you've submitted that the resultant page render will use a token with only 2 minutes of validity, so a multipage form would quickly fail.

As to being the most overengineered in the business, I have one word for you, Plone.

PR #2859, opened 2016-12-08
Fix pluralization in default domain for non-germanic languages

MatthewWilkes:
This is a reimplementation of PR #2102 which is less invasive and tracks the bug to its original location.

When loading a new `.mo` file, Pyramid checks if the domain already exists, and if so it merges the new catalog into the old one. It does not update the `plural` function when this happens, as it's assumed that it is consistent inside a domain.

The default domain is a special case, as it contains the other domains in a dictionary, so it is created ahead of time with an empty catalog, and the default pluralisation rule from #235. As all other files in the `messages` domain will be merged to this catalog, the plural rule is never overwritten.

This change introduces a flag to state if the default plural from #235 has been overridden. When merging two translations it checks if the domain is the default domain, and if the flag isn't set it replaces the `plural` lambda expression with the gettext generated one and sets the flag.

### DIFF
--- a/pyramid/tests/test_i18n.py
+++ b/pyramid/tests/test_i18n.py
@@ -357,6 +357,36 @@ class TestTranslations(unittest.TestCase):
         inst.add(inst2)
         self.assertEqual(inst._catalog['a'], 'b')
 
+    def test_add_default_domain_replaces_plural_first_time(self):
+        # Create three empty message catalogs in the default domain
+        inst = self._getTargetClass()(None, domain='messages')
+        inst2 = self._getTargetClass()(None, domain='messages')
+        inst3 = self._getTargetClass()(None, domain='messages')
+        inst._catalog = {}
+        inst2._catalog = {}
+        inst3._catalog = {}
+
+        # The default plural scheme is the germanic one
+        self.assertEqual(inst.plural(0), 1)
+        self.assertEqual(inst.plural(1), 0)
+        self.assertEqual(inst.plural(2), 1)
+
+        # inst2 represents a message file that declares french plurals
+        inst2.plural = lambda n: n > 1
+        inst.add(inst2)
+        # that plural rule should now apply to inst
+        self.assertEqual(inst.plural(0), 0)
+        self.assertEqual(inst.plural(1), 0)
+        self.assertEqual(inst.plural(2), 1)
+
+        # We load a second message file with different plural rules
+        inst3.plural = lambda n: n > 0
+        inst.add(inst3)
+        # It doesn't override the previously loaded rule
+        self.assertEqual(inst.plural(0), 0)
+        self.assertEqual(inst.plural(1), 0)
+        self.assertEqual(inst.plural(2), 1)
+
     def test_dgettext(self):
         t = self._makeOne()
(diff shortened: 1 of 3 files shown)


pyramid/tests/test_i18n.py
```diff
@@ -357,6 +357,36 @@ def test_add_same_domain_merge_true(self):
...
+        self.assertEqual(inst.plural(1), 0)
+        self.assertEqual(inst.plural(2), 1)
+
+        # We load a second message file with different plural rules
+        inst3.plural = lambda n: n > 0
+        inst.add(inst3)
+        # It doesn't override the previously loaded rule
+        self.assertEqual(inst.plural(0), 0)
+        self.assertEqual(inst.plural(1), 0)
+        self.assertEqual(inst.plural(2), 1)
```
mmerickel: I don't have a lot of translation experience but is this really correct? So the system doesn't choose the correct plural rules based on the selected language?

MatthewWilkes: This is all in the context of a single language. 

The current behaviour is that the first plural wins, which for the default domain is the default pluraliser.

This change makes it so that for the default domain the second wins, which is the first user defined one.

There is an existing assumption that plurals within a language and domain pair are consistent which this patch doesn't affect

MatthewWilkes: I am happy with it personally, but I know that @tseaver had some concerns about the previous PR's approach, so maybe it would be good to have him confirm that this is sufficiently different?

mmerickel:
FWIW I might have defined a `DEFAULT_PLURAL = lambda n: int(n != 1)` and then done tests against it to see `if self.plural is DEFAULT_PLURAL` instead of introducing a second variable.

MatthewWilkes:
You mean like #2102 last year?

mmerickel:
Hmm, yes apparently... why did we abandon that pr again? :-) Is the main issue that the previous PR was doing things for all domains and this one focuses on the default domain? Sorry I'm trying to help review this but maybe just causing more problems.

MatthewWilkes:
I also didn't really understand it previously, but do more so now. I'm happy for you to go back to that PR and encourage a contribution now we have worked it out, if that is better.

mmerickel:
The fundamental difference between this PR and the other one is that this only changes the plural when adding to the `DEFAULT_DOMAIN` instead of preferring the `plural` from any translation that is added to any domain.

It appears as if this is done because the default domain is considered a fallback for all other domains and in general if a translation occurs on the non-default domain it's expected it comes with an appropriate plural function whereas the default domain currently always has the default plural function. So if someone is adding a new translation to the default domain, we should pull the plural function from that translation.

Please correct me if I'm wrong but with that in mind I prefer your location for the patch, as it's only done once at the top instead of twice. However I prefer the other patch for its approach of using only a single variable. Could you push that particular change and I'm happy to merge.

MatthewWilkes:
@mmerickel I believe that's accurate. I have pushed the change to using the lambda function identity to check for overrides.

tseaver:
@mmerickel Feel free.

mmerickel:
Thank you @MatthewWilkes and @quantum-omega!

PR #2866, opened 2016-12-13
pcreate to cookiecutter

stevepiercy:
refs: #2798, #2384


pyramid/scripts/pcreate.py
```diff
@@ -224,5 +231,12 @@ def confirm_bad_name(self, prompt): # pragma: no cover
         answer = input_('{0} [y|N]: '.format(prompt))
         return answer.strip().lower() == 'y'
 
+    def _warn_pcreate_deprecated(self):
+        self.out('''\
+Note: As of Pyramid 1.8, this command is deprecated. Use a specific
+cookiecutter instead:
+https://github.com/pylons/?query=cookiecutter
```
mmerickel: https://github.com/Pylons?q=cookiecutter

stevepiercy: Done in 4393f8b

mmerickel: Did you try the url? That wasn't the only change...

mmerickel:
This PR takes a different approach to the jinja2 conversion than what I did in the pyramid-cookiecutter-starter. It should be converted to use the `layout.jinja2` similar to what's happening here. Note that it's not a straight copy as the syntax will change to fit a cookiecutter template.

PR #2870, opened 2016-12-15
fix Travis docs build

stevepiercy:
(cherry picked from commit 5700b99)


stevepiercy:
The build fails on pypy3, but we don't test 1.7 or master on pypy3. I'm merging this PR and opened an issue #2871.

PR #2873, opened 2016-12-16
defer config.add_translation_dirs

### LINKED ISSUE #2046: Document interaction of .override_asset() with .add_translation_dirs() and others
It looks like `.override_asset()` is not effective in the subsequent `.add_translation_dirs()` unless `.commit()` is called between these two calls. However I can't find this stated neither at http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#assets-chapter nor at http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.override_asset
I think interaction of `.override_asset()` with other directives should be described in documentation.

mmerickel:
This resolves #2046.

1. Added docs about importance of `config.begin()` when invoking `add_translation_dirs`.

2. Deferred the computation of the specs until the default config phase in order to allow asset overrides defined later to have an effect.

3. Used the `AssetResolver` to resolve the entire spec instead of just the package name such that translation directories can override each other even if they don't have the same name / subpath.

4. `config.override_asset` now occurs in an earlier config phase to ensure assets are updated prior to `add_translation_dirs` executing.

I would appreciate it if @piotr-dobrogost could review this.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -140,6 +140,29 @@ Features
   ``--help`` output as well as enabling nicer documentation of their options.
   See https://github.com/Pylons/pyramid/pull/2864
 
+- Any deferred configuration action registered via ``config.action`` may now
+  depend on threadlocal state, such as asset overrides, being active when
+  the action is executed.
+  See https://github.com/Pylons/pyramid/pull/2873
+
+- Asset specifications for directories passed to
+  ``config.add_translation_dirs`` now support overriding the entire asset
+  specification, including the folder name. Previously only the package name
+  was supported and the folder would always need to have the same name.
+  See https://github.com/Pylons/pyramid/pull/2873
+
+- ``config.begin()`` will propagate the current threadlocal request through
+  as long as the registry is the same. For example:
+
+  .. code-block:: python
+
+     request = Request.blank(...)
+     config.begin(request)  # pushes a request
+     config.begin()         # propagates the previous request through unchanged
+     assert get_current_request() is request
+
+  See https://github.com/Pylons/pyramid/pull/2873
+
 Bug Fixes
 ---------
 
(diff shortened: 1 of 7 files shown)


mmerickel:
I will amend this PR tonight to have `add_translation_dirs` call `begin` and `end` itself during the register cycle allowing it to do what it needs to do properly. This will remove the new docs added.

mmerickel:
Ok I upped the ante and am now calling begin/end during every commit cycle so **any** deferred action can depend on threadlocals. I think this makes a lot of sense.

This is ready for review.

mmerickel:
I'm going to merge this tonight unless someone reviews it by then and has an issue.

bertjwregeer:
Will review tonight.

CHANGES.txt
```diff
@@ -140,6 +140,29 @@ Features
...
   See https://github.com/Pylons/pyramid/pull/2864
 
+- Any deferred configuration action registered via ``config.action`` may now
+  depend on threadlocal state, such as asset overrides, being active when
+  the action is executed.
+  See https://github.com/Pylons/pyramid/pull/2873
+
+- Asset specifications for directories passed to
+  ``config.add_translation_dirs`` now support overriding the entire asset
+  specification, including the folder name. Previously on the package name
```
bertjwregeer: s/on/only/

pyramid/config/i18n.py
```diff
@@ -69,32 +66,32 @@ def add_translation_dirs(self, *specs):
...
-            if not os.path.isdir(os.path.realpath(directory)):
-                raise ConfigurationError('"%s" is not a directory' %
-                                         directory)
-            intr = self.introspectable('translation directories', directory,
-                                       spec, 'translation directory')
-            intr['directory'] = directory
-            intr['spec'] = spec
-            introspectables.append(intr)
-            directories.append(directory)
+        resolver = AssetResolver(self.package_name)
```
bertjwregeer: Nice cleanup!

CHANGES.txt
mmerickel: fixed, thanks

piotr-dobrogost:
Looks good. Thank you for fixing this.

PR #2874, opened 2016-12-16
turn the Configurator into a context manager

### LINKED ISSUE #2872: make the configurator a context manager
There's some subtle bugs in apps that hit certain APIs during configuration because parts of pyramid rely on threadlocal state that may or may not be available.

An example:

```python
config = Configurator()
config.override_asset(to_override='myapp:locale/', override_with='yourapp:locale/')
config.add_translation_dirs('myapp:locale')
```

The result here is that the translations are still using `myapp:locale` because `add_translation_dirs` resolves those asset paths at config-time. Asset overrides work by using custom `pkg_resources` hooks that only affect `pkg_resources` when a registry is pushed on the threadlocal stack. Normally this would require surrounding relevant code with `config.begin()` and `config.end()`. We've done a poor job of documenting this.

I propose an optional context manager that we can implement easily (we already did something like this for `p.paster.bootstrap` in https://github.com/Pylons/pyramid/pull/2760) in order to manage the lifecycle.

```
with Configurator() as config:
    config.override_asset(to_override='myapp:locale/', override_with='yourapp:locale/')
    config.add_translation_dirs('myapp:locale')
```

related: https://github.com/Pylons/pyramid/pull/2760 https://github.com/Pylons/pyramid/issues/2046

mmerickel:
The main goal here is to provide a common pattern for calling `config.begin()` and `config.end()` such that the registry, settings, asset overrides, etc are all properly available at config-time. It's rare right now for apps to be invoking `config.begin()` and it would be good to improve on this situation.

Unfortunately the context manager is not a silver bullet. For example the `config` object in the below example still has a strong ref to it outside the scope of the `with`-statement, but at least its intended lifetime is more clear perhaps.

Note the new hello world example:

```python

def hello_world(request):
    return Response('Hello %(name)s!' % request.matchdict)

if __name__ == '__main__':
    with Configurator() as config:
        config.add_route('hello', '/hello/{name}')
        config.add_view(hello_world, route_name='hello')
        app = config.make_wsgi_app()
    server.serve_forever()
```

fixes #2872

### DIFF
--- a/pyramid/config/__init__.py
+++ b/pyramid/config/__init__.py
@@ -110,6 +110,17 @@ class Configurator(
     A Configurator is used to configure a :app:`Pyramid`
     :term:`application registry`.
 
+    The Configurator lifecycle can be managed by using a context manager to
+    automatically handle calling :meth:`pyramid.config.Configurator.begin` and
+    :meth:`pyramid.config.Configurator.end` as well as
+    :meth:`pyramid.config.Configurator.commit`.
+
+    .. code-block:: python
+
+        with Configurator(settings=settings) as config:
+            config.add_route('home', '/')
+            app = config.make_wsgi_app()
+
     If the ``registry`` argument is not ``None``, it must
     be an instance of the :class:`pyramid.registry.Registry` class
     representing the registry to configure.  If ``registry`` is ``None``, the
@@ -265,6 +276,11 @@ class Configurator(
     .. versionadded:: 1.6
        The ``root_package`` argument.
        The ``response_factory`` argument.
+
+    .. versionadded:: 1.9
+       The ability to use the configurator as a context manager with the
+       ``with``-statement to make threadlocal configuration available for
+       further configuration with an implicit commit.
     """
     manager = manager # for testing injection
     venusian = venusian # for testing injection
@@ -646,12 +662,22 @@ class Configurator(
     _ctx = action_state # bw compat
 
     def commit(self):
-        """ Commit any pending configuration actions. If a configuration
+        """
+        Commit any pending configuration actions. If a configuration
         conflict is detected in the pending configuration actions, this method
         will raise a :exc:`ConfigurationConflictError`; within the traceback
         of this error will be information about the source of the conflict,
         usually including file names and line numbers of the cause of the
-        configuration conflicts."""
+        configuration conflicts.
+
+        .. warning::
+           You should think very carefully before manually invoking
+           ``commit()``. Especially not as part of any reusable configuration
+           methods. Normally it should only be done by an application author at
+           the end of configuration in order to override certain aspects of an
+           addon.
+
+        """
         self.begin()
         try:
             self.action_state.execute_actions(introspector=self.introspector)
@@ -932,6 +958,16 @@ class Configurator(
         value.
         """
         return self.manager.pop()
+
+    def __enter__(self):
+        self.begin()
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        self.end()
+
+        if exc_value is None:
+            self.commit()
 
     # this is *not* an action method (uses caller_package)
     def scan(self, package=None, categories=None, onerror=None, ignore=None,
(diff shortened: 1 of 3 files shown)


pyramid/config/__init__.py
```diff
@@ -903,6 +921,13 @@ def end(self):
         """
         return self.manager.pop()
 
+    def __enter__(self):
+        self.begin()
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        self.end()
```
tseaver: Do we need to do any clean-up, or mark the configuration instance as b0rked, if the `with` block exits with an error?

mmerickel: The error will just propagate up since we aren't returning `True` to squash it. I can add a test to clarify but that's the context manager contract.

mmerickel:
One consideration is adding an implicit `config.commit()` in the exit.

pyramid/config/__init__.py
tseaver: What about:

```python
    def __exit__(self, exc_type, exc_value, exc_traceback):
        if exc_type is not None:
            # self.abort()  # we don't have an 'abort()' method.
            self.action_state = ActionState()
        else:
            self.commit()
            self.end()
```

mmerickel: What do you want to signify by clearing the action state? Is it telling the user that there's a way to try stuff out and rollback?

I like the idea. Practically I think it's dangerous to claim we can do anything about rollbacks. We generally expect any exceptions at config-time to kill your app and do not document error handling for any directives. Also much of the third party extensions I've ever seen do all their work in the directive without deferring it to the commit cycle.

tseaver: > What do you want to signify by clearing the action state?

Just that any work set up to be done within the scope of the `with` block is cancelled.

> Also much of the third party extensions I've ever seen do all their work in the directive without deferring it to the commit cycle.

https://youtu.be/QDvQ77JP8nw?t=50

mmerickel: Well played sir. Well played.

> Just that any work set up to be done within the scope of the with block is cancelled.

This is reasonable to me. I do wish we had better examples in the docs for creating custom directives, but we can consider that separately. Want me to make this change using your example?

tseaver: Is there any reason *not* to invoke `commit()` during a successful exit?

mmerickel: I've thought a bit more about this and I think adding a `config.commit()` is the right answer here but it would probably be nice if we can avoid doing a commit if it's embedded in another context manager. @tseaver Good idea? Bad idea?

tseaver: I can't think of any sane way to detect that we are inside another context manager.  Maybe we could add an opt-out parameter to the `Configurator`?  E.g.:

```python

if __name__ == '__main__':
    with Configurator() as outer:

        # do stuff with outer

        with Configurator(commit_on_exit=False) as inner:
            inner.add_route('hello', '/hello/{name}')
            inner.add_view(hello_world, route_name='hello')

        # do more stuff with outer

    app = outer.make_wsgi_app()
    server.serve_forever()

mmerickel: I think it could be done similar to how we handle ``config.route_prefix`` and a few other state variables where we shuffle them through each new configurator. We could actually use that state to warn a user if they try to do a manual commit as well.

mmerickel:
@tseaver I've updated this per your comments to perform a commit. I decided to avoid any complex nesting logic and simply added a warning to the ``config.commit()`` docs.

pyramid/config/__init__.py
bertjwregeer: Were the comments here addressed? Or did we decide not to move forward with some of the comments here?

mmerickel: I decided that tracking and avoiding nested commits wasn't our problem and was more trouble than it's worth.

The only other issue was clearing the action state which we do not do on normal commit failures nor do we have an api to do it in general and as such I won't be adding it to this type of commit either until we decide to support it in a standardized way across all commit mechanisms.

mmerickel:
Thanks @tseaver and @bertjwregeer .

PR #2875, opened 2016-12-18
Fix typo

PavloKapyshin:


tseaver:
@PavloKapyshin Thanks for the patch!

Issue #2871, opened 2016-12-15
Travis fails to build Pyramid 1.6 on pypy3


stevepiercy:
Travis fails to build Pyramid 1.6 on pypy3:
https://travis-ci.org/Pylons/pyramid/jobs/184200134

Suggest removing pypy3 from 1.6 matrix.

mmerickel:
Yes, that's fine. Drop it from tox as well.

PR #2877, opened 2016-12-20
remove pypy3 from the testing matrix and docs

stevepiercy:
refs #2871

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ matrix:
           env: TOXENV=py35
         - python: pypy
           env: TOXENV=pypy
-        - python: pypy3
-          env: TOXENV=pypy3
         - python: 3.5
           env: TOXENV=py2-cover,py3-cover,coverage
         - python: 3.5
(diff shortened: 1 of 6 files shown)


stevepiercy:
Tests for py26 are failing now. Perhaps pin webtest to 2.0.23 for pyramid 1.6?
https://github.com/Pylons/webtest/blob/master/CHANGELOG.rst

mmerickel:
It only really needs to be dropped from the travis test matrix. It runs fine on pypy3... just not the version on travis.

As far as webtest I would pin waitress itself to < 1.0 instead.

bertjwregeer:
Ah, looks like webtest now requires Python 2.7 or higher, waitress shouldn't have broken Py2.6 compat, it's just no longer officially supported.

mmerickel:
> Ah, looks like webtest now requires Python 2.7 or higher, waitress shouldn't have broken Py2.6 compat, it's just no longer officially supported.

Yes, I saw that but if you read the webtest changelog it says "newer versions may still work if you pin waitress < 1.0" so I went for the quick fix. However, webtest stopped installing ordereddict on py26 which broke things as well so the changelog was a bit deceiving / not quite clear on how much you had to do to make it work on 2.6.

Issue #2882, opened 2016-12-24
wiki2 tests on master have import errors


stevepiercy:
Ref: https://github.com/Pylons/pyramid/pull/2855#issuecomment-268483317

On the `master` branch, after running through the wiki2 tutorial from a cookiecutter or scaffold, then running the tests in the Adding Tests step, there are import errors. To save time, one could follow these steps to reproduce:

* http://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/wiki2/installation.html
* overwrote my generated tutorial files with those from https://github.com/Pylons/pyramid/tree/master/docs/tutorials/wiki2/src/tests
* ran `pip install -e ".[testing]" to install dependencies
* then pickup from my above transcript, starting with `$VENV/bin/py.test -q`

I did not install the `mock` package (it was not included in instructions or setup.py), so I had to change the imports.

Here's the traceback.

```python
$VENV/bin/pytest -q

========================================================================== ERRORS ==========================================================================
______________________________________________________ ERROR collecting tutorial/tests/test_initdb.py ______________________________________________________
ImportError while importing test module '/Users/stevepiercy/projects/__pcs/tutorial/tutorial/tests/test_initdb.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tutorial/tests/test_initdb.py:1: in <module>
    import mock
E   ImportError: No module named 'mock'
_____________________________________________________ ERROR collecting tutorial/tests/test_security.py _____________________________________________________
ImportError while importing test module '/Users/stevepiercy/projects/__pcs/tutorial/tutorial/tests/test_security.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tutorial/tests/test_security.py:1: in <module>
    import mock
E   ImportError: No module named 'mock'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
2 error in 1.01 seconds
```

mmerickel:
The issue is that as @zupo said, python 3 has `unittest.mock` but python 2 does not and requires a dependency.

[...]

PR #2855, opened 2016-12-07
bring test coverage of wiki2 up to 100%

zupo:

PR #2883, opened 2016-12-24
docs cookiecutter changes only - wiki2

stevepiercy:
This set includes the SQLAlchemy Tutorial


stevepiercy:
Note that there are some failing tests in the tutorial that came from master. See #2882 and #2855.

PR #2889, opened 2016-12-27
Docs cookiecutter changes only - Quick Tutorial and all other files

stevepiercy:
ref: #2384. I rewrote tutorial_approach twice because it was klunky and inaccurate. There were a few other minor corrections as well. Otherwise this is all that was needed for Quick Tutorial.


docs/glossary.rst
```diff
@@ -1063,7 +1063,7 @@ Glossary
 
    Waitress
       A :term:`WSGI` server that runs on UNIX and Windows under Python 2.6+
-      and Python 3.2+.  Projects generated via Pyramid scaffolding use
+      and Python 3.2+.  Projects generated via Pyramid cookiecutters use
```
mmerickel: 3.3+

stevepiercy: Heh, looks like the deprecations include 2.6 as well as 3.2, but the narrative docs for Waitress still say 2.6 and 3.2. I'll take care of that bit of housekeeping in that repo.

PR #2895, opened 2017-01-03
Adding pip install -e .

jhb:
The pip install was missing, leading to the files from authentication still being used.


mmerickel:
Isn't this instruction above along with `cp -r authentication authorization` on line 45?

jhb:
oh, so sorry, you are right of course. /me is too tired after reading through lots of chapters...

mmerickel:
Great, of course don't let that stop you from opening more issues. It's no trouble!

PR #2896, opened 2017-01-08
Update project.rst

ramgoer:
incorrect cookietcutter path for windows environment
%VENV%\Scripts\cookiecutter


stevepiercy:
This change seems reasonable, but I don't have a Windows VM built yet to verify at the moment. Maybe in a couple of days.

PR #2902, opened 2017-01-17
add ``override`` option to ``add_translation_dirs``

### LINKED ISSUE #1474: add_translation_dirs behaves the wrong way around
This is a follow-up to #1473. See the reproducer and notes there.

I would propose an API change that changes behaviour of add_translation_dirs to append specs at the end of existing specs instead of prepending it. That way, API users simply need to understand "last spec wins". This matches much closer to the mental model of "add".

It matches the current documented behaviour of how the method is to _behave_ in case of multiple calls.

mmerickel:
I debated with myself for a very long time whether to set ``override=True`` by default or not and in the end I simply cannot come up with a compelling reason to break bw-compat. I took a page out of pyramid_jinja2's book with `add_jinja2_search_path` which also adds new items with lower priority than previously added items. This is consistent with that, as well as previous behaviors.

fixes #1474

### DIFF
--- a/pyramid/config/i18n.py
+++ b/pyramid/config/i18n.py
@@ -42,7 +42,7 @@ class I18NConfiguratorMixin(object):
         self.registry.registerUtility(locale_negotiator, ILocaleNegotiator)
 
     @action_method
-    def add_translation_dirs(self, *specs):
+    def add_translation_dirs(self, *specs, **kw):
         """ Add one or more :term:`translation directory` paths to the
         current configuration state.  The ``specs`` argument is a
         sequence that may contain absolute directory paths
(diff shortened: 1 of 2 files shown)


piotr-dobrogost:
> `add_jinja2_search_path` which also adds new items with lower priority than previously added items

Heh, then it's the wrong default there, too :)

> I simply cannot come up with a compelling reason to break bw-compat. 

The fact no one before @rbu raised this means that those, who use `add_translation_dirs` use it **only** to add translations and not to overwrite them. This means they have no overwrites **intended** to work because if they had got any they wouldn't have worked and they would have submitted issue. Conclusion – if `overwrite=True` breaks something it means it was already broken (by having overwrites not meant to overwrite).

PR #2917, opened 2017-01-25
restore the registry signature and add tests to show why

### LINKED ISSUE #2916: Restore the Registry(*args, **kw) API
In the course of #2891, the Registry API was changed in a backwards-incompatible way. While the author of the change (PR 2893) couldn't imagine anyone using the keywords API, I immediately ran into the incompatibility with a number of customer projects. Since the change back to the old API shouldn't introduce another backwards incompatibility, I suggest reverting that change as soon as possible.

mmerickel:
fixes #2916

### DIFF
--- a/pyramid/registry.py
+++ b/pyramid/registry.py
@@ -56,7 +56,7 @@ class Registry(Components, dict):
 
     _settings = None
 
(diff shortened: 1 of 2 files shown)


mmerickel:
No need to for a changelog here. Backported to 1.8-branch.

PR #2919, opened 2017-01-26
Fix development.ini for wsgiref

yleteigner:
Looks like the listen statement is ignored by wsgiref, but using the "port = 6543" instead causes the server to listen on the correct port.

### DIFF
--- a/docs/quick_tutorial/ini/development.ini
+++ b/docs/quick_tutorial/ini/development.ini
@@ -3,4 +3,4 @@ use = egg:tutorial
 
 [server:main]
 use = egg:pyramid#wsgiref
-listen = *:6543
+port = 6543


bertjwregeer:
This needs to be merged into master too...

stevepiercy:
There may be other instances across the docs at each step, as each step builds upon previous steps.

stevepiercy:
Oh, yes, indeed. There's 53 potential occurrences of `wsgiref` using `listen` instead of `port`. I'll comb through the lot in the next day or two. Thanks for bringing this matter to our attention.

bertjwregeer:
Ugh, what a mess :-/ 

why are we using `wsgiref` for those anyway? Instead of `waitress`?

stevepiercy:
https://www.youtube.com/watch?v=gRdfX7ut8gw

stevepiercy:
Closing this PR in favor of #2923 and #2924. Thanks for the report @yleteigner.

PR #2932, opened 2017-01-28
Minor wording changes

kikuchiyo:
"or any callable that accepts" flows a little better.  Another option could be, "any callables that accept", matches the plural "functions" and "classes", and allows use of "accept", if that sounds better.


kikuchiyo:
I'm not sure which one works better -

PR #2933, opened 2017-01-28
Minor change

kikuchiyo:
"classes, functions, or any callable that accept" - the mix of plural and singular followed by "accept" which requires "accept" for "classes" and "functions", but "accepts" for "callable".  Proposed change sticks with only singular "any class, function, or callable that accepts" - flows a little better while reading.


stevepiercy:
Thank you!

PR #2944, opened 2017-02-04
Minor change

kikuchiyo:
"configuration' should read "configuration" with two double quotes instead of one double quote and one single quote.


stevepiercy:
@kikuchiyo thanks for the report. The word should not be quoted at all, so I'm closing this PR in favor of #2945.

Issue #2958, opened 2017-02-16
Configuration is changing the settings parameter object


Natim:
This is a regression from Pyramid 1.7.

```
.venv/lib/python3.5/site-packages/pyramid/config/__init__.py:329: in __init__
    exceptionresponse_view=exceptionresponse_view,
.venv/lib/python3.5/site-packages/pyramid/config/__init__.py:367: in setup_registry
    self._set_settings(settings)
.venv/lib/python3.5/site-packages/pyramid/config/settings.py:9: in _set_settings
    settings = Settings(mapping)
.venv/lib/python3.5/site-packages/pyramid/config/settings.py:59: in Settings
    d.update(**kw)
```

We should `copy` or `copy.deepcopy` the settings object to prevent this.

mmerickel:
Fixed in #2959. Thank you.

PR #2959, opened 2017-02-16
The Configurator object should not alter the input settings dict object.

Natim:
Test showing #2958 behavior.

### DIFF
--- a/pyramid/config/settings.py
+++ b/pyramid/config/settings.py
@@ -56,6 +56,7 @@ def Settings(d=None, _environ_=os.environ, **kw):
     keyword args)."""
     if d is None:
         d = {}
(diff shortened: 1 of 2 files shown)


mmerickel:
Can you please rewrite your commits without GPG signatures? I'm pretty confident we don't have any other commits in our history with signatures and I'm concerned about allowing them without a discussion on the implications. I'd rather avoid that for now.

If we need to have this discussion because you feel so strongly about it then it'll have to wait until I have time to do more research, meaning this fix doesn't get applied for longer.

Natim:
Signed commit shouldn't cause you any trouble but I will change that ok.

mmerickel:
Thank you @Natim !

PR #2962, opened 2017-02-20
Pserve --reload does not keep worker arguments. Fixes #2961

### LINKED ISSUE #2961: pserve --reload call hupper which call pserve with sys.argv instead of the parameters
If you are calling `pserve.main(argv)` and `argv` contains `--reload` it will call upper with `sys.argv` instead of `argv`.

Natim:
Refs #2961

### DIFF
--- a/pyramid/tests/test_scripts/test_pserve.py
+++ b/pyramid/tests/test_scripts/test_pserve.py
@@ -1,8 +1,10 @@
+import mock
 import os
 import unittest
 from pyramid.tests.test_scripts import dummy
 
 here = os.path.abspath(os.path.dirname(__file__))
+
 
 class TestPServeCommand(unittest.TestCase):
     def setUp(self):
@@ -48,10 +50,11 @@ class TestPServeCommand(unittest.TestCase):
         inst.loadserver = self._get_server
 
         app = dummy.DummyApp()
+
         def get_app(*args, **kwargs):
             app.global_conf = kwargs.get('global_conf', None)
-        inst.loadapp = get_app
 
+        inst.loadapp = get_app
         inst.run()
         self.assertEqual(app.global_conf, {'a': '1', 'b': '2'})
 
@@ -76,6 +79,21 @@ class TestPServeCommand(unittest.TestCase):
             os.path.abspath('/baz'),
             os.path.abspath(os.path.join(here, '*.py')),
         ])
+
+    def test_reload_call_hupper_with_correct_args(self):
+        with mock.patch('pyramid.scripts.pserve.hupper') as hupper_mock:
+            hupper_mock.is_active.side_effect = [False, True]
+            inst = self._makeOne('--reload', 'development.ini')
+            inst.loadserver = mock.MagicMock()
+            inst.loadapp = mock.MagicMock()
+            inst.run()
+            hupper_mock.start_reloader.assert_called_with(
+                'pyramid.scripts.pserve.main',
+                reload_interval=1,
+                verbose=1,
+                worker_kwargs={'argv': ['pserve', '--reload', 'development.ini'],
+                               'quiet': False})
+
 
 class Test_main(unittest.TestCase):
     def _callFUT(self, argv):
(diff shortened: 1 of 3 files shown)


bertjwregeer:
Please don't mix whitespace/pep8 fixes in with the rest of the commit in the future. It makes it more difficult to see what the changes are.

pyramid/tests/test_scripts/test_pserve.py
```diff
@@ -1,9 +1,11 @@
+import mock
```
bertjwregeer: We don't currently use mock in the the Pyramid tests, instead we create dummy objects that have just the functionality required.

I don't think it is a great idea to add yet another dependency...

setup.py
```diff
@@ -65,9 +65,10 @@
     ]
 
 testing_extras = tests_require + [
+    'mock',
```
bertjwregeer: I'm a -1 on this, we use dummy objects everywhere else.

Natim:
> Please don't mix whitespace/pep8 fixes in with the rest of the commit in the future. It makes it more difficult to see what the changes are.

That's why I put them in another commit so that you can look at the diff without them.

> I don't think it is a great idea to add yet another dependency...

It doesn't change anything to add another dependency, but I will try to rewrite my test with a Dummy object instead. (mock just makes it easy to do so with less code to maintain.)

setup.py
Natim: By the way this dependencies are only installed when you run tests which is not something you do when you use the project.

bertjwregeer: Sure, I understand that.

mmerickel:
ping @Natim

Natim:
Hi @mmerickel, I didn't forget about this and I am going to rewrite tests without mocks.

mmerickel:
@Natim Okie sounds good.. just wanted to make sure it wasn't forgotten!

Natim:
To be honest I find the code much harder to maintain with a test like that. Are you sure it worth getting rid of the mock dependency?

Natim:
@bertjwregeer @mmerickel how does it look to you now?

Natim:
Any news about it?

pyramid/tests/test_scripts/test_pserve.py
```diff
@@ -77,6 +80,32 @@ def test_config_file_finds_watch_files(self):
...
+
+        def dummy_start_reloader(*args, **kwargs):
+            dummy_start_reloader.args = args
+            dummy_start_reloader.kwargs = kwargs
+
+        pserve.hupper = AttrDict(is_active=lambda: False,
+                                 start_reloader=dummy_start_reloader)
+
+        inst = self._makeOne('--reload', 'development.ini')
+        inst.run()
```
mmerickel: I know we aren't using mock (sorry!) but if you're patching a module-scope value, you need to replace it after the test is over. For example:

```
orig_hupper = pserve.hupper
try:
    pserve.hupper = ...
    inst = ...
    inst.run()
finally:
    pserve.hupper = orig_hupper
```

Natim:
> I know we aren't using mock (sorry!) 

Why cannot we start?

mmerickel:
> Why cannot we start?

Well I'm not adding a new dependency in a bugfix release and I imagine we want to backport this to 1.8. I'm definitely open to improving this situation in 1.9 though.

Natim:
> Well I'm not adding a new dependency in a bugfix release and I imagine we want to backport this to 1.8. I'm definitely open to improving this situation in 1.9 though.

Ok it makes sense 👍

mmerickel:
Just waiting on @bertjwregeer to approve this before I continue.

@Natim if you come back with more bugs the day after I cut a new release again I'm going to bonk you on the head if I ever see you in person!

bertjwregeer:
I will say that I am not all that excited about using Mock. I feel like mock is overused for "magic" to make tests pass, and it just overall makes a bad experience for writing tests. I think we can try to use it, but we have to be careful to not be overly reliant on it.

Natim:
Hopefully releases are cheap :)

mmerickel:
@Natim I wish that were true. :-(

PR #2964, opened 2017-02-25
add an IExecutionPolicy

mmerickel:
The goal here is a hook that can actually create new requests. This gives the execution policy the ability to control the execution of a request in pyramid to the extent that it can throw away all state and try again similar to what could be done in middleware.

I'm writing this for the purpose of pyramid_tm which wants to support "retryable" requests in which the request failed for transient reason where there is a reasonable idea that the request will succeed if tried again. When performing a retry we want to teardown state as much as possible to avoid any state leaking from the failing request into the new request (such as managed objects for aborted transactions).

The key difference between this API and a tween is that the execution policy can create a new request with an assurance that nothing else is wrapping it, depending on the old request object and its old state. Whether this is worth the introduction of the API or not is up for debate but I think it is after trying to implement pyramid_tm without it.

### DIFF
--- a/docs/api/config.rst
+++ b/docs/api/config.rst
@@ -70,6 +70,7 @@
      .. automethod:: add_subscriber_predicate
      .. automethod:: add_view_predicate
      .. automethod:: add_view_deriver
+     .. automethod:: set_execution_policy
      .. automethod:: set_request_factory
      .. automethod:: set_root_factory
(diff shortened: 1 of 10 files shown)


bertjwregeer:
I sent you a PR to fix the docs issue. Besides that, this looks good to me. I am looking forward to this!

PR #2967, opened 2017-03-01
Fixed several reference cycles to prevent memory leaks.

Cykooz:
I have corrected three places  in the code where appear reference cycles due to closure.
Also I have added a simple test for detect memory leaks after application closing.

### DIFF
--- a/pyramid/config/views.py
+++ b/pyramid/config/views.py
@@ -933,7 +933,7 @@ class ViewsConfiguratorMixin(object):
             if not exception_only and isexc:
                 derived_view = runtime_exc_view(derived_view, derived_exc_view)
 
-            derived_view.__discriminator__ = lambda *arg: view_intr.discriminator
+            derived_view.__discriminator__ = lambda *arg: discriminator
             # __discriminator__ is used by superdynamic systems
             # that require it for introspection after manual view lookup;
             # see also MultiView.__discriminator__
--- a/pyramid/config/views.py
+++ b/pyramid/config/views.py
@@ -933,7 +933,7 @@ class ViewsConfiguratorMixin(object):
             if not exception_only and isexc:
                 derived_view = runtime_exc_view(derived_view, derived_exc_view)
 
-            derived_view.__discriminator__ = lambda *arg: discriminator
+            derived_view.__discriminator__ = lambda *arg: view_intr.discriminator
             # __discriminator__ is used by superdynamic systems
             # that require it for introspection after manual view lookup;
             # see also MultiView.__discriminator__
--- a/pyramid/tests/test_integration.py
+++ b/pyramid/tests/test_integration.py
@@ -9,6 +9,7 @@ import unittest
 from pyramid.wsgi import wsgiapp
 from pyramid.view import view_config
 from pyramid.static import static_view
+from pyramid.testing import skip_on
 from pyramid.compat import (
     text_,
     url_quote,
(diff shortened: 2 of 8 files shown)


pyramid/tweens.py
```diff
@@ -42,7 +42,7 @@ def excview_tween(request):
             provides = providedBy(exc)
             try:
                 response = _call_view(
-                    registry,
+                    request.registry,
```
mmerickel: Again, I guess the issue here is that the registry eventually retains the `excview_tween` and the `excview_tween` retains the registry via the closure, but this is all cleaned up via GC in general and likely occurs in many other parts of code.

What practical problem is this solving that I should accept this change? Is the submitted integration test going to prevent me from writing this code again later?

mmerickel: Sorry but just to clarify I think the integration test and the weakref fixes on the WeakOrderedSet are good. I'm just nitpicking these other two changes.

pyramid/config/views.py
```diff
@@ -933,7 +933,7 @@ def register(permission=permission, renderer=renderer):
             if not exception_only and isexc:
                 derived_view = runtime_exc_view(derived_view, derived_exc_view)
 
-            derived_view.__discriminator__ = lambda *arg: discriminator
+            derived_view.__discriminator__ = lambda *arg: view_intr.discriminator
```
Cykooz: This reference cycle prevents the removal of instance of `Сonfigurato`r after a WSGI application has created.
If skip some steps then cycle looks like that:
`discriminator` (always instance of Deferred in this place) => `discrim_func` =>  `Configurator` (as `self` in closure) => `Registry` => `Introspector` => `default_exceptionrespone_view` => `__discrimantor__` => lambda => closure with `discriminator`.
My fix removes the instance of `Deferred` from closure. Instead it, closure contains now instance of `Introspectable` with **undeferred** value of descriminator.

For details, see the attached image with graph of backrefs of Configurator instance after running of this code:
```
app = Configurator().make_wsgi_app()
gc.collect()
```
cycle1

pyramid/tweens.py
Cykooz: Hmm... Yes, I agree with you - this closure has not matter and do not prevents GC to resolve a cycle "registry-tween-registry" .
I could remove this change, if you think that it is absolutely useless.

pyramid/tests/test_integration.py
```diff
@@ -741,3 +743,29 @@ def _assertBody(body, filename):
...
+
+class MemoryLeaksTest(unittest.TestCase):
+
+    def tearDown(self):
+        import pyramid.config
+        pyramid.config.global_registries.empty()
+
+    def get_gc_count(self):
+        last_collected = 0
+        while True:
```
Cykooz: I don't know :) May be it possible in some cases with threads that works simultaneously with this test. Then GC will collects different count of objects in each step of cycle.

pyramid/config/views.py
mmerickel: Would this be solved if we fixed the `Deferred` to release the closure when it is resolved? Imagine `self.func = None` in https://github.com/Pylons/pyramid/blob/7fb889349d40e361397ed31f3535689182aa5cf0/pyramid/registry.py#L278

Cykooz: Yes, it was a first solution that I have tried when investigate this reference cycle. It works too.

mmerickel: I think I'd prefer that solution because it feels less error prone. The current fix is quite subtle.

mmerickel:
Thank you @Cykooz. I'm leaving this open for now until we backport these fixes to 1.7 and 1.8 as well.

PR #2974, opened 2017-03-06
Make sure PServeCommand kwargs are passed to the hupper worker.

mmerickel:
Backport of #2962 to 1.8-branch.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,11 @@ Bug Fixes
   object in which the registry was stored in a closure preventing it from
   being deallocated. See https://github.com/Pylons/pyramid/pull/2973
 
+- Fix a bug directly invoking ``pyramid.scripts.pserve.main`` with the
+  ``--reload`` option in which ``sys.argv`` is always used in the subprocess
+  instead of the supplied ``argv``.
+  See https://github.com/Pylons/pyramid/pull/2974
+
 Documentation Changes
 ---------------------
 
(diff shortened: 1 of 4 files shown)


Natim:
Hi @mmerickel Would you have some bandwith to cut a 1.8.3 release with this fix please?

bertjwregeer:
@Natim Are there any other issues that need fixing first? Cutting releases takes a bit of time and we don't want to have to spin 1.8.4 tomorrow...

Natim:
If they were I would have filed PR...

mmerickel:
> If they were I would have filed PR...

We were just referring to the last release when you opened a new bug report a day after, so it would be appreciated if you tested the unreleased 1.8-branch first to avoid anything obvious showing up immediately again. Plus you will get them fixed sooner without waiting a few weeks for another release.

Regardless, please don't take that as an insult, I certainly appreciate the bug reports. I will cut a release this weekend at the latest with the changes.

stevepiercy:
@Natim releasing is not hard, it is merely time consuming. See our complete checklist, [RELEASING.txt](https://github.com/Pylons/pyramid/blob/master/RELEASING.txt).

If `zest.releaser` can automate all or most of all those details, I'd like to know more, but let's discuss in a new separate issue as a feature request.

Natim:
Apparently we actually did :( https://github.com/Pylons/pyramid/issues/2975

mmerickel:
@Natim that's a change of a default... that won't be going into a 1.8 bugfix release unfortunately.

Natim:
Ok then we are good for 1.8.3

Issue #1394, opened 2014-08-09
Allow alternatives to Paste ini files to work with pshell/pserve


inklesspen:
At my workplace, we've "standardized" on [an alternate configuration file format](http://axialcorps.com/2014/02/06/betterconfig/). Other workplaces may want to use JSON or YAML or some such. These can basically be turned into a dict the same way ConfigParser ini files are, so why not allow specifying a config parser function? Perhaps via a prefix, the same way asset specifications can reference a different package's assets. So I could do myfile.ini for the standard functionality, or json:myfile.json to specify that it's a json file. That way I can still use all the nice Pyramid tooling.

(Somewhat related to issue #560.)

mmerickel:
related #2419

mmerickel:
Closed via https://github.com/Pylons/pyramid/pull/2985

PR #2985, opened 2017-03-28
migrate pyramid to use plaster

### LINKED ISSUE #2419: improve config file support
Pyramid's config file support has been dependent for a long time on PasteDeploy much to the chagrin of many users who hate ini files due to lack of support for types, environment variable support and non-wsgi scripts.

This issue is about coming up with a better design, so let me first layout some goals:
- Alternate file formats: YAML, TOML, INI, etc. The only stipulation is that one of these formats is capable of supporting PasteDeploy-style files to maintain bw-compat.
- Environment variables. There are no specific requirements here but each individual format may support some way of injecting environment variables into the config, along with fallback values for missing values.
- Alternate runners. PasteDeploy is broken into wsgi components, servers and logging. A missing component is an arbitrary cli application. Along with starting wsgi servers, commonly it should be possible to also pull settings for other cli-based applications such as background workers and other scripts. These apps are about defining groups of settings, not about a Python API for defining a command line app such as `main(global_conf, **settings)`.

Looking at what we have right now. We have `pserve` which is cli app that loads settings from a couple parts of an ini file. 1) It loads a `[server]` and then 2) it loads a wsgi app which can be composed from other wsgi apps. In a very general sense `pserve` is just running some code which is loading settings for isolated targets that it composes together. Great. That's pserve.

I think it should be possible to accomplish these goals without completely overhauling what exists but rather just adding a layer of indirection between Pyramid and PasteDeploy as well as implementing some alternate backends for popular formats (YAML).

So, what can we simplify?

PasteDeploy is doing too much. It is a tightly coupled library between an ini file and wsgi components (app, server). We need to remove PasteDeploy from the pipeline as much as possible and focus on converting arbitrary file formats into sections. Each parser should, on its own, provide support for default values as well as values from environment vars.

A good parser can then be written to support the PasteDeploy format that `pserve` requires.

@inklesspen has put a lot of work into [montague](https://github.com/inklesspen/montague) and it improves on several of these points including abstracting the file format from the PasteDeploy format. I think the parser itself is where things falls short. An explicit goal I have is to be able to use montague (or simply the underlying parser) to load sections. The WSGI component is a layer on top of this. The difference here is that I'm trying to make it an explicit goal to make the config file itself reusable outside of a WSGI context. This is only semi-possible right now, and a source of endless problems when using PasteDeploy.
- PasteDeploy modifies the ini parser but does not expose it publicly so no one else can use it.
- As a result there are a lot of inconsistencies on whether the file is parsed using PasteDeploy or whether it is parsed by the stdlib `ConfigParser`.
- Issues involving escaping variables differently between sections.
- No inheritance or shared settings. PasteDeploy defines some options, but they aren't usable outside of its supported sections.

mmerickel:
- [x] Replace low-level functions with plaster.
- [x] Update proutes.
- [x] Update pserve.
- [x] Update pshell.
- [x] Update logging chapter.
- [x] Update PasteDeploy chapter.
- ~~Update cookiecutters to depend on plaster_pastedeploy.~~

Fixes #2419, #1394.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,14 +1,32 @@
 unreleased
 ==========
 
-Features
---------
+Major Features
+--------------
+
+- The file format used by all ``p*`` command line scripts such as ``pserve``
+  and ``pshell``, as well as the ``pyramid.paster.bootstrap`` function
+  is now replaceable thanks to a new dependency on
+  `plaster <http://docs.pylonsproject.org/projects/plaster/en/latest/>`.
+
+  For now, Pyramid is still shipping with integrated support for the
+  PasteDeploy INI format by depending on the ``plaster_pastedeploy`` binding.
+
+  See https://github.com/Pylons/pyramid/pull/2985
 
 - Added an execution policy hook to the request pipeline. An execution
   policy has the ability to control creation and execution of the request
-  objects before they enter rest of the pipeline. This means for a given
-  request that the policy may create more than one request for retry
-  purposes. See https://github.com/Pylons/pyramid/pull/2964
+  objects before they enter rest of the pipeline. This means for a single
+  request environ the policy may create more than one request object.
+
+  The first library to use this feature is
+  `pyramid_retry
+  <http://docs.pylonsproject.org/projects/pyramid-retry/en/latest/>`.
+
+  See https://github.com/Pylons/pyramid/pull/2964
+
+Features
+--------
 
 - Support an ``open_url`` config setting in the ``pserve`` section of the
   config file. This url is used to open a web browser when ``pserve --browser``
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,7 +7,7 @@ Major Features
 - The file format used by all ``p*`` command line scripts such as ``pserve``
   and ``pshell``, as well as the ``pyramid.paster.bootstrap`` function
   is now replaceable thanks to a new dependency on
-  `plaster <http://docs.pylonsproject.org/projects/plaster/en/latest/>`.
+  `plaster <http://docs.pylonsproject.org/projects/plaster/en/latest/>`_.
 
   For now, Pyramid is still shipping with integrated support for the
   PasteDeploy INI format by depending on the ``plaster_pastedeploy`` binding.
@@ -16,12 +16,12 @@ Major Features
 
 - Added an execution policy hook to the request pipeline. An execution
   policy has the ability to control creation and execution of the request
-  objects before they enter rest of the pipeline. This means for a single
+  objects before they enter the rest of the pipeline. This means for a single
   request environ the policy may create more than one request object.
 
   The first library to use this feature is
   `pyramid_retry
-  <http://docs.pylonsproject.org/projects/pyramid-retry/en/latest/>`.
+  <http://docs.pylonsproject.org/projects/pyramid-retry/en/latest/>`_.
 
   See https://github.com/Pylons/pyramid/pull/2964
 
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -366,6 +366,14 @@ Glossary
      :term:`WSGI` components together declaratively within an ``.ini``
      file.  It was developed by Ian Bicking.
 
+   plaster
+     `plaster <http://docs.pylonsproject.org/projects/plaster/en/latest/>`_ is
+     a library used by :app:`Pyramid` which acts as an abstraction between
+     command-line scripts and the file format used to load the :term:`WSGI`
+     components and application settings. By default :app:`Pyramid` ships
+     with the ``plaster_pastedeploy`` library installed which provides
+     integrated support for loading a :term:`PasteDeploy` INI file.
+
    Chameleon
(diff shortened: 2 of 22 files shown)


stevepiercy:
What's the milestone/target version for these items?

mmerickel:
At least for now I've updated pyramid to depend on plaster_pastedeploy which makes this PR bw-compatible. Shipping pyramid without any specific config-file loader is something we can debate. It's a pretty substantial amount of stuff, even just in core, that needs fixes if we rip out that dependency.

mmerickel:
I think I'm done with this for now, pending review. I updated a couple locations in the documentation but tried not to make too big of a deal about it at this time. It's a new feature and there's no sense in jamming it down people's throats so I just tried to give some clues that pastedeploy could be replaced.

CHANGES.txt
```diff
@@ -1,14 +1,32 @@
...
-Features
---------
+Major Features
+--------------
+
+- The file format used by all ``p*`` command line scripts such as ``pserve``
+  and ``pshell``, as well as the ``pyramid.paster.bootstrap`` function
+  is now replaceable thanks to a new dependency on
+  `plaster <http://docs.pylonsproject.org/projects/plaster/en/latest/>`.
+
```
stevepiercy: rst links require a trailing underscore:

```rst
`label <link>`_
```

CHANGES.txt
```diff
@@ -1,14 +1,32 @@
...
-  objects before they enter rest of the pipeline. This means for a given
-  request that the policy may create more than one request for retry
-  purposes. See https://github.com/Pylons/pyramid/pull/2964
+  objects before they enter rest of the pipeline. This means for a single
+  request environ the policy may create more than one request object.
+
+  The first library to use this feature is
+  `pyramid_retry
+  <http://docs.pylonsproject.org/projects/pyramid-retry/en/latest/>`.
+
```
stevepiercy: add trailing _

CHANGES.txt
```diff
@@ -1,14 +1,32 @@
...
+  PasteDeploy INI format by depending on the ``plaster_pastedeploy`` binding.
+
+  See https://github.com/Pylons/pyramid/pull/2985
 
 - Added an execution policy hook to the request pipeline. An execution
   policy has the ability to control creation and execution of the request
-  objects before they enter rest of the pipeline. This means for a given
-  request that the policy may create more than one request for retry
-  purposes. See https://github.com/Pylons/pyramid/pull/2964
+  objects before they enter rest of the pipeline. This means for a single
```
stevepiercy: enter /the/ rest

docs/glossary.rst
```diff
@@ -366,6 +366,14 @@ Glossary
      :term:`WSGI` components together declaratively within an ``.ini``
      file.  It was developed by Ian Bicking.
 
+   plaster
+     `plaster <http://docs.pylonsproject.org/projects/plaster/en/latest/>` is
```
stevepiercy: add trailing _

docs/narr/paste.rst
```diff
@@ -96,3 +91,8 @@ applications, servers, and :term:`middleware` defined within the configuration
 file.  The values in a ``[DEFAULT]`` section will be passed to your
 application's ``main`` function as ``global_config`` (see the reference to the
 ``main`` function in :ref:`init_py`).
+
+Alternative Configuration File Formats
+--------------------------------------
+
+It is possible to use different file formats with :app:`Pyramid` if you do not like :term:`PasteDeploy`. Under the hood all command-line scripts such as ``pserve`` and ``pshell`` pass the ``config_uri`` (e.g. ``development.ini``, ``production.ini``, etc) to the :term:`plaster` library which performs a lookup for an appropriate parser. For ``.ini`` files it uses PasteDeploy but you can register your own configuration formats that plaster will find instead.
```
stevepiercy: (for example, ``development.ini`` or ``production.ini``)

mmerickel: What's wrong with `e.g.`? It's not wrong and from a quick grep of our sources it's used in like 266 spots already.

CHANGES.txt
mmerickel: Thanks I pushed some changes.

docs/narr/paste.rst
stevepiercy: 1. Many style guides use plain English over the Latin abbreviation—"for example" is preferred over "e.g.", "in other words" and "that is" over "i.e.", and "and so on" over "etc.". Reasons include:
    - the intention is clearer
    - avoids misuse of the abbreviations, usually due to the lack of knowledge of Latin
    - has a less erudite tone
    - easier to translate
2. The use of "e.g." with "etc." in the same phrase is redundant.
3. If "e.g." or "i.e." must be used, then they are followed normally by a comma in American english, and confined within parentheses.

Eventually I would like to replace Latin abbreviations with common American english in the narrative documentation so that it reads smoother.

bertjwregeer:
I forgot to comment on this, thanks Github UI. :+1: glad to see this happen :D.

PR #2989, opened 2017-04-02
push threadlocals while executing config.include functions

mmerickel:
We don't push threadlocals in the `main` because there is no clear contract on its lifecycle. However we can push them during `config.include(...)` calls so that they are available there. Just a little more incentive to put all your code into include functions!

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -18,6 +18,12 @@ Features
   requirement that the server is being run in this format so it may fail.
   See https://github.com/Pylons/pyramid/pull/2984
 
+- The threadlocals are now available inside any function invoked via
+  ``config.include``. This means the only config-time code that cannot rely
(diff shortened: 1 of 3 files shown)


piotr-dobrogost:
> We don't push threadlocals in the main because there is no clear contract on its lifecycle. 

Michael, could you please elaborate on this lifecycle issue (I'm curious)?
Btw, is there any section in the docs for this kind of information?

mmerickel:
The lifecycle issue is that there is no clear ending point to the Configurator lifetime. This means that it cannot automatically push the threadlocals because it cannot remove them at an appropriate spot. The only possible spot is `make_wsgi_app` but that is not good enough because the Configurator can be used to make multiple wsgi apps in testing scenarios and such.

PR #2993, opened 2017-04-10
add execution policy to pyramid request processing diagrams

stevepiercy:


piotr-dobrogost:
It is really nice to have this diagram. I wonder if placing name of current stage of processing in the request object would have any use?

stevepiercy:
@piotr-dobrogost perhaps. Can you describe a specific example?

The diagram is pretty dense as it is. If there were to be any additional information presented, I would have to research ways of including tool tips or raw HTML with Sphinx.

piotr-dobrogost:
Well, I was talking about changes in source code not diagram. What I had in mind is on start of each phase of processing shown on the diagram to *tag* request object with the name of the phase. This would allow to inspect current phase when for example exception in thrown. In other words this would allow to map place in source code with the phase from the diagram.

PR #2994, opened 2017-04-12
cgi.escape() is deprecated. (For Python 3)

jeremy886:
I just realised this can be a compatible issue for Python 2 and 3 because no warning in the Python 2.7 doc. I am not sure how the Pylon project deals with this. I will leave this decision to you and learn from it. From a newbie. Cheers.

Source: https://docs.python.org/3.6/library/cgi.html
Deprecated since version 3.2: This function is unsafe because quote is false by default, and therefore deprecated. Use html.escape() instead.

### DIFF
--- a/docs/tutorials/wiki2/src/views/tutorial/views/default.py
+++ b/docs/tutorials/wiki2/src/views/tutorial/views/default.py
@@ -1,4 +1,4 @@
-import cgi
+import html
 import re
 from docutils.core import publish_parts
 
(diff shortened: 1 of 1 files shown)


uralbash:
@jeremy886 pyramid.compat.escape is the more better choice (http://docs.pylonsproject.org/projects/pyramid/en/latest/api/compat.html#pyramid.compat.escape)

mmerickel:
Currently we have a caveat in the docs that says the tutorial runs on both python 2 and 3 without modification which means we cannot accept this PR as is. We could add a test import.

```python
try:
    from html import escape
except ImportError:  # python 2 fallback
    from cgi import escape
```

ztane:
I'd too import `escape` from `pyramid.compat`

ztane:
Additionally `cgi.escape` and `html.escape` are not even compatible (which is the reason behind the deprecation).

jeremy886:
I learned a lot.  Do I need to do anything? Close pull request?

stevepiercy:
@jeremy886 It would be nice if you put into practice what you learned by updating the pull request so we can accept it.

mmerickel:
@jeremy886 Please make a new pull request that uses `pyramid.compat.escape` and use the master branch.

PR #3016, opened 2017-04-27
Fix underlined title.

Natim:
Refs https://github.com/Pylons/pyramid-cookiecutter-starter/pull/28


bertjwregeer:
Thanks!

stevepiercy:
@bertjwregeer in the future before merging docs source files, please be aware that there is usually more than just one change. In this case, we have 3 cookiecutters being updated and all the README.txt that they generate in the Pyramid docs. I like to keep things in a consistent state and do them all in one swoop. I'll work on the other changes tonight when I have the time. Thank you!

PR #3019, opened 2017-04-29
Decouple CSRF protection from the session machinery (replaced #2854)

mmerickel:
Replaces #2854.

- `SessionCSRFStoragePolicy` should really just use the session as a storage mechanism. I added `LegacySessionCSRFStoragePolicy` for the version that uses the legacy `ISession.get_csrf_token` and `ISession.new_csrf_token` apis and set that as the default. I also updated `SessionCSRFStoragePolicy` to just store data in the session similar to how the `SessionAuthenticationPolicy` works.

- `CookieCSRFStoragePolicy` did not properly return the newly generated token from `get_csrf_token` after calling `new_csrf_token`. It needs to cache the new value since the response callback does not affect the current request.

- `CookieCSRFStoragePolicy` was not forwarding the `domain` value to the `CookieProfile` causing that setting to be ignored.

- ~I removed `check_csrf_token` from the `ICSRFStoragePolicy` interface because I can't see the point to making it pluggable. On top of that it makes implementation of storage policies more difficult.~

- Added an introspectable item for the configured storage policy so that it appears on the debugtoolbar.

- Added a change note on `ISession` that it no longer required the csrf methods.

- Restored (but deprecated) shims in `pyramid.session` for `check_csrf_origin` and `check_csrf_token`.


bertjwregeer:
> I removed check_csrf_token from the ICSRFStoragePolicy interface because I can't see the point to making it pluggable. On top of that it makes implementation of storage policies more difficult.

@MatthewWilkes makes a strong point for keeping it, although there is a work-around, but not one I would think is that great :-/

The rest of it looks great :+1:

mmerickel:
I've restored the `ICSRFStoragePolicy.check_csrf_token` api. I've decided it's worth keeping as checking a token is legitimately the concern of the storage mechanism and is a separate operation from querying the current token.

PR #3056, opened 2017-05-23
[WIP] Make pyramid.crsf.check_csrf_token to read from config's default crs value

### LINKED ISSUE #3021: update check_csrf_token to use the default csrf options, by default
We have `config.set_default_csrf_options` which can control the key/header used to search for tokens. It would be nice if `pyramid.csrf.check_csrf_token` would use those settings, by default, unless values are explicitly passed to the function. This would probably also involve respecting the `safe_methods` argument which would be a compatibility change.

fangpenlin:
Fix #3021

## In progress

### DIFF
--- a/pyramid/csrf.py
+++ b/pyramid/csrf.py
@@ -15,7 +15,10 @@ from pyramid.exceptions import (
     BadCSRFOrigin,
     BadCSRFToken,
 )
-from pyramid.interfaces import ICSRFStoragePolicy
+from pyramid.interfaces import (
+    ICSRFStoragePolicy,
+    IDefaultCSRFOptions
+)
 from pyramid.settings import aslist
 from pyramid.util import (
     is_same_domain,
@@ -179,7 +182,7 @@ def new_csrf_token(request):
 
 
 def check_csrf_token(request,
-                     token='csrf_token',
+                     token=None,
                      header=None,
                      raises=True):
     """ Check the CSRF token returned by the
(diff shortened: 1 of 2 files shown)


pyramid/csrf.py
```diff
@@ -214,8 +218,37 @@ def check_csrf_token(request,
...
+            # TODO: extract as constant var?
+            token = 'csrf_token'
+    if safe_methods is None:
+        if csrf_options.safe_methods is not None:
+            safe_methods = csrf_options.safe_methods
+        else:
+            # TODO: I saw these default values appear at many places, maybe we
+            # should extract them as constant values somehwere and use them,
+            # so that we won't be out of sync?
+            safe_methods = frozenset(["GET", "HEAD", "OPTIONS", "TRACE"])
```
fangpenlin: @berkerpeksag @mmerickel @stevepiercy is there a shared module for these repeating values? not sure what's the practice here

mmerickel: I think one good option is just to always define CSRF options by calling `config.set_default_csrf_options(require_csrf=False)` in `add_default_security` here:

https://github.com/Pylons/pyramid/blob/f6f6d13be6c9ca384fefc35729ecf32a20b8b50d/pyramid/config/security.py#L21

Then you can remove the duplication and just assume you can query a `IDefaultCSRFOptions` object from the registry.

bertjwregeer: The frozenset here is not that big if a problem. It is not likely to ever change, and moving it elsewhere to import it won't necessarily help make much of a difference.

fangpenlin:
I only stay one day at dev sprints, will continue working on this during weekend. in the meantime if anyone wants to takes over this issue, please feel free.

mmerickel:
I will probably look at this tonight.

mmerickel:
I'm waffling on this feature right now.

My original intention was just to pull the token/header from the default csrf options. The problems arise when I start to think about pulling the `safe_methods` setting and possibly even `callback` from the default csrf options. It begins the debate about whether `check_csrf_token` should be equivalent to `require_csrf=True` except usable manually or whether it should be a low level function that simply checks the header/token but doesn't care about the current request method, etc.

So far I'm probably leaning toward rejecting this change.

fangpenlin:
@mmerickel cool, thank you for looking into this. and sorry I was super busy these days had no time to continue the work, wish I had more time on sprints with youguys. 

And I think it makes sense to leave the method as a low level one 👍

Issue #3053, opened 2017-05-22
update pyramid's default execution policy to attempt to render any uncaught exception


mmerickel:
If an exception propagates all the way to the execution policy there should be a last-ditch effort to render a response... this is part of the larger consistency change to pyramid_tm and pyramid_retry. With the addition of `request.invoke_exception_view` it's hopefully clear that exception views should be callable from various parts of the lifecycle and pyramid_retry will attempt to invoke exception views at the end if an exception is uncaught. Also pyramid_tm will attempt to invoke exceptions on commit/abort to handle those specifically.

I've documented the process here in a little ascii diagram[1] reproduced here:
```
   pyramid_retry   +-------->    if exc not retryable, render response
                                 if exc retryable, ignore exc and try again
        +
        |
        |
        v

pyramid_debugtoolbar

        +
        |
        |
        v

    pyramid_tm     +-------->    if exc caught then abort and reraise
                                 if request.exception then abort and return response
        +                        if abort or commit raises, render response
        |
        |
        v

     excview       +-------->    if exc caught, render response and set request.exception
                                                if no exception then reraise
        +
        |
        |
        v

       app
```

[1] https://gist.github.com/mmerickel/2e83f4af6c7d5eae34947b8f1f075f61

bertjwregeer:
```
    def test_it_supports_alternate_requests(self):
        def exc_view(exc, request):
            self.assertTrue(request is other_req)

# Next two lines are new
            from pyramid.threadlocal import get_current_request
            self.assertTrue(get_current_request() is other_req)

            return DummyResponse(b'foo')
        self.config.add_view(exc_view, context=RuntimeError)
        request = self._makeOne()
        other_req = self._makeOne()
        try:
            raise RuntimeError
        except RuntimeError:
            response = request.invoke_exception_view(request=other_req)
            self.assertEqual(response.app_iter, [b'foo'])
        else: # pragma: no cover
            self.fail()
```

This test, changed from the original in `test_view.py` will fail:

```
======================================================================
FAIL: test_it_supports_alternate_requests (pyramid.tests.test_view.TestViewMethodsMixin)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/xistence/Projects/pyramid/pyramid/tests/test_view.py", line 800, in test_it_supports_alternate_requests
    raise RuntimeError
RuntimeError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #3060, opened 2017-05-23
When invoking an exception view, push the new threadlocals

bertjwregeer:
This way when calling the threadlocal get_current_request() you get the
same request object as the one that was passed to the view.

Fixes the comment on: https://github.com/Pylons/pyramid/issues/3053#issuecomment-303520185

### DIFF
--- a/pyramid/view.py
+++ b/pyramid/view.py
@@ -700,10 +700,9 @@ class ViewMethodsMixin(object):
             # https://github.com/Pylons/pyramid/issues/700
             request_iface = attrs.get('request_iface', IRequest)
 
-            try:
-                if request is not self:
-                    manager.push({'request': request, 'registry': registry})
+            manager.push({'request': request, 'registry': registry})
 
+            try:
                 response = _call_view(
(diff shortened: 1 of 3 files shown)


pyramid/view.py
```diff
@@ -690,17 +699,23 @@ def invoke_exception_view(
...
-                context_iface,
-                '',
-                view_types=None,
-                view_classifier=IExceptionViewClassifier,
-                secure=secure,
-                request_iface=request_iface.combined,
-                )
+
+            try:
+                manager.push({'request': request, 'registry': registry})
```
mmerickel: This `push` should go before the `try`. You do not want to try to `pop` if the `push` fails. It's pedantic but could avoid raising an exception from an exception.

bertjwregeer: Gotcha. Will make the change.

mmerickel: *reminder bump*

bertjwregeer: Done.

PR #3063, opened 2017-05-24
backport contributors

stevepiercy:


tseaver:
Ideally we would *not* be backporting on this file:  tying the commit adding the name to the account "signing" it is the purpose of having those entries.

stevepiercy:
Yeah, I couldn't figure out a clean way to backport this due to merge conflicts. I could revert these changes, but that might cause more of a mess. Not sure what to do in this case.

PR #3064, opened 2017-05-24
Increase Professionalism

amoliski:
"Drops turds" isn't something I want to see in technical documentation.

### DIFF
--- a/docs/narr/install.rst
+++ b/docs/narr/install.rst
@@ -206,7 +206,7 @@ After installing Python as described previously in :ref:`for-mac-os-x-users` or
    ``$VENV/bin/pip`` clearly specifies that ``pip`` is run from within the
    virtual environment and not at the system level.
 
-   ``activate`` drops turds into the user's shell environment, leaving them
+   ``activate`` makes changes to the user's shell environment, leaving them
    vulnerable to executing commands in the wrong context. ``deactivate`` might
    not correctly restore previous shell environment variables.
 


chrisrossi:
I see no need for editorializing.  LGTM as-is.

amoliski:
Is there a technical definition of a 'turd' when it comes to environmental variables, or is it the author attempting to be cute? I think cuteness and jokes should be saved for articles written about how to use a tool, not in the documentation of the tool itself.

Will users who don't speak English as their primary language have to go look up what 'turd' is supposed to mean when they come across an explanation that tells them that they are left in their environments if they follow the previously accepted way of doing things?

chrisrossi:
Just to clarify, I'm saying that @amoliski's change looks good, as-is.  I don't agree with @stevepiercy that we need to pass moral judgement on virtuaulenv's "activate" feature in our documentation, regardless of the language used.

tseaver:
People routinely ask why we don't recommend `activate`.  Part of the point of the language here *is* to cast stones at the effects of `activate`:  it isn't just that it changes the environment, but that the changes are hidden, hard to understand, and tend to break things for the user in hard-to-predict ways.

Themanwithoutaplan:
I understand the reservations about `activate` but have been using it on a day-to-day basis for several years now and I find that it reduces unwanted typos. Sticking with one shell per project significantly reduces the risk of side-effects. Shell experts might decry it but it certainly makes tutorials easier to work with and hectoring users is often counter-productive. So +1 from me.

mmerickel:
If I end up being the one that merges this I'll be changing the text myself to something like "``activate`` makes changes to the user's shell environment - and in the context of a tutorial this can easily be forgotten - leading to executing commands in the wrong context."

Anyway I'm also fine with these changes but this PR is against 1.8-branch and we also need to update master.

stevepiercy:
I'd be fine with something that clearly warns the user of what the changes actually are in understandable terms or how it could negatively impact their experience, and does not offend the professional-minded.

mmerickel:
I rephrased the paragraph in my own words. Thank you for starting the discussion @amoliski.

PR #3067, opened 2017-06-01
clarify badges for releasing

stevepiercy:


mmerickel:
Since we are in mid-release cycle right now master is actually 1.9-branch. We'll fix master to point to master again after we freeze / officially release 1.9-branch.

stevepiercy:
OK, I think I got it right this time. ping @mmerickel

PR #3074, opened 2017-06-05
Ensure that instances of 'AllPermissionsList' are iterable.

### LINKED ISSUE #3073: `AllPermissionsList.__iter__` returns tuple(). Should return iter(tuple())?
My understanding is that __iter__ should [return an iterator object](https://docs.python.org/3.5/library/stdtypes.html#container.__iter__).

However, [`AllPermissionsList.__iter__`](https://github.com/Pylons/pyramid/blob/master/pyramid/security.py#L25) returns `()`.

Indeed, this raises a TypeError as expected:
```
In [1]: from pyramid.security import ALL_PERMISSIONS

In [2]: iter(ALL_PERMISSIONS)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-22-f8a3d5a1d337> in <module>()
----> 1 iter(ALL_PERMISSIONS)

TypeError: iter() returned non-iterator of type 'tuple'
```

Lastly, I don't see where this method is used in Pyramid. Should this code (the `__iter__` method on `AllPermissionsList`) even exist?

tseaver:
Closes #3073.

### DIFF
--- a/pyramid/security.py
+++ b/pyramid/security.py
@@ -21,10 +21,13 @@ _marker = object()
 
 class AllPermissionsList(object):
     """ Stand in 'permission list' to represent all permissions """
+
     def __iter__(self):
-        return ()
+        return iter(())
+
     def __contains__(self, other):
         return True
+
     def __eq__(self, other):
         return isinstance(other, self.__class__)
 
--- a/pyramid/tests/test_security.py
+++ b/pyramid/tests/test_security.py
@@ -16,11 +16,31 @@ class TestAllPermissionsList(unittest.TestCase):
     def _makeOne(self):
         return self._getTargetClass()()
 
-    def test_it(self):
+    def test_equality_w_self(self):
         thing = self._makeOne()
         self.assertTrue(thing.__eq__(thing))
-        self.assertEqual(thing.__iter__(), ())
+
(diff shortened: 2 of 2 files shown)


mmerickel:
The permissions in the `ACLAuthorizationPolicy` are allowed to be iterables of permissions. This object is clearly just trying to support that contract such that the object looks like an iterable even if it doesn't define any permissions itself.

The `__contains__` is by far the most important method in the policy and is used in https://github.com/Pylons/pyramid/blob/cc8ce57839d26e82466df9f74c79281450bf18a5/pyramid/authorization.py#L131.

Arguably the object itself could just override `__eq__` and the policy would wrap it in a list and iterate with `==`. I'd argue that the `__contains__` + `__iter__` is a micro-optimization here. Would you agree @tseaver or do you think the object being iterable is more important? The exposed API is actually `ALL_PERMISSIONS` which doesn't need to be a "list" at all I think. The class could just as easily be called `AllPermissions`.

I think this PR is the correct fix if we agree that dropping `__contains__` and `__iter__` entirely is too much of an incompatibility.

mmerickel:
It seems no one wants to talk about that. Ah well.

PR #3076, opened 2017-06-05
Ensure that instances of 'AllPermissionsList' are iterable.

### LINKED ISSUE #3073: `AllPermissionsList.__iter__` returns tuple(). Should return iter(tuple())?

tseaver:
Cherry-pick 1814cda1f5ec85c24651bfd29645e4057ae5200e from master.

Closes #3073.

### DIFF
--- a/pyramid/security.py
+++ b/pyramid/security.py
@@ -21,10 +21,13 @@ _marker = object()
 
 class AllPermissionsList(object):
     """ Stand in 'permission list' to represent all permissions """
+
     def __iter__(self):
-        return ()
+        return iter(())
+
     def __contains__(self, other):
         return True
+
     def __eq__(self, other):
         return isinstance(other, self.__class__)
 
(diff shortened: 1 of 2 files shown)


mmerickel:
This probably also needs to be backported to 1.7-branch.

Issue #3066, opened 2017-06-01
Improve forbidden_view for BasicAuthAuthenticationPolicy


vog:
The recommended `forbidden_view` of `BasicAuthAuthenticationPolicy` is slightly flawed:

* http://docs.pylonsproject.org/projects/pyramid/en/latest/api/authentication.html#pyramid.authentication.BasicAuthAuthenticationPolicy

The manual currently recommends:

```python
from pyramid.httpexceptions import HTTPUnauthorized
from pyramid.security import forget
from pyramid.view import forbidden_view_config

@forbidden_view_config()
def basic_challenge(request):
    response = HTTPUnauthorized()
    response.headers.update(forget(request))
    return response
```

This implementation has the huge drawback that it does not distinguish between authentication issues (nonexisting user, wrong password) and authorization issues (missing permissions). In particular, if a user is successfully logged in via HTTP Auth and clicks on a forbidden resource, the user will be asked to re-login again and again. That doesn't make any sense because the user can login as often as they want - nothing will ever change as the resource is forbidden to the user.

It makes more sense to ask for login (401) only on authentication failure (nonexisting user, wrong password), while an authorization failure (missing permissions) should lead to a plain classic "forbidden" page (403).

I propose the following `forbidden_view` instead:

```python
from pyramid.httpexceptions import HTTPForbidden, HTTPUnauthorized
from pyramid.security import forget
from pyramid.view import forbidden_view_config

@forbidden_view_config()
def forbidden_view(request):
    if request.authenticated_userid is None:
        response = HTTPUnauthorized()
        response.headers.update(forget(request))
        return response
    return HTTPForbidden()
```

Some more ideas:

* The cookbook should mention this forbidden_view as well: https://pyramid-cookbook.readthedocs.io/en/latest/auth/basic.html
* Since this view is a bit more involved, doesn't it make sense to provide that as a helper function by the framework, instead of having it just as example code in the docs?
  * Perhaps that helper could be automatically installed when BasicAuthAuthenticationPolicy in set in the config?

mmerickel:
I'd be very happy if you wanted to add this example to the cookbook as well as the `BasicAuthAuthenticationPolicy` docs. However, adding a mechanism to automatically register this view is a -1 from me... the example is dead simple but most apps want to customize these error views in some way and are unlikely to use this default implementation so the effort to register it automatically would not be worthwhile.

vog:
[...]

PR #3079, opened 2017-06-09
Fix forbidden_view for BasicAuthAuthenticationPolicy (#3066)

vog:
See also https://github.com/Pylons/pyramid_cookbook/pull/183


mmerickel:
Looks good, thank you!

PR #3080, opened 2017-06-09
Proposal for optional split between HTTPForbidden and HTTPUnauthorized (#3066)

vog:
Add optional use_http_unauthorized_exception to split between HTTPForbidden and HTTPUnauthorized.

Discussion: https://github.com/Pylons/pyramid/issues/3066#issuecomment-307409096

Note that the documentation adjustment in this pull request builds upon the documentation corrections provided #3079.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -302,3 +302,5 @@ Contributors
 - Jeremy(Ching-Rui) Chen, 2017/04/19
 
 - Fang-Pen Lin, 2017/05/22
+
+- Volker Diels-Grabsch, 2017/06/09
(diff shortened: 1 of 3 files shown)


mmerickel:
I'm sorry but similar to my thoughts in https://github.com/Pylons/pyramid/issues/3066#issuecomment-306242977 this simply isn't going to fly. There are several technical reasons aside from simply the code quality of this specific PR. 

Primarily the 401 Unauthorized HTTP response modeled by the `HTTPUnauthorized` exception **requires** a challenge. It is not optional per the HTTP standard. This makes it a bad object to use for a generic authorization error within the framework. Most authentication systems do not supply an HTTP challenge specification and that makes 401 almost worthless aside from use in HTTP Basic Auth. Since you already need to define a forbidden view, you can easily determine if you should issue a challenge in this scenario yourself without relying on the policy / system to do it for you. If we were fully supporting an `HTTPUnauthorized` exception raised by our system it would need to support some sort of challenge model such that the policy could communicate that to the user. As of now that's a larger change than a simple boolean flag.

Issue #3087, opened 2017-06-15
Add quote_via keyword argument to urlencode


drnextgis:
How about adding `quote_via` keyword argument to `pyramid.encode.urlencode` to allow the caller to specify the encoding function as it was made in Python 3? See [#13866](https://bugs.python.org/issue13866). Something like [this](https://github.com/drnextgis/pyramid/commit/4e7a2eba304567c004b4f059740285928939d1d1):

```
>>> from pyramid.encode import url_quote, urlencode
>>> urlencode({'k': 'some value'})
'k=some+value'
>>> urlencode({'k': 'some value'}, quote_via=url_quote)
'k=some%20value'
```

mmerickel:
Thanks @drnextgis. Would you like to provide your patch in the form of a PR with your name in CONTRIBUTORS.txt?

mmerickel:
Closed via #3088.

PR #3088, opened 2017-06-15
quote_via urlencode argument

drnextgis:
See https://github.com/Pylons/pyramid/issues/3087


mmerickel:
The failed checks here are because of some weird thing happening when I pushed a pr/3088 branch to the repo and then deleted it. Just going to ignore them for now.

Thanks @drnextgis you can see I added some docs and tests.

PR #3097, opened 2017-06-20
Make DummyBootstrap return a context manager

### LINKED ISSUE #3096: DummyBootstrap doesn't support context manager mode
Since #1822 bootstrap can return a context manager, but DummyBootstrap hasn't been converted to support the new mode.

corehack:
Closes #3096
Currently no scripts use context manager mode, and there was a missing coverage, so I converted ptweens to use the context manager mode

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -308,3 +308,5 @@ Contributors
 - Volker Diels-Grabsch, 2017/06/09
 
 - Denis Rykov, 2017/06/15
+
+- Andrey Tretyakov, 2017/06/20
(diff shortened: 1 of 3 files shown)


mmerickel:
This doesn't seem to cover an actual use-case. Where are the usage docs? Is someone supposed to import `DummyBootstrap` instead of `pyramid.paster.boostrap`? This API is not written in a way that's actually injectable into an app that's currently using bootstrap. The other fixtures that pyramid provides do not require monkey patching... they are objects that you just pass to functions that want them. Bootstrap does not work in this way.

corehack:
If it's useless than why is Pyramid using it to test own scripts  huh ? Actually that's how I discovered it because I thought testing my scripts the way Pyramid tests it would be a good place to start.
https://github.com/Pylons/pyramid/blob/master/pyramid/tests/test_scripts/test_ptweens.py

Nvm, just passing by

PR #3120, opened 2017-07-07
Ignore here and __file__ for pshell configuration

invisibleroads:
If the folder containing the configuration file (`development.ini`) is not a package, then `pshell` fails with `ImportError` because `resolver.maybe_resolve` tries to import `__file__` and `here`.

This pull request tells `pshell` to ignore `__file__` and `here`.

### DIFF
--- a/pyramid/scripts/pshell.py
+++ b/pyramid/scripts/pshell.py
@@ -101,6 +101,8 @@ class PShellCommand(object):
                 self.setup = v
             elif k == 'default_shell':
                 self.preferred_shells = [x.lower() for x in aslist(v)]
+            elif k in ('__file__', 'here'):
+                continue
             else:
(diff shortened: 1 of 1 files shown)


mmerickel:
This looks good but obviously needs a test case. Also please add your name to CONTRIBUTORS.txt.

mmerickel:
I think this probably needs to be solved by plaster_pastedeploy which is inadvertently inserting extra values into the settings dict that didn't used to be there.

mmerickel:
Also I confirmed this is not a problem without plaster_pastedeploy (such as on pyramid < 1.9).

mmerickel:
This should be fixed by upgrading to plaster_pastedeploy 0.4.

invisibleroads:
Okay, I'll try upgrading `plaster_pastedeploy` and tell you if it fixes the issue.

PR #3134, opened 2017-07-21
remove dead link

adamcheasley:
Looks like plope.com is now dark.

### DIFF
--- a/docs/quick_tutorial/routing.rst
+++ b/docs/quick_tutorial/routing.rst
@@ -119,6 +119,3 @@ Extra credit
 
 #. What happens if you to go the URL http://localhost:6543/howdy? Is this the
    result that you expected?
-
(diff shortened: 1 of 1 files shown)


stevepiercy:
@mcdonc plope.com is not responding. What's up with it?

tseaver:
I'm in the process of decommissioning its host, so it should be moving shortly. It may have crashed while I was poking around.  Can we hold off on merging this until Chris and I have a chance to confer?

mmerickel:
I think it's actually been down for quite a while. I had assumed that Chris just killed it already. I've tried to open the posts often in the last year and only remember tcp connection errors.

mmerickel:
Can we just replace the dead plope.com links with webarchive? https://web.archive.org/web/20170131192830/http://www.plope.com/weird_pyramid_urldispatch

mmerickel:
Fixed in 681a79f56.

PR #3140, opened 2017-08-03
Drop repoze.lru on Python 3

bochecha:
Starting with Python 3.2, the functools module grew a lru_cache function which can replace our usage of repoze.lru.

### DIFF
--- a/pyramid/url.py
+++ b/pyramid/url.py
@@ -2,8 +2,6 @@
 
 import os
 
-from repoze.lru import lru_cache
-
 from pyramid.interfaces import (
     IResourceURL,
     IRoutesMapper,
@@ -12,6 +10,7 @@ from pyramid.interfaces import (
 
 from pyramid.compat import (
     bytes_,
+    lru_cache,
     string_types,
     )
 from pyramid.encode import (
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,8 @@
 #
 ##############################################################################
 
+import sys
+
 from setuptools import setup, find_packages
 
 def readfile(name):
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -35,6 +36,9 @@ def readfile(name):
     'hupper',
     ]
 
+if sys.version_info < (3, 2):
```
bertjwregeer: This should be an `install_requires` tag instead, not a dynamic change, this way it ends up correctly in the wheels and things work as expected.

pyramid/url.py
```diff
@@ -2,7 +2,11 @@
 
 import os
 
-from repoze.lru import lru_cache
+try:
```
bertjwregeer: Move this to `compat.py` and import from `compat.py` instead of repeating this try/importerror everywhere.

mmerickel:
Thanks! One of the maintainers will need to do some legwork to determine if this library actually covers the required features or not.

Environment markers are the correct way to handle conditional dependencies right now. For example: https://github.com/Pylons/pyramid_debugtoolbar/blob/7c32a1fc909e182f2f26ed5cef680784ffde6263/setup.py#L77

bochecha:
> One of the maintainers will need to do some legwork to determine if this library actually covers the required features or not.

They both implement the same thing (an LRU cache for function results), but indeed have subtle differences in behaviour, which would be better checked by a maintainer.

One such difference that I can see from a quick glance at the repoze.lru code is that it seems to allow expiring things after a certain time (with its `timeout` kwarg) which `functools.lru_cache` doesn't support.

`functools.lru_cache` also supports optionally caching results by type, but not `repoze.lru.lru_cache`.

However the subset of the API that is used in the Pyramid code base seems to work similarly in both cases.

They also handle differently a `maxsize` argument set to `0` or `None`, but that isn't used in Pyramid either.

mmerickel:
Thanks @bochecha, I reviewed the two codebases and could not see a compelling reason not to move forward with `functools.lru_cache`.

PR #3146, opened 2017-08-10
Fix file extension typo

larsblumberg:
``myproject:templates/blog.pt`` –> ``myproject:templates/blog.py``


tseaver:
@larsblumberg Thanks for the patch!  However, the "asset specification" is actually a reference to a template file, not a module, and so the existing `.pt` extension is correct.

stevepiercy:
This is not a typo. Template files typically have a filename extension that corresponds to the templating engine. Chameleon typically uses `.pt`, Jinja2 uses `.jinja2`, and Mako uses `.mako`.

larsblumberg:
Thank you guys for your explanation!

PR #3152, opened 2017-08-14
Add a functional test for the static file

larsblumberg:
Static files were introduced in the quick tutorial step "[static assets](https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/static_assets.html)". As the tutorials emphasise to write tests, this PR adds a test for the static file `app.css` that was introduced with this tutorial step. This PR also updates the according tutorial page.

### DIFF
--- a/docs/quick_tutorial/static_assets.rst
+++ b/docs/quick_tutorial/static_assets.rst
@@ -43,13 +43,18 @@ Steps
    .. literalinclude:: static_assets/tutorial/static/app.css
     :language: css
 
-#. Make sure we haven't broken any existing code by running the tests:
+#. We add a functional test that asserts that the newly added static file is delivered:
+
+   .. literalinclude:: static_assets/tutorial/tests.py
+    :linenos:
+
+#. Now run the tests:
 
    .. code-block:: bash
 
     $ $VENV/bin/py.test tutorial/tests.py -q
     ....
-    4 passed in 0.50 seconds
+    5 passed in 0.50 seconds
 
 #. Run your Pyramid application with:
 
--- a/docs/quick_tutorial/static_assets.rst
+++ b/docs/quick_tutorial/static_assets.rst
@@ -46,7 +46,9 @@ Steps
 #. We add a functional test that asserts that the newly added static file is delivered:
 
    .. literalinclude:: static_assets/tutorial/tests.py
-    :linenos:
+    :language: python
+    :pyobject: TutorialFunctionalTests.test_css
+    :lineno-match:
 
 #. Now run the tests:
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -310,3 +310,6 @@ Contributors
 - Denis Rykov, 2017/06/15
 
 - Tosh Lyons, 2017/06/27
+
+- Lars Blumberg, 2017/08/14
+
--- a/docs/quick_tutorial/static_assets/tutorial/tests.py
+++ b/docs/quick_tutorial/static_assets/tutorial/tests.py
@@ -42,3 +42,7 @@ class TutorialFunctionalTests(unittest.TestCase):
     def test_hello(self):
         res = self.testapp.get('/howdy', status=200)
         self.assertIn(b'<h1>Hi Hello View', res.body)
+
+    def test_css(self):
+        res = self.testapp.get('/static/app.css', status=200)
+        self.assertIn(b'body', res.body)


docs/quick_tutorial/static_assets.rst
```diff
@@ -43,13 +43,18 @@ Steps
    .. literalinclude:: static_assets/tutorial/static/app.css
     :language: css
 
-#. Make sure we haven't broken any existing code by running the tests:
+#. Add a functional test that requires the newly added static file to be delivered:
+
+   .. literalinclude:: static_assets/tutorial/tests.py
+    :linenos:
```
stevepiercy: Instead of including the entire file, please change this to the following:

```
   .. literalinclude:: json/tutorial/tests.py
    :pyobject: TutorialFunctionalTests. test_css
    :lineno-match:
```

stevepiercy:
Nice! Three points:

1. This contribution is more than just a trivial one like the others you recently made, and as such would you please sign `CONTRIBUTORS.txt`, commit, and push to this PR? It's needed only one time.
2. This change will affect subsequent steps in the Quick Tutorial, so this change must be carried forward in those steps, too. The accompanying narrative files might also need to be updated. Please ensure that the source and narrative files maintain consistency from Step 13 onward.
3. See code review.

docs/quick_tutorial/static_assets.rst
larsblumberg: All tutorial pages print the complete file, even if only small modifications are made in respect to previous versions of the same file. This said, do you suggest to break with this patterns and only show a fragment of `tests.py` here?

larsblumberg: I've updated the wording a little bit to include "We" as that's the writing style of all tutorial pages.

larsblumberg:
Thanks for your comments.
1. I have signed the contributors file as part of this PR
2. I have checked all subsequent steps of the Quick Tutorial. No other step reuses step 13 making it unnecessary to propagate the change that is introduced by this PR.
3. I have replied to your code review.

docs/quick_tutorial/static_assets.rst
stevepiercy: Quick Tutorial has been neglected. In other tutorials, we only recently started to implement `:lineno-match:` and `:pyobject:` (and other options) to replace the somewhat lazy `:linenos:` along with the more difficult to maintain and problematic `:lines:` and `:lineno-start:`. This would be a start for Quick Tutorial.

In this case, the line numbers tell the user where to place the snippet, which they can still copy-pasta.

Also the source files are available in the repository, If the user wants to just drag and drop or copy-pasta from them. I just realized that we don't advertise that in this tutorial, but we do in others, which is an oversight. I've created a new issue #3154 to track that. If you're feeling ambitious, you're welcome to it!

larsblumberg: Understood. Do you suggest that this PR introduces the usage of `:pyobject:` instead of ` :linenos:`? I'm happy to do that.

stevepiercy: When it is appropriate and an obvious improvement, yes indeed, but not just for the sake of changing syntax. Sometimes we do want to include the entire file (e.g., "Create a new file and edit it as follows"). Other times we want to use `:emphasize-lines:` with `:lines:` and `:lineno-match:` such as when it's too much to include the entire `:pyobject:`. Holler if you have any doubts or questions. And thank you!

stevepiercy: Oh, if you meant for this single instance, yes, please. I answered with the assumption toward wholesale replacement.

larsblumberg: Now only the added lines are included in this tutorial page: larsblumberg@7b3dd65

I'm happy to volunteer updating all quick tutorials steps. We however want to keep _this_ PR focussed on one single change (adding a test case for the added static file).

docs/quick_tutorial/static_assets.rst
```diff
@@ -43,13 +43,21 @@ Steps
...
     :language: css
 
-#. Make sure we haven't broken any existing code by running the tests:
+#. We add a functional test that asserts that the newly added static file is delivered:
+
+   .. literalinclude:: static_assets/tutorial/tests.py
+    :lines: 46-48
+    :lineno-match:
+    :linenos:
+    :language: python
```
stevepiercy: Sorry for the confusion. This should be:

```
.. literalinclude:: static_assets/tutorial/tests.py
    :language: python
    :pyobject: TutorialFunctionalTests.test_css
    :lineno-match:
```

`:lineno-match:` should replace `:lines:` and `:linenos:`.

larsblumberg: Pardon my ignorance. This is what you already write before. I addressed it in larsblumberg@c6ede2d

stevepiercy: No worries, I goofed, too, omitting the language and using the wrong path. Teamwork! I'll merge this as soon as the CI passes.

stevepiercy:
Thank you!

PR #3178, opened 2017-10-19
Update requirements.rst

joray:


stevepiercy:
This change is unnecessary, [as far as I can tell](https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/requirements.html#create-a-project-directory-structure). Why did you add the backticks?

PR #3183, opened 2017-10-22
fix pypy by using a specific version

mmerickel:


mmerickel:
```
======================================================================
ERROR: test_instance_conforms (pyramid.tests.test_events.TestBeforeRender)
----------------------------------------------------------------------
 File "/home/travis/build/Pylons/pyramid/pyramid/tests/test_events.py", line 242, in test_instance_conforms
 verifyObject(IBeforeRender, event)
 File "/home/travis/build/Pylons/pyramid/.tox/pypy/site-packages/zope/interface/verify.py", line 105, in verifyObject
 return _verify(iface, candidate, tentative, vtype='o')
 File "/home/travis/build/Pylons/pyramid/.tox/pypy/site-packages/zope/interface/verify.py", line 97, in _verify
 raise BrokenMethodImplementation(name, mess)
BrokenMethodImplementation: The implementation of pop violates its contract
 because implementation requires too many arguments.
```

mmerickel:
Ok I bisected pypy versions and found that the tests pass on 5.4.1 and fail on 5.6.0.

mmerickel:
tracking issue: https://bitbucket.org/pypy/pypy/issues/2685/regression-with-pyramid-zopeinterface

tseaver:
@mmerickel on the PyPy issue, you say we have a bug in `IDict` -- I can't find that bug here, or a fix for it.  Ideally, we would fix the bug and then un-pin the PyPy version in `.travis.yml`.

mmerickel:
@tseaver did you see the explanation at https://bitbucket.org/pypy/pypy/issues/2685/regression-with-pyramid-zopeinterface#comment-40776650 ? I have not gotten around to fixing this yet.

tseaver:
@mmerickel I didn't expect that you had fixed it:  I just wondered if I had missed where you opened the bug against *Pyramid* (since you agreed on the Bitbucket issue that we have a bug).

mmerickel:
thanks @tseaver I opened https://github.com/Pylons/pyramid/issues/3237 to track it now.

bertjwregeer:
I may not be understanding the issue correctly, but I am not sure how using `marker` would change anything since it's z.i that needs to understand about the non-normal dict.pop.

mmerickel:
My comment in that issue was about how I would fix a dict implementation to conform to the contract in response to the comment that it cannot be described in python. It was not a proposed fix for getting z.i to broadcast the situation correctly. I don’t have a fix at this time unless taking `*args, **kwargs` works to satisfy the check because the arg could be passed by either method.

PR #3192, opened 2017-10-28
Synch source files with cookiecutter

stevepiercy:


stevepiercy:
RE: https://github.com/Pylons/pyramid-cookiecutter-alchemy/pull/24

stevepiercy:
Anyone know how to fix the builds for this PR and #3193 and #3194, or should I just ignore and merge them?

stevepiercy:
Per IRC, @bertjwregeer suggested that a new version of flake8 was the culprit, and I found the release notes that matched the new error. See:
* http://flake8.pycqa.org/en/latest/release-notes/3.5.0.html#new-dependency-information
* https://gitlab.com/pycqa/flake8/issues/361

Should I insert an exclusion for E722?
https://github.com/Pylons/pyramid/blob/master/setup.cfg#L59

Or should the code be updated?

bertjwregeer:
I need to take a closer look, but likely we should update the code.

PR #3196, opened 2017-11-03
views.py: prevent exception on unknown user login

silum:
Attempting a login without specifying the user login, or when the user login is not known, i.e. not 'editor' nor 'viewer', an unhandled exception is raised in `security.py` because `None` is passed to `check_password()` as the hashed password to check against.

This patch checks for a non-None hashed password before calling `check_password()`.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -313,3 +313,4 @@ Contributors
 
 - Lars Blumberg, 2017/08/14
 
+- Deneys Maartens, 2017/11/03
--- a/docs/quick_tutorial/authentication/tutorial/views.py
+++ b/docs/quick_tutorial/authentication/tutorial/views.py
@@ -43,7 +43,8 @@ class TutorialViews:
         if 'form.submitted' in request.params:
             login = request.params['login']
             password = request.params['password']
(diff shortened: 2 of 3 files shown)


stevepiercy:
Thank you! This looks good, but before I merge:

1. @silum would you please sign contributors.txt in the repo and push that commit to this PR? I'm kinda on the fence about whether this change is "material", but I'd rather be safe.
2. The change also needs to be applied to the next step in the tutorial on the source file `docs/quick_tutorial/authorization/tutorial/views.py`
3. The narrative docs need to checked to make sure that any references in wording or line numbers still align.

@silum you are welcome to do items 2 and 3 in your PR, or I can take them on later today.

I also need to backport, but that's all up to me.

silum:
I hadn't gotten as far as the authorisation section yet, so hadn't noticed that it also needs the patch.  :)
I had a quick read-through of the relevant docs and didn't notice line-numbers that would need changing; it couldn't hurt to double-check though.

stevepiercy:
Thank you for the follow through!

silum:
You are most welcome.

tseaver:
This PR is missing an entry in `CHANGES.txt`.

PR #3203, opened 2017-11-15
Clarify property descriptor in add_request_method

### LINKED ISSUE #3202: add_request_method only accepts property descriptor that subclasses property
https://github.com/Pylons/pyramid/blob/cc6a6ed60260b0391aa38f372b26311d58edf0d6/pyramid/util.py#L70

Looks like we check the type of a property descriptor by `isinstance(callable, property)`. This is not true for property descriptors subclasses from `object` which is pretty common. Maybe we should document this behavior [here](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.add_request_method).

tjwudi:
Fix #3202 - Doc change.

### DIFF
--- a/pyramid/config/factories.py
+++ b/pyramid/config/factories.py
@@ -144,9 +144,10 @@ class FactoriesConfiguratorMixin(object):
 
         When adding a property to the request, ``callable`` can either
         be a callable that accepts the request as its single positional
-        parameter, or it can be a property descriptor. If ``name`` is
-        ``None``, the name of the property will be computed from the
-        name of the ``callable``.
+        parameter, or it can be a property descriptor. If ``callable`` is
+        a property descriptor, it has to be a instance of a class which is
+        a subclass of ``property``. If ``name`` is ``None``, the name of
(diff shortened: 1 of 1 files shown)


pyramid/config/factories.py
```diff
@@ -144,9 +144,10 @@ def add_request_method(self,
 
         When adding a property to the request, ``callable`` can either
         be a callable that accepts the request as its single positional
-        parameter, or it can be a property descriptor. If ``name`` is
-        ``None``, the name of the property will be computed from the
-        name of the ``callable``.
+        parameter, or it can be a property descriptor. If ``callable`` is
+        a property descriptor, it has to be a instance of a class which is
```
stevepiercy: Grammar fix:

an instance

tjwudi:
@stevepiercy @tseaver   updated =)

stevepiercy:
Should this be backported to 1.9-branch or 1.8-branch?

tjwudi:
I think so but am not quite sure how we operate that.

PR #3204, opened 2017-11-25
setup.py: add waitress requirement

silum:
This adds a waitress requirement missed out by https://github.com/Pylons/pyramid/commit/cfc8fbe0e1380c06b01643f8310ae59ea0af820b.

### DIFF
--- a/docs/quick_tutorial/package/setup.py
+++ b/docs/quick_tutorial/package/setup.py
@@ -2,6 +2,7 @@ from setuptools import setup
 
 requires = [
     'pyramid',
+    'waitress',
 ]
 
 setup(name='tutorial',


tseaver:
Thanks for the patch, @silum!

stevepiercy:
Sorry, I've reverted this change.

This change is not necessary if the user followed the [installation step under Requirements](https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/requirements.html#install-pyramid).

Also, in general, changes to tutorials should not be merged until subsequent steps in the tutorial that depend on a given step are reviewed, and this change would effect later steps.

Sequential/step-dependent maintenance is a chore. If it is a bug fix or other substantial improvement, then it's worth doing the work.

silum:
@stevepiercy, in principle I agree.  However, the following step (step 3) adds `waitress` to `requires` without noting it; the narration at this point discusses adding an "entry point".  Most importantly, if one starts from a clean virtual environment at step 2, running `pip install -e .` will not result in a functional installation.

stevepiercy:
@silum thanks for pointing out this oversight. I've made a PR #3208 that should address your point.

It is a common problem that users will jump in to a random step without performing earlier steps, and wonder what is wrong. I'm almost tempted to slap a banner at the top of every file in this tutorial as a warning, but users who tend to jump in without reading the documentation from the beginning will tend to skip over such warnings. Silly human behavior. ¯\_(ツ)_/¯

PR #3221, opened 2018-01-24
Emphasize additional line in Subrequest docs

di:
This PR emphasizes an additional line which has been changed between the first and second example of [the Subrequest documentation](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/subrequest.html), specifically:

```
config.add_view(view_two, route_name='two', renderer='string')
```


stevepiercy:
Nice catch. Thank you! I'll backport to 1.9 and 1.8, too.

Issue #3195, opened 2017-10-29
wsgiserver has been moved to cheroot


jardakotesovec:
As explained here https://github.com/cherrypy/cherrypy/issues/1585  `wsgiserver ` has been moved to `cheroot`.

Therefore importing in `https://github.com/Pylons/pyramid/blob/e78aa24cda85368c3507c145e1e604e7335778dc/pyramid/scripts/pserve.py#L344` should probably look like:

```
try:
    from cheroot.wsgi import Server as WSGIServer
except ImportError:
    from cherrypy.wsgiserver import CherryPyWSGIServer as WSGIServer 
```

goodwillcoding:
@jardakotesovec any chance you can submit a PR?

heron182:
Anyone working on this issue ? 
I'd be happy to help

tseaver:
@heron182 Please feel free to add a PR.

mmerickel:
Fixed via #3235.

PR #3235, opened 2018-03-08
Adjusting cherrypy WSGI Server import path according to new release

heron182:
Fixing `cherrypy` WSGI import, ref #3195 

A couple of questions:

* cherrypy_server_runner has a explicit `# pragma: no cover` comment. Should I remove it and add a specif test for it ? 

* CHANGES.txt doesn`t look like is beeing mainted lately. Should I add an entry to it ?


heron182:
Fixing formatting and `tox` failed test.

pyramid/scripts/pserve.py
```diff
@@ -341,18 +341,21 @@ def cherrypy_server_runner(
...
         if PY2:
             server.ssl_certificate = server.ssl_private_key = ssl_pem
         else:
             # creates wsgiserver.ssl_builtin as side-effect
-            wsgiserver.get_ssl_adapter_class()
-            server.ssl_adapter = wsgiserver.ssl_builtin.BuiltinSSLAdapter(
-                ssl_pem, ssl_pem)
+            from cherrypy.wsgiserver import get_ssl_adapter_class, ssl_builtin
+            get_ssl_adapter_class()
+            server.ssl_adapter = ssl_builtin.BuiltinSSLAdapter(ssl_pem, ssl_pem)
```
tseaver: This part needs to use `cheroot`, too.  E.g.:

```python
try:
    from cheroot.server import get_ssl_adapter_class
except ImportError:
    from cherrypy.wsgiserver import get_ssl_adapter_class

try:
    from cheroot.ssl.builtin import BuiltinSSLAdapter
except ImportError:
    from cherrypy.wsgiserver.ssl_builtin import BuiltinSSLAdapter
```

heron182:
Thanks @tseaver , I updated whar you pointed out :+1: 

As for `CHANGES.txt`, I'ḿ not sure which would be the best section to add it. As its not really related to `Pyramid` itself I don't if it should be considered a *Bug Fix*. Maybe created a new *Backward Compatibilities* section ?
What do you think ?

mmerickel:
A note belongs in features and another in backward incompatibilities mentioning the change in import paths for the cherrypy server runner.

heron182:
Done

tseaver:
@mmerickel I don't think there is a backward-incompatibility here for Pyramid:  we try the `cheroot` imports first, but fall back to the old `cherrypy` ones, which means that even users who have an old `cherrypy` installed will not have to adjust their code (although why they would be installing a new Pyramid and not updating to the lastest `cherrypy` is beyond me).

CHANGES.txt
```diff
@@ -35,6 +39,9 @@ Backward Incompatibilities
   depending on it directly within your project.
   See https://github.com/Pylons/pyramid/pull/3140
 
+- On CherryPy 9+ all functionality from ``cherrypy.wsgiserver`` was moved to ``cheroot``.
+  See https://github.com/Pylons/pyramid/issues/3195
```
tseaver: As I noted, I think we can drop this mention:  users won't need to adjust their code, which means this is not a backward-incompatibility *for Pyramid.*

CHANGES.rst
```diff
@@ -21,6 +21,10 @@ Features
   instead of ``pyramid.util.Request``.
   See https://github.com/Pylons/pyramid/pull/3129
 
+- Added a fallback ``try/except`` block on ``pyramid/scripts/pserve.cherrypy_server_runner``
+  due to module/import changes on CherryPy 9+
```
tseaver: Can we phrase this more clearly as:

```
- In `cherrypy_server_runner`, prefer imports from the `cheroot` package over the legacy
  imports from `cherrypy.wsgiserver`.
```

mmerickel:
@tseaver you're right of course, I missed that the old imports were preserved as a fallback!

bertjwregeer:
> although why they would be installing a new Pyramid and not updating to the lastest cherrypy is beyond me

There was a published version of cheroot that didn't parse query string parameters correctly and dropped them from the WSGI env... so there are plenty of good reasons not to upgrade CherryPy/Cheroot without doing ample verification you actually need the changes.

Issue #3234, opened 2018-03-07
Secret length for SignedCookieSessionFactory


hongyuan1306:
The doc of `SignedCookieSessionFactory` in pyramid.session says:

```
    ``secret``
      A string which is used to sign the cookie. The secret should be at
      least as long as the block size of the selected hash algorithm. For
      ``sha512`` this would mean a 128 bit (64 character) secret.  It should
      be unique within the set of secret values provided to Pyramid for
      its various subsystems (see :ref:`admonishment_against_secret_sharing`).
```

Shouldn't it be 512 bit (64 characters)??

bertjwregeer:
You are correct.

mmerickel:
Fixed in #3244.

PR #3244, opened 2018-03-16
Fix secret length in doc of SignedCookieSessionFactory

hongyuan1306:
see #3234


stevepiercy:
Thanks for the PR. In the future, please avoid white space fixes that don't pertain to the change.

PR #3249, opened 2018-04-04
Add bcrypt pre requirement for chapter 20

lp177:


stevepiercy:
Thanks for the PR. However, this is already covered in https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/authentication.html The Quick Tutorial is progressive and follows a typical development pattern of installing packages as needed, even though the Requirements step has this optional installation statement. If Step 20 fails for the learner because bcrypt is not installed, then they must learn how to configure their `setup.py` and install the package. This process is useful when you distribute your project. I think that bcrypt is the only package that we omit in the Requirements optional install step, partly because it forces this lesson upon the learner to succeed.

I'm -0 on this change, but I'm open to a counterargument.

If persuaded, then the change should also be made for Windows, not just macOS and Linux.

lp177:
If it's voluntary, it seems to me totally correct, I just thought it might have been an oversight because as you say it's the only one missing

bertjwregeer:
I'm a -1 on this change. I like the fact that our tutorials push you to problem solve for yourself and in the process learn. It's like when you learn how to bike, we'll run alongside you and try and catch you, but you still have to pedal.

PR #3253, opened 2018-04-11
updated docs/interfaces to note change from webob 1.7/1.5

jvanasco:
webob 1.5 deprecated 'key' kwarg in favor of 'name'.
webob 1.7 removed 'key'.

This just updates the docs/interfaces to note change from webob 1.7 of `set_cookie(key` to `set_cookie(name`.

after the 1.8 release, i audited my code for any missed deprecations. a search picked up the old name in a tox virtualenv.

### DIFF
--- a/docs/narr/webob.rst
+++ b/docs/narr/webob.rst
@@ -412,7 +412,7 @@ Here are some highlights:
     (you may also use a timedelta object).  The ``Expires`` key will also be
     set based on the value of ``max_age``.
 
-``response.delete_cookie(key, path='/', domain=None)``
+``response.delete_cookie(name, path='/', domain=None)``
     Delete a cookie from the client.  This sets ``max_age`` to 0 and the cookie
     value to ``''``.
 
--- a/docs/narr/webob.rst
+++ b/docs/narr/webob.rst
@@ -406,7 +406,7 @@ Here are some highlights:
     ``response.text``. ``response.content_type_params`` is a dictionary of all
     the parameters.
 
-``response.set_cookie(key, value, max_age=None, path='/', ...)``
+``response.set_cookie(name, value, max_age=None, path='/', ...)``
     Set a cookie.  The keyword arguments control the various cookie parameters.
     The ``max_age`` argument is the length for the cookie to live in seconds
     (you may also use a timedelta object).  The ``Expires`` key will also be
--- a/pyramid/interfaces.py
+++ b/pyramid/interfaces.py
@@ -244,7 +244,7 @@ class IResponse(Interface):
         """ Gets and sets and deletes the Server header. For more information
         on Server see RFC216 section 14.38. """)
 
-    def set_cookie(key, value='', max_age=None, path='/', domain=None,
+    def set_cookie(name, value='', max_age=None, path='/', domain=None,
                    secure=False, httponly=False, comment=None, expires=None,
                    overwrite=False):
         """ Set (add) a cookie for the response """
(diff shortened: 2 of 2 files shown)


bertjwregeer:
Thanks @jvanasco. This should get back ported to 1.8 as well, so I'll leave this open until we get a chance to do that and verify no other usage of key.

stevepiercy:
I'll take care of this today. Thanks for the report!

stevepiercy:
@bertjwregeer:

* When the narrative mentions "key", should I change it to "name"? I'm unsure of the context for "key" in this example:
```
The ``Expires`` key will also be
    set based on the value of ``max_age``.
```
* Does this change also apply to `delete_cookie(key)`?

jvanasco:
It should be `delete_cookie(name)` as it's just setting an empty cookie.
WebOb changed it to `name` as well: https://docs.pylonsproject.org/projects/webob/en/stable/api/response.html#webob.response.Response.delete_cookie

``Expires`` is a key in the cookie's dict (reference: the `max_age` and `expires` kwarg docstrings under https://docs.pylonsproject.org/projects/webob/en/stable/api/response.html#webob.response.Response.set_cookie)

stevepiercy:
@jvanasco would you please update your PR to include `delete_cookie(name)`?

Regarding `Expires`, to confirm, the "key" in this context should remain as "key", and not be changed to "name", correct?

New related question, per https://docs.pylonsproject.org/projects/webob/en/stable/api/response.html#webob.response.Response.set_cookie, I noted that there are two kwargs, `name` and `key`, but the `key` kwarg does not have a description in the docstring. Is that an oversight or deliberate?

jvanasco:
@stevepiercy PR updated.

It is my interpretation of those docs that `key` in relation to `Expires` is a dict key and not the param, so it should not be changed and remain "key". 

Bert will have to answer the related question.

`set_cookie` still accepts the deprecated kwarg (https://github.com/Pylons/webob/blob/81e262a44bc754950fc449553f7bc93769163d8c/src/webob/response.py#L950-L953)

however it doesn't do anything with it...

https://github.com/Pylons/webob/blob/81e262a44bc754950fc449553f7bc93769163d8c/src/webob/response.py#L1042-L1071

I don't know if that was on purpose or by accident.  According to the docs, it shouldn't be accepted.

bertjwregeer:
It's an accident if it is still accepted, and should be removed. It is likely an oversight.

bertjwregeer:
Removed: https://github.com/Pylons/webob/commit/4e8c7ecc20bed6ce6c64daa3dcb97cc328058e8c

stevepiercy:
I found a few more possible instances in Pyramid core (nothing in tests) for review of whether to change "key" to "name":

1. https://github.com/Pylons/pyramid/blob/master/pyramid/interfaces.py#L183
2. https://github.com/Pylons/pyramid/blob/master/pyramid/interfaces.py#L247
3. https://github.com/Pylons/pyramid/blob/master/pyramid/interfaces.py#L262
4. https://github.com/Pylons/pyramid/blob/master/pyramid/csrf.py#L64-L86

Could I get a second pair of eyes on these?

I think it's mostly custodial work, and makes it consistent with WebOb.

jvanasco:
> I found a few more possible instances in Pyramid core (nothing in tests) for review of whether to change "key" to "name”:

> 	• https://github.com/Pylons/pyramid/blob/master/pyramid/interfaces.py#L183

The PR missed this one, but I just updated it


> 	• https://github.com/Pylons/pyramid/blob/master/pyramid/interfaces.py#L247
This  was the one I originally found...

> 	• https://github.com/Pylons/pyramid/blob/master/pyramid/interfaces.py#L262
This one you found above, and I already updated the PR with

> 	• https://github.com/Pylons/pyramid/blob/master/pyramid/csrf.py#L64-L86
I don’t think this is related. I believe this just sets the CSRF key/token into the session dict.

stevepiercy:
Thank you, both!

PR #3278, opened 2018-05-14
Added additional diagram for 01: Single-File Web Applications

LuisCastilloH:
I was reading the documentation and I felt having this diagram would be good to understand how my application is at the moment.

### DIFF
--- a/docs/quick_tutorial/hello_world.rst
+++ b/docs/quick_tutorial/hello_world.rst
@@ -56,6 +56,17 @@ Steps
    .. literalinclude:: hello_world/app.py
     :linenos:
 
+The following tree diagram shows the current structure.
+
+   .. code-block:: text
+
+    `──/path/to/project/
+        `── projects
+            `── quick_tutorial
+                │── env
+                `── hello_world
+                       `── app.py
+
 #. Run the application:
 
    .. code-block:: bash


mmerickel:
I like the idea but I or @stevepiercy needs to check to see whether the proposed structure is actually correct. It seems curious to me that the path wouldn't just be to the `quick_tutorial` folder but maybe that's how the tutorial is written.

stevepiercy:
Although I see the advantage, I'm a -1 on this proposal because the following reasons outweigh the advantage for me:

* I think it adds a distraction, interrupting the flow.
* Adding the tree diagram sets a precedent to add one to each step of the tutorial. That is busy work and not fun to maintain.
* The tree diagram would duplicate these sections, where it is covered from a meta perspective:
  * https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/requirements.html#create-a-project-directory-structure
  * https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/tutorial_approach.html#directory-tree

cewing:
@stevepiercy I understand your reluctance to add the burden of this structure documentation, but I can say that from an educational standpoint, it is exactly the kind of touch that makes all the difference to a beginner.  

Just to take this particular location for example, the docs currently read "Copy the following into `hello_world/app.py`:" and then show some code.  Many beginners will look at the filesystem structure they currently have, see that there is no `app.py` file present, and immediately have a question.  I would suggest that having intermittent check-ins on what your source tree should look like can be *very* helpful as a validation of the work a beginner has done, or as a corrective for mis-readings of the text step instructions along the way.

Perhaps there's a compromise to be reached, where we can have this kind of project structure check-in as a sidebar at the end of each tutorial step?

stevepiercy:
The incorrect language that causes the confusion should be fixed. For example, "Create a new file at `step_name/path/to/file.py`, then copy and paste the following code into the new file."

I'll go along with someone else (not it!) creating directory tree diagrams for steps where it makes sense (use your judgment). If it becomes a maintenance burden in the future, I may delete them. I would want the diagrams not to be within the list of steps (breaks up flow, PITA to format correctly to preserve numbering in a list), but in a new section called **Directory tree** between **Steps** and **Analysis** at the same heading level, and no further explanatory narrative text is needed. Finally syntax should use 4 spaces for indentation because it makes code editors happy. Would all that be OK?

mmerickel:
@LuisCastilloH can you do the necessary improvements requested by @stevepiercy like adding the standard heading and update the syntax?

docs/quick_tutorial/hello_world.rst
```diff
@@ -56,6 +56,17 @@ Steps
    .. literalinclude:: hello_world/app.py
     :linenos:
 
+The following tree diagram shows the current structure.
```
stevepiercy: Change this line to:

```rst
Directory tree
==============
```

docs/quick_tutorial/hello_world.rst
```diff
@@ -56,6 +56,17 @@ Steps
...
+
+   .. code-block:: text
+
+    `──/path/to/project/
+        `── projects
+            `── quick_tutorial
+                │── env
+                `── hello_world
+                       `── app.py
+
```
stevepiercy: Move the entire section and its new heading after **Steps** and just before **Analysis**.

Also `.. code-block:: text` should be flush left with no indentation, and the code sample should be indented with 4 spaces.

docs/quick_tutorial/hello_world.rst
```diff
@@ -56,6 +56,17 @@ Steps
    .. literalinclude:: hello_world/app.py
```
stevepiercy: Change the above line:

```rst
#. Copy the following into ``hello_world/app.py``:
```

to:

```rst
#. Create a new file at ``~/projects/quick_tutorial/hello_world/app.py``, then copy and paste the following code into the new file.
```

mmerickel:
I'm happy to reopen this or accept a different PR if someone puts in the work to polish it up. Closing this for now though as it seems inactive.

Issue #3212, opened 2017-12-18
HTTPClientExceptions code is 520


vlcinsky:
The docstring for [HTTPClientExceptions](https://github.com/Pylons/pyramid/blob/40d71e805bfcf8522c6af71995c05c496f1c4b4f/pyramid/httpexceptions.py#L607) states, 
> Unless specialized, this is a '400 Bad Request'

The truth is, this class is missing `code`, `title` and `explanation` so it inherits code 520 from parent error.

The same probably applies to `HTTPServerError`.

tseaver:
@bertjwregeer You updated the base class's response code in 88fcdc4e.  Can you confirm that the docs should be adjusted accordingly?  Or should we override in the two subclasses?

bertjwregeer:
It should be overridden in the two subclasses. The base class's response was modified because WebOb no longer allowed "None None" as a valid status code.

Basically what is in: https://github.com/Pylons/pyramid/blob/40d71e805bfcf8522c6af71995c05c496f1c4b4f/pyramid/httpexceptions.py#L611 should be hoisted up to the super class one further up, and things will be back to where they were.

Even before my change this would have been an issue, except that instead of sending back a `400 Bad Request` it would have sent back `None None`...

mmerickel:
Maybe `HTTPBadRequest` should be an alias for `HTTPClientError`? It does seem bad that it's a 520 right now.

mmerickel:
The only reason I'm against an alias is because I don't think it's smart to build an api where this works, even accidentally:

```python
foo = HTTPNotFound()
isinstance(foo, HTTPBadRequest) # -> True
```

With that in mind I'd just hoist the properties from HTTPBadRequest into HTTPClientError and let bad request inherit the defaults from it. So basically keep them separate classes but both looking like 4xx responses.

stevepiercy:
Can this issue be closed, now that #3280, #3281, and #3282 have been merged?

mmerickel:
Yep!

PR #3280, opened 2018-05-14
Fix inadvertent 520 error codes

cewing:
This PR fixes Issue #3212.

The error codes 400 and 500 are pulled up from the `HTTPRequestError` and `HTTPInternalServerError` to their respective base classes.  

In addition, the algorithm for calculating the `status_map` dict used to get the right exception type for the `error_response` function is updated to account for the fact that now two classes share the same error code, ensuring that the specific exception classes are returned, rather than the base exception classes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,11 @@ Features
 Bug Fixes
 ---------
 
+- Set appropriate ``code`` and ``title`` attributes on the ``HTTPClientError`` and
+  ``HTTPServerError`` exception classes. This prevents inadvertently returning a 520
+  error code.
+  See https://github.com/Pylons/pyramid/pull/3280
+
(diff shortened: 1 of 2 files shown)


stevepiercy:
Does this PR and its backports in #3281 and #3282 need a change log entry?

cewing:
I debated that.  It doesn't feel enormously significant, but I'm happy to add it if you think it appropriate.

mmerickel:
we'll want changelog entries for the backports for sure

for master it's less important because if you're upgrading from the bugfix release then there's no actual change to note, but we'll probably want one for completeness anyway

cewing:
got it.  Will add

mmerickel:
Thank you @cewing !

PR #3299, opened 2018-05-25
remove the permission arg from add_route

mmerickel:


bertjwregeer:
Should this be back ported?

mmerickel:
No, it should not be backported - even if it's a no-op in old versions it's still a bw-incompat change to remove it and could break someone's code that's passing it in which would be unacceptable behavior for a bugfix release.

PR #3300, opened 2018-06-06
Feature: Default session cookies to SameSite=Lax

bertjwregeer:
This is a follow-up to: https://github.com/Pylons/pyramid/pull/3298 that fixes the `b'Lax'` issues and bumps the version of WebOb that is required.

### DIFF
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -189,7 +189,8 @@ def BaseCookieSessionFactory(
       session cookie. Default: ``False``.
 
     ``samesite``
-      The 'samesite' option of the session cookie. Default ``'Lax'``.
+      The 'samesite' option of the session cookie. Set the value to ``None``
+      to turn off the samesite option.  Default: ``'Lax'``.
 
     ``timeout``
       A number of seconds of inactivity before a session times out. If
@@ -220,6 +221,10 @@ def BaseCookieSessionFactory(
       while rendering a view. Default: ``True``.
 
     .. versionadded: 1.5a3
+
+    .. versionchanged: 1.10
+
+       Added the ``samesite`` option and made the default ``'Lax'``.
     """
 
     @implementer(ISession)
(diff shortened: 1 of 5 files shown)


mmerickel:
How much of an issue is it to start setting samesite options on cookies when people upgrade? I imagine this will break a lot of people's test suites if they're checking the cookies pyramid sets - is that okay? I assume at least that old browsers just ignore the new option on cookies so it should have no runtime effects that would make it important for this feature to be opt-in versus old behavior but I want to ask explicitly.

pyramid/session.py
```diff
@@ -187,6 +188,9 @@ def BaseCookieSessionFactory(
       Hide the cookie from Javascript by setting the 'HttpOnly' flag of the
       session cookie. Default: ``False``.
 
+    ``samesite``
+      The 'samesite' option of the session cookie. Default ``'Lax'``.
```
mmerickel: Could we enumerate here the other valid options for samesite or is that not worthwhile? Is `None` valid to make a cookie without the option set? It would be worth mentioning one way or the other in the docs. Also as @stevepiercy mentioned on the old issue a `.. versionchanged: 1.10` should be in the docstrings of the public apis.

bertjwregeer:
Browsers ignore any unknown options. That's why Safari and others continue to work without issues even though out is not officially supported.

This will likely break test suites if users are matching cookies exactly. This already happened in the past when the double quoting around the cookie value was removed from WebOb if the cookie didn't need to be quoted. If validating cookies, one should be parsing the cookies and validating individual parts.

pyramid/session.py
bertjwregeer: Setting the `SameSite` to `None` will not set it in the cookie.

bertjwregeer: Yes, we could enumerate the other valid options here if we want.

PR #3302, opened 2018-06-11
Add http exception for status code 308

jaswilli:
This adds support for [HTTP status code 308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) ([RFC](https://tools.ietf.org/html/rfc7538)).

If I missed any place this needs to be documented please let me know.

### DIFF
--- a/pyramid/httpexceptions.py
+++ b/pyramid/httpexceptions.py
@@ -7,7 +7,7 @@ single HTTP status code.  Each class is a subclass of the
 :class:`~HTTPException`.  Each exception class is also a :term:`response`
 object.
 
-Each exception class has a status code according to :rfc:`2068`:
+Each exception class has a status code according to :rfc:`2068` or :rfc:`7538`:
 codes with 100-300 are not really errors; 400s are client errors,
 and 500s are server errors.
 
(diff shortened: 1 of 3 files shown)


pyramid/httpexceptions.py
```diff
@@ -121,10 +122,11 @@
 
 The subclasses of :class:`~_HTTPMove`
 (:class:`~HTTPMultipleChoices`, :class:`~HTTPMovedPermanently`,
-:class:`~HTTPFound`, :class:`~HTTPSeeOther`, :class:`~HTTPUseProxy` and
-:class:`~HTTPTemporaryRedirect`) are redirections that require a ``Location``
-field. Reflecting this, these subclasses have one additional keyword argument:
-``location``, which indicates the location to which to redirect.
+:class:`~HTTPFound`, :class:`~HTTPSeeOther`, :class:`~HTTPUseProxy`
+:class:`~HTTPTemporaryRedirect` and :class: `~HTTPPermanentRedirect) are
```
stevepiercy: A comma was dropped, and while we're at it, let's use the Oxford comma.

```rst
:class:`~HTTPFound`, :class:`~HTTPSeeOther`, :class:`~HTTPUseProxy`,
:class:`~HTTPTemporaryRedirect`, and :class: `~HTTPPermanentRedirect) are
```

jaswilli: Nice catch, thanks.

Updated to fix the dropped comma and also added the oxford comma.

PR #3311, opened 2018-07-04
Allow append_slash arg to add_notfound_view to cause a subrequest rather than a redirect

### LINKED ISSUE #2607: append_slash_notfound_view might should use request.invoke_subrequest instead of request.redirect
Right now the append slash notfound view redirects to the right URL, which means the redirected request is always a GET request.  If we used invoke_subrequest instead, the verb could be maintained.

mcdonc:
Closes #2607


mcdonc:
We should enable "squash and merge" here I guess, so it's easy.

mcdonc:
Enabled "squash and merge" in repo settings.

docs/narr/urldispatch.rst
```diff
@@ -908,12 +908,28 @@ exactly the same job:
...
+information contained in the original request.  Instead of returning a redirect
+response when a slash-appended route is detected during the not-found
+processing, Pyramid will call the view associated with the slash-appended route
+"under the hood" and will return whatever response is returned by that view.
+This has the potential downside that both URLs (the slash-appended and the
+non-slash-appended URLs) in an application will be "canonical" to clients; they
+will behave exactly the same, and the client will never be notified that the
+slash-appended URL is "better than" the non-slash-appended URL by virtue of a
+redirect.  It, however, has the upside that a POST request with a body can be
+handled successfully with an append-slash during notfound processing.
```
stevepiercy: Is it "notfound" or "not-found" processing? Both spellings are used in this paragraph. Please choose one.

pyramid/view.py
```diff
@@ -396,11 +412,32 @@ def aview(request):
...
+    redirect response when a slash-appended route is detected during the
+    not-found processing, Pyramid will call the view associated with the
+    slash-appended route "under the hood" and will return whatever response is
+    returned by that view.  This has the potential downside that both URLs (the
+    slash-appended and the non-slash-appended URLs) in an application will be
+    "canonical" to clients; they will behave exactly the same, and the client
+    will never be notified that the slash-appended URL is "better than" the
+    non-slash-appended URL by virtue of a redirect.  It, however, has the
+    upside that a POST request with a body can be handled successfully
+    with an append-slash during notfound processing.
```
stevepiercy: Ditto, "notfound" or "not-found"?

mmerickel:
1) Is the real motivation for this to fix #2607? Because these days HTTP 307 should be supported in most places which preserves the verb. What if we just started using `HTTPTemporaryRedirect` as the default value if `append_slash=True`?

2) Can we do this in a way that avoids adding such a specific knob? For example, some generic callback like `def append_slash_handler(request: IRequest, newpath: Str) -> IResponse`? Clearly any existing `IResponse` implementation passed as `redirect_class` could be turned into one of these in a bw-compat way and the `UseSubrequest` singleton could maybe be implemented as a handler instead of a marker object.

bertjwregeer:
I'd prefer not to use the Github Squash and Merge, especially if there are multiple commits that actually do have meaning. Since now a large PR will get squashed to a single commit.

Instead I'd recommend you `git rebase origin/master` this branch, which should get rid of all of the merge commits you've made, and then make sure that your `master` branch or wherever you are branching from matches `origin/master`.

This way we don't squash useful history, if any.

In this case/this pr it may be okay, but it's going to become a foot gun.

mcdonc:
@mmerickel headslap, I did not know about a 307 redirect; this PR is moot as a result.  I'm closing.

@bertwjregeer no worries, I just made it possible to squash and merge from the PR; the default is still just-merge.

domenkozar:
Is there any harm replacing 302 with 307 as default behavior?

mcdonc:
I don't see any harm in it.  It would not break anything, and I can't imagine someone being angry that redirects of POST requests work instead of fail in a weird way.

mcdonc:
Here's the best resource about the history and practicality of 307 I could find:

http://sebastians-pamphlets.com/the-anatomy-of-http-redirects-301-302-307/#choosing-a-redirect-response-code

domenkozar:
https://github.com/Pylons/pyramid/pull/3328

PR #3314, opened 2018-07-11
Minor update to modwsgi tutorial

jackdesert:
such that cookiecutter uses "latest" instead of "bleeding edge"

As was suggested for a tutorial that used this file as a template.
See this pull request comment in the cookbook:
https://github.com/Pylons/pyramid_cookbook/pull/208#discussion-diff-201600555R21

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -326,3 +326,5 @@ Contributors
 - Jason Williams, 2018/06/11
 
 - Benjamin Gmurczyk, 2018/06/14
+
+- Jack Desert, 2018/07/11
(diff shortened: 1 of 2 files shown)


stevepiercy:
@jackdesert on the Pyramid `master` branch, we _do_ want to pull in the `--master` cookiecutter. Similarly for other branches:

* https://github.com/Pylons/pyramid/blob/1.9-branch/docs/tutorials/modwsgi/index.rst
* https://github.com/Pylons/pyramid/blob/1.8-branch/docs/tutorials/modwsgi/index.rst

But thank you for taking time to examine the matter. I understand it is tricky, but [it is deliberate](https://github.com/Pylons/pyramid/blob/master/RELEASING.txt#L44-L47).

PR #3318, opened 2018-07-29
enable the setup function in pshell to wrap the command lifecycle

mmerickel:
The next step here is to decide whether to:

1. Integrate support directly into pshell for detecting `env['request'].tm` and doing "the right thing".

2. Add documentation in the command line chapter about how to handle transactions in pshell.

3. Add a setup function to each cookiecutter which does the right thing in that codebase.

For example:

```python
from contextlib import suppress
from transaction.interfaces import NoTransaction

def setup(env):
    tm = env['request'].tm
    tm.begin()
    try:
        yield
    finally:
        with suppress(NoTransaction):
            tm.abort()
```

The above code is what most people should want: abort any pending changes when exiting the shell. If I build this into pshell by default then it'd just be a default. If anyone defined their own `setup` func then it would override this behavior.

### DIFF
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -196,51 +196,61 @@ Extending the Shell
 It is convenient when using the interactive shell often to have some variables
 significant to your application already loaded as globals when you start the
 ``pshell``. To facilitate this, ``pshell`` will look for a special ``[pshell]``
-section in your INI file and expose the subsequent key/value pairs to the
+section in your ``.ini`` file and expose the subsequent key/value pairs to the
 shell.  Each key is a variable name that will be global within the pshell
 session; each value is a :term:`dotted Python name`. If specified, the special
 key ``setup`` should be a :term:`dotted Python name` pointing to a callable
 that accepts the dictionary of globals that will be loaded into the shell. This
 allows for some custom initializing code to be executed each time the
 ``pshell`` is run. The ``setup`` callable can also be specified from the
-commandline using the ``--setup`` option which will override the key in the INI
+commandline using the ``--setup`` option which will override the key in the ``.ini``
 file.
 
 For example, you want to expose your model to the shell along with the database
 session so that you can mutate the model on an actual database. Here, we'll
-assume your model is stored in the ``myapp.models`` package.
+assume your model is stored in the ``myapp.models`` package and that you're
+using ``pyramid_tm`` to configure a transaction manager on the request as
+``request.tm``.
 
 .. code-block:: ini
    :linenos:
 
    [pshell]
    setup = myapp.lib.pshell.setup
-   m = myapp.models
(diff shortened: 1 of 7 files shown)


stevepiercy:
> 3. Add a setup function to each cookiecutter which does the right thing in that codebase.

This could be useful in `alchemy` and `zodb` cookiecutters, but I don't understand its use in `starter` since it lacks persistent storage. Can you think of a use case?

mmerickel:
> This could be useful in alchemy and zodb cookiecutters, but I don't understand its use in starter since it lacks persistent storage. Can you think of a use case?

Yes it only belongs on cookiecutters that use pyramid_tm. The code could be written in such a way that it only does anything if pshell detects the `request.tm` variable, which would integrate the support into pyramid itself.

mmerickel:
> The only part I have trouble understanding is how the closer is defined. I see us poping it from env.

The `closer` is documented as part of the bootstrap api and is used to remove threadlocals that were pushed during bootstrap. Between calling bootstrap and invoking the `closer` you can call `get_current_registry` and `get_current_request` to get the ones defined by bootstrap.

You can see it defined here: https://github.com/Pylons/pyramid/blob/57a896f103c7378fb285f6594228578a5324d672/pyramid/scripting.py#L92-L96

docs/narr/commandline.rst
```diff
@@ -208,39 +208,53 @@ file.
...
+        env['tm'] = request.tm
+        env['dbsession'] = request.dbsession
+        try:
+            yield
+
+        finally:
+            with suppress(NoTransaction):
+                request.tm.abort()
+
+When this INI file is loaded, the extra variable ``models`` will be available for use immediately. Since a ``setup`` callable was also specified, it is executed and new variables ``testapp``, ``tm`` and ``dbsession`` are exposed, and the request is configured to generate urls from the host ``http://www.example.com``. For example:
```
stevepiercy: This file has more instances of:

```
``.ini``
```
than "INI", and I see the former more often throughout the docs. Would it be too much trouble to make them consistent?

Also Oxford comma:

```
``tm``, and ``dbsession``
```

and "URLs".

stevepiercy:
For the cookiecutters, I'm comfortable with updating docs in Pyramid, but I'm wobbly with implementing the code in the cookiecutters. @mmerickel would you take that on?

docs/narr/commandline.rst
mmerickel: fixed

bertjwregeer:
I'd be okay with having this support directly in Pyramid, so long as we make sure that `pyramid_tm` is loaded and `request.tm` is valid.

I am a bigger +1 on having this in the cookie cutters though, so that it is easier for people to understand how to change the "setup" function for pshell to match their own needs.

Issue #2733, opened 2016-08-06
support Same-site cookies


philtay:
Please support [Same-site cookies](https://tools.ietf.org/html/draft-west-first-party-cookies-07) in `pyramid.response.set_cookie`. This new standard provides protection against CSRF attacks and Google Chrome has [already implemented it](https://developer.chrome.com/extensions/cookies).

mmerickel:
This is actually an issue for webob and a duplciate of https://github.com/Pylons/webob/pull/255. However I'm going to leave this open and rephrase it.

Pyramid should add support for `same_site=True` on `AuthTktAuthenticationPolicy` and `SignedCookieSessionFactory` once webob supports it in its `CookieProfile` API.

bertjwregeer:
Is this actually RFC yet? Or is it still experimental? I need to write some docs with the experimental stuff so that rapid changes can be made as the standard evolves rather than having to worry about backwards compatibility.

ztane:
Actually now that I read it, there is no `same_site=True` but there are 2 distinct behaviours, strict and lax; strict would mean that even a GET request wouldn't send the cookie if it originates from a link on another site, so if you Google something and it directs to your site, then open the link, you are *not* logged in to your site in that tab.

AvnerCohen:
@mmerickel @ztane Any ideas if this planned any time soon?

bertjwregeer:
I have merged support for it into WebOb, but it is not yet in a release.

alex:
What is the next step here, now that this is landed in webob. Is this something that I could help with?

ztane:
Well, `pyramid.response` is a WebOb response so it supports it.

alex:
Should this bug be closed then? Or is there more work to do?

mmerickel:
@alex thanks for checking! This bug relates to Pyramid's usage / support of same-site cookies in APIs that wrap or use cookies. This means adding relevant support to `BaseCookieSessionFactory`, `SignedCookieSessionFactory`, `AuthTktAuthenticationPolicy` and probably `CookieCSRFStoragePolicy`. This basically entails adding an option to each policy, docs for the option, and tests to each object to ensure they support the option.

mmerickel:
Sessions were updated via https://github.com/Pylons/pyramid/pull/3300.

- [X] BaseCookieSessionFactory
- [X] SignedCookieSessionFactory
- [ ] AuthTktAuthenticationPolicy
[...]

PR #3319, opened 2018-08-01
Support samesite option in AuthtktAuthenticationPolicy and CookieCSRFStoragePolicy

mcdonc:
See #2733


stevepiercy:
It would be good to add a mention of what are same site cookies, why a developer would want to use them, and what the setting does in the narrative documentation. Perhaps in [Security](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/security.html)?

Are we going to support Lax only, or Strict, too?
https://github.com/Pylons/pyramid/issues/2733#issuecomment-238455333

Finally perhaps linking to the change in WebOb would be useful?
https://github.com/Pylons/pyramid/issues/2733#issuecomment-328362821

pyramid/authentication.py
```diff
@@ -565,8 +565,16 @@ class AuthTktAuthenticationPolicy(CallbackAuthenticationPolicy):
         steps.  The output from debugging is useful for reporting to maillist
         or IRC channels when asking for support.
 
+    ``samesite``
+
+        Default: ``'Lax'``.  The 'samesite' option of the session cookie. Set
+        the value to ``None`` to turn off the samesite option.
+
+        This option is available as of :app:`Pyramid` 1.10.
```
mmerickel: Move the version info down to a `.. versionchanged:: 1.10`` directive at the bottom of the docstring and document it as "Added the ``samesite`` option and made the default ``'Lax'``." similar to below in the csrf storage policy.

CHANGES.rst
```diff
@@ -37,6 +37,10 @@ Features
   ``pyramid.session.UnencryptedCookieSessionFactoryConfig``.
   See https://github.com/Pylons/pyramid/pull/3300
 
+- Modify ``pyramid.authentication.AuthTktAuthenticationPolicy`` and
+  ``pyramid.csrf.CookieCSRFStoragePolicy`` to support the SameSite option on
+  cookies.  See https://github.com/Pylons/pyramid/pull/3319
+
```
mmerickel: Document that the default value was changed to ``Lax``.

mmerickel:
@stevepiercy 

> It would be good to add a mention of what are same site cookies, why a developer would want to use them, and what the setting does in the narrative documentation. Perhaps in Security?

This is fine but I don't think it's required for this PR. That has more to do with speculative best practices and hardening like https://github.com/Pylons/pyramid/issues/1845 is requesting.

> Are we going to support Lax only, or Strict, too?

lax/strict is not something to "support", it's just something that gets set on the cookie and the browser handles.

> Finally perhaps linking to the change in WebOb would be useful?

I don't think that's useful.

bertjwregeer:
@mcdonc please run the following commands:

```
git checkout master
git fetch --all
git stash -u
git reset --hard origin/master
```

So next time you branch from `master` it doesn't contain all of the merges you've made in the mean time.

AvnerCohen:
Is this expected to be released any time soon ?

stevepiercy:
@AvnerCohen no hard date yet. Progress on the release of [Pyramid 1.10 is tracked in its milestone](https://github.com/Pylons/pyramid/milestone/7). The four "docs" tagged issues are nearly complete or will be dropped from the milestone. Other issues are in progress. Feedback is sought.

PR #3325, opened 2018-08-08
catch other pickle errors when loading content

mmerickel:
Hypothetical: you have more pyramid projects than dummy secrets and the browser starts sending back signed-pickled session cookies from project A to the server for project B. In this case it passes all signature checks but fails to unpickle with a non-`ValueError` exception.

### DIFF
--- a/pyramid/session.py
+++ b/pyramid/session.py
@@ -121,7 +121,11 @@ class PickleSerializer(object):
 
     def loads(self, bstruct):
         """Accept bytes and return a Python object."""
-        return pickle.loads(bstruct)
+        try:
+            return pickle.loads(bstruct)
+        # at least ValueError, AttributeError, ImportError but more to be safe
+        except Exception:
+            raise ValueError
 
     def dumps(self, appstruct):
         """Accept a Python object and return bytes."""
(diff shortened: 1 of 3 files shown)


pyramid/session.py
```diff
@@ -121,7 +121,11 @@ def __init__(self, protocol=pickle.HIGHEST_PROTOCOL):
 
     def loads(self, bstruct):
         """Accept bytes and return a Python object."""
-        return pickle.loads(bstruct)
+        try:
+            return pickle.loads(bstruct)
+        # at least ValueError, AttributeError, ImportError but more to be safe
+        except Exception:
+            raise ValueError
```
tseaver: It seems odd that we have no direct unit tests for `PickleSerializer`.

mmerickel: Well, you can rest easy because we have some for the `SimpleSerializer` that does nothing. :-)

mmerickel: Anyway the short answer is that the `PickleSerializer` was pulled out of existing session code and so I just assumed pyramid was exercising those paths well enough. Turns out it was missing a few things. Certainly it'd be easier to write tests for such things directly to the `PickleSerializer` if it got pulled out but I'm also not upset with more integration tests for sessions either.

mmerickel: I was going to let this slide but we are looking at changing the default serializer so I pushed some specific `PickleSerializer` tests so they don't get lost in translation later.

Issue #3317, opened 2018-07-24
DeprecationWarning from WebOb 1.8.2 (Accept parsing uses old method)


fschulze:
Since WebOb 1.8.2 the AcceptPredicate (and possibly others) cause a DeprecationWarning, see https://github.com/Pylons/webob/commit/503f4f864631a7b48a1a419068fd2173baed8075

Just wanted to leave this here so it's not forgotten.

mmerickel:
@fschulze Can you confirm that this was fixed in master via https://github.com/Pylons/pyramid/pull/3251?

bertjwregeer:
The `AcceptPredicate` is a special case, and I likely missed it when I did my initial fixes.

/assign: bertjwregeer

mmerickel:
this is fixed via #3326

Issue #2807, opened 2016-11-08
Pyramid i18n issues and documentation improvement suggestions


raspi:
Hi,

I just started using Pyramid few weeks ago and it also took that time to set up i18n with gettext. My setup: Python 3.5, chameleon templates and gettext. Ubuntu and Arch as dev environment and PyCharm as IDE. I added translation to templates first and code later. Here's some struggles I had. 

### Issues with `setup.py extract_messages`

I added 
```
message_extractors = { 'myapp': [
    ('**.py', 'python', None ),
    ('**.pt', 'chameleon', None ),
    ('static/**', 'ignore', None),
    ]}
```
To `setup.py` as was stated in manual. Running `$VENV/bin/setup.py extract_messages` couldn't find chameleon extractor. I tried xml, lingua_chameleon, lingua_xml and lingua.xml, lingua.chameleon, etc. All sorts of combinations. I couldn't find a way to list those extractors. I also tried `setup.cfg` approach.

So first suggestion: don't give extractor not found as error. Either list the possible extractors or give instructions how to list them.

Later I found that according to [this message](http://stackoverflow.com/a/30287865/71964) in SO the extraction method has been changed somewhere in or before **2015**.

I abandoned `setup.py extract_messages` and moved to `$VENV/bin/pot-create`.

### Issues with `$VENV/bin/pot-create`
I ran `$VENV/bin/pot-create --output myapp/locale/default.pot myapp`.
Error: Nothing found. Hmm. And the `.pot` file is completely empty. Next try: 
`strace -e trace=open $VENV/bin/pot-create --output myapp/locale/default.pot myapp`
Ok, it **is** scanning `.py` and `.pt` files from right locations.

First suggestion: Please list scanned files while `pot-create` is running or add `--verbose` or other parameter to give this information. I got the impression that it was doing nothing.

I found that the issue was missing `i18n:domain="myapp"` attribute in the templates. After that strings started to appear to the `default.pot`.

I created my first translation and I could change to it through `pyramid.default_locale_name` in the `.ini` file. Finally some progress.

### Issues with changing language from URL
Next thing I wanted to change language based on URL. For example `/fi/home`, `/en/home`, etc. And without adding `{language}` parameter to `add_route()`. This route pattern is **very common** in many web frameworks and there's example(s) how to implement it and it's usually documented in URL routing and/or i18n section. I think it should be added to i18n documentation. I couldn't figure out even where to start how to implement it so I went to IRC for help.

[...]

Issue #1407, opened 2014-09-05
Accept predicate with wildcard not allowed


ltvolks:
The [docs](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html) for the `accept` predicate state that wildcards are allowed, but due to the way the 'offer' is checked in WebOb, an `accept` predicate with any wildcard component will fail: 

`ValueError: The application should offer specific types, got 'text/*'`

Here's a simple example:

```python
from wsgiref.simple_server import make_server
from pyramid.config import Configurator
from pyramid.response import Response


def index(request):
    return Response("Index")

if __name__ == '__main__':
    config = Configurator()
    config.add_route('index', path='/')
    config.add_view(index, route_name='index', accept='text/*')
    app = config.make_wsgi_app()
    server = make_server('0.0.0.0', 8080, app)
    server.serve_forever()
```

```python
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 85, in run
    self.result = application(self.environ, self.start_response)
  File "/pyenv/test-accept/lib/python2.7/site-packages/pyramid/router.py", line 242, in __call__
    response = self.invoke_subrequest(request, use_tweens=True)
  File "/pyenv/test-accept/lib/python2.7/site-packages/pyramid/router.py", line 217, in invoke_subrequest
    response = handle_request(request)
  File "/pyenv/test-accept/lib/python2.7/site-packages/pyramid/tweens.py", line 46, in excview_tween
    response = view_callable(exc, request)
  File "/pyenv/test-accept/lib/python2.7/site-packages/pyramid/config/views.py", line 329, in attr_view
    return view(context, request)
  File "/pyenv/test-accept/lib/python2.7/site-packages/pyramid/config/views.py", line 300, in predicate_wrapper
    if not predicate(context, request):
  File "/pyenv/test-accept/lib/python2.7/site-packages/pyramid/config/predicates.py", line 136, in __call__
    return self.val in request.accept
  File "/pyenv/test-accept/lib/python2.7/site-packages/webob/acceptparse.py", line 120, in __contains__
    if self._match(mask, offer):
  File "/pyenv/test-accept/lib/python2.7/site-packages/webob/acceptparse.py", line 303, in _match
    _check_offer(offer)
  File "/pyenv/test-accept/lib/python2.7/site-packages/webob/acceptparse.py", line 320, in _check_offer
[...]

PR #3326, opened 2018-08-08
Overhaul HTTP Accept (mime type) handling in Pyramid

mmerickel:
@bertjwregeer does this look right? If so I can probably fix the tests.

fixes https://github.com/Pylons/pyramid/issues/3317


pyramid/predicates.py
```diff
@@ -137,7 +137,10 @@ def text(self):
     phash = text
 
     def __call__(self, context, request):
-        return self.val in request.accept
+        return any(
+            self.val == offer
+            for offer, q in request.accept.acceptable_offers([self.val])
```
tseaver: Since `q` is unused, would `_` be better?

Actually, for predicate purposes, should this just be `bool(request.accept.acceptable_offers([self.val])`?  Because `acceptable_values` does some normalization, etc.

mmerickel: I think `bool(request.accept.acceptable_offers([self.val])` is a great idea. This PR was a late night attempt to get something out there to spur things along so I'm happy for the feedback before fixing the tests. :-)

whiteroses: Currently according to the [documentation](https://github.com/Pylons/pyramid/blob/master/pyramid/config/views.py#L541) the value of the `accept` predicate can be media ranges (e.g. `text/*`, `*/*`) as well as media types, but `acceptable_offers()` only takes media types for its argument. I tried to lay out some of the issues [here](https://github.com/Pylons/pyramid/issues/1264#issuecomment-326554548) (point 1) and [here](https://github.com/Pylons/pyramid/issues/1264#issuecomment-329023914) (second half of comment) &mdash; the support for media ranges in the `accept` predicate was never working fully, as the old `NilAccept`'s (and so the old `MIMENilAccept`'s) [`__contains__`](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L198) called [`_check_offer`](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L340), which raised `ValueError("The application should offer specific types, got %r" % offer)`; and using media ranges for the `accept` predicate is ambiguous, and in my opinion does not make sense. If the predicate were to only take media types though, then `bool(request.accept.acceptable_offers([self.val])` would work well (...though only for a single media type. It could however be easily extended to take an iterable of media type strings &mdash; a finite number of media type offers makes sense I think.)

mmerickel: Can you show a failing test (ideally a PR to pyramid's test suite but something standalone is fine too) that demonstrates what you're describing? Presumably a request with no accept header, and a view defined with a wildcard range?

whiteroses: According to RFC 7231, e.g. `*/*` is (confusingly) *both* a media type and a media range. `acceptable_offers()` treats it as a media type (so the `*` has no special meaning), and a request with no accept header and a view defined with a wildcard range would match:

    >>> create_accept_header(header_value=None).acceptable_offers(['*/*'])
    [('*/*', 1.0)]

But

    >>> create_accept_header(header_value='text/*').acceptable_offers(['*/*'])
    []
    >>> create_accept_header(header_value='text/html').acceptable_offers(['*/*'])
    []

And 

    >>> create_accept_header(header_value=None).acceptable_offers(['text/*'])
    [('text/*', 1.0)]
    >>> create_accept_header(header_value='text/*').acceptable_offers(['text/*'])
    [('text/*', 1.0)]
    >>> create_accept_header(header_value='text/html').acceptable_offers(['text/*'])
    []

Working on branches and tests to help clarify what the behaviour was before, what has changed and why &mdash; should be done sometime this week.

mmerickel:
@whiteroses I pushed an updated version that uses just media types as an example. I'm afraid this is a big bw-incompatibility though so that needs to be considered carefully.

whiteroses:
Thanks @mmerickel. I made three branches to help us see what changed in WebOb and what needs to change in Pyramid:

https://github.com/whiteroses/pyramid/tree/accept-arg-current-behavior:
This branch has some tests that show what currently happens when we use media ranges in the `accept` argument to `add_route` and `add_view`.

https://github.com/whiteroses/pyramid/tree/switch-to-acceptable_offers:
Following on from accept-arg-current-behavior, I added your changes involving the switch to `acceptable_offers`, but *without* changing to accept only media types, to show the tests that would fail.

https://github.com/whiteroses/pyramid/tree/before-webob-changes
This is branching off Pyramid's 1.9-branch, so that we can pin WebOb to 1.7.4, before the GSOC changes. Then I added the tests for media ranges from the accept-arg-current-behavior branch, to show that the use of media ranges was already broken for `MIMENilAccept`. The call to `_check_offer()` &mdash; the reason you got the `ValueError` [here](https://github.com/Pylons/pyramid/pull/1259#issuecomment-36934346) -- was removed in the GSOC changes, which fixed the `ValueError` and the inconsistency with the behaviour of `MIMEAccept` and the documentation; but the use of media ranges for the `accept` argument remained ambiguous, as explained [here](https://github.com/Pylons/pyramid/issues/1264#issuecomment-329023914) and can be seen in some of the tests I added (`test_header_ruled_out_by_specific_media_type_q0`, `test_header_ruled_out_by_type_range_q0`, and `test_header_ruled_out_by_all_range_q0`). (As the `ValueError` would have been raised whenever the `Accept` header was not present in the request or was empty and the `accept` argument had an asterisk in it, and no one had raised this issue as far as I know, does that mean that no one was really using media ranges with the `accept` argument?) 

There is more information in some of the commit messages. Please let me know if there is anything I can clarify!

docs/narr/viewconfig.rst
```diff
@@ -285,6 +285,23 @@ Non-Predicate Arguments
   are just developing stock Pyramid applications. Pay no attention to the man
   behind the curtain.
 
+``accept``
+  This value represents a match query for one or more mimetypes in the
```
whiteroses: I'm not sure if "one or more mimetypes" was referring to the possibility of a media range in the argument representing multiple mimetypes (in which case that would no longer apply); or if it was referring to "one or more mimetypes in the header", in which case the match would not be only against mimetypes in the header, but also media ranges? Perhaps we could say something like "This is a media type, such as ``text/html``. If the ``Accept`` header is not present in the request or is invalid, or if the media type is considered acceptable by the ``Accept`` header, then this will match the current view..."?

docs/narr/viewconfig.rst
```diff
@@ -285,6 +285,23 @@ Non-Predicate Arguments
...
+  ``Accept`` HTTP request header.  If this value is specified, it must
+  be a specific media type, such as ``text/html``.  If the media type
+  matches the ``Accept`` header of the request, or if the ``Accept``
+  header isn't set at all in the request, this will match the current
+  view. If this does not match the ``Accept`` header of the request,
+  view matching continues.
```
whiteroses: For view configuration, wouldn't view matching (unlike route matching) continue whether or not there is a match?

I see you've moved `accept` from under "Predicate Arguments" to under "Non-Predicate Arguments" &mdash; currently it seems the `accept` argument is handled as if it were both a non-predicate *and* a predicate argument? Both `best_match` and `__contains__` are used, with the call to `__contains__` seemingly unnecessary; and with the switch to `acceptable_offers`, `acceptable_offers` would be called twice. Am I reading it correctly, in that it seems `accept` shouldn't be a predicate at all for view configuration?

pyramid/predicates.py
```diff
@@ -132,14 +132,19 @@ def __call__(self, context, request):
 class AcceptPredicate(object):
     def __init__(self, val, config):
         self.val = val
+        if '*' in self.val:
+            raise ConfigurationError(
+                '"accept" predicate does not accept media ranges, only '
+                'specific media types',
+            )
```
whiteroses: Might it be better for the `ConfigurationError` to be raised within `add_view()` and `add_route`? It'd happen earlier; plus as mentioned above, the call to the predicate in view matching seems (if I haven't missed something) unnecessary and duplicates the work of the call to `best_match`/the first call to `acceptable_offers` that treats it as a non-predicate argument?

docs/narr/viewconfig.rst
mmerickel: Accept handling in pyramid is a predicate "with special extra support if you register more than one view for the same other-predicates but with different accept values".

It is used directly as a predicate in the following cases:

1) When used as a route predicate.
2) When only one view is registered and thus a MultiView is not created.
3) After the MultiView selects the view, it is checked again (an redundant check which we could optimize away).

pyramid/predicates.py
mmerickel: The reason that I have done it here is that otherwise the check would need to be done in both `add_route` and `add_view` at the very least and this will raise at config-time anyway which is the primary goal here.

docs/narr/viewconfig.rst
mmerickel: As far as why I moved this to the non-predicate arguments section, it was more than anything just to make it consistent with the current `add_view` docs. I'm probably going to move them both back to the predicate section before this is all said and done.

mmerickel:
Here is my current plan:

- [ ] deprecate any usage of `*` in the `AcceptPredicate` but if it's there then fall back to using `__contains__` as before.
- [ ] keep the old behavior in `MultiView` where it only supports negotiation if the accept value does not include `*` (otherwise it punts to `__contains__` later in the `AcceptPredicate`).
- [ ] add `config.add_accept_order(media_type, weighs_more_than=..., weighs_less_than=...)` which will build up an `IAcceptOrder` object on the registry with a sorted list of media types.
- [ ] modify `MultiView` to sort the view offers based on `IAcceptOrder` before passing the offers into `request.accept.acceptable_offers()` so that the selected views are sorted based on user preferences (instead of the currently random behavior which causes issues when the client specifies missing/invalid `Accept` header, or the client is ambiguous such as `Accept: */*`).

future enhancements:

- come up with a way for the `MultiView` to short-circuit the `AcceptPredicate` if the `MultiView` already did all the work, such that `AcceptPredicate` can be a no-op instead of checking it again (this double checking currently happens in Pyramid in this scenario).
- a future PR based on this work can add support for a list of media types being passed to `AcceptPredicate` (such as `accept=['text/html', 'application/xml+xhtml']`) as a simple way to get around using media ranges.
- add some type of support for predicates to return a 2-tuple of `(True, info)` which would be set on the request only if the view was selected (something like `request.view_predicates[predicate_name] = info`).
- use the `request.view_predicates` support to tell the view which `accept` media type was selected.

mmerickel:
@whiteroses As I get further into this I think media ranges are an acceptable thing to use for `add_route` and am concerned that there's no way to do that anymore in webob. The use case in routes is specifically the `__contains__` feature where we simply want to know if the request allows any response satisfying a media type or media range. There is no content negotiation here. How do I do this in webob now without using the deprecated `__contains__` method?

whiteroses:
What does it mean to "check if a range is considered acceptable by the header", given that the idea does not exist in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.3.2)? The `Accept` header is used "to specify response *media types* that are acceptable", not media ranges; media ranges are only used *in the header*, to specify the range of media types that are acceptable.

Say the `accept` argument to an `add_route` is 'text/*'. The `Accept` header coming in is `text/html;q=0, */*`: every media type is acceptable except `text/html`. Should we consider that a match?

This is the same scenario I mentioned in the second part of [the comment here](https://github.com/Pylons/pyramid/issues/1264#issuecomment-329023914), and applies just as much to `add_route` as to `add_view`.

The old [`MIMEAccept._match`](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L278) invented its own algorithm that is not in any of the RFCs &mdash; it didn't just accept media ranges, but also [just '*'](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L292). It did not handle '`q=0`'s properly: `__contains__` [used `._parsed_nonzero`](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L111) &mdash; so in the scenario I mentioned above, it would respond as if the header were simply '`*/*`'.

As mentioned before, if the `Accept` header was [not present or was the empty string](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L351), a `ValueError` [would be raised](https://github.com/Pylons/webob/blob/1.7-branch/webob/acceptparse.py#L199) &mdash; and yet no one had complained. So were there many people actually using '`*`' in the `accept` argument to `add_route` or `add_view`, and how did they not run into this `ValueError`?

I don't understand why Pyramid would want to borrow the media range concept, extend it by adding '`*`' on its own, create its own algorithm so that instead of media ranges being only found in the header as in the RFC, make it so that it could be in *both* the header and the `accept` argument to `add_route` &mdash; such that Pyramid would then have to define the meaning of how media ranges on both ends would interact (with the added complication of '`q=0`'s), and create its own extension to the already complicated and underspecified RFC?

bertjwregeer:
One of the issues is that `add_route` can be used for many views, those views may limit the `accept` to one or more types, but they should be well defined.

However, for `add_route(accept='text/*')` which has two views: `add_view('viewA', accept='text/html')` and `add_view('viewB', accept='text/other')` in the case of an `Accept: text/html;q=0, */*` we would still want the route to match, and then let the views decide which one is actually acceptable, in this case only `viewB`.

I don't think it'll be hard to add support for `something/*` in `acceptable_offers` and we can put it behind a flag that has to be true if necessary.

----

Reading about what you are saying about `__contains__` calling `_check_offer`, I'm going to see if I can write a quick test in the pyramid test suite that we can use to reason through what is expected.

The question I have, does anyone rely on the ability to pass `whatever/*` to `accept` as a predicate?

pyramid/config/views.py
```diff
@@ -122,7 +122,7 @@ def add(self, view, order, accept=None, phash=None):
                     self.views[i] = (order, view, phash)
                     return
 
-        if accept is None or '*' in accept:
```
bertjwregeer: We don't even try to do the right thing here, if accept were `text/*` then they wouldn't even get counted...

pyramid/predicates.py
```diff
@@ -132,14 +132,19 @@ def __call__(self, context, request):
...
 
     def text(self):
         return 'accept = %s' % (self.val,)
 
     phash = text
 
     def __call__(self, context, request):
-        return self.val in request.accept
```
bertjwregeer: As @whiteroses mentioned, I don't think this allowed you to pass in `text/*` for example...

whiteroses:
@bertjwregeer So if I understand correctly, the '`*`' in the `accept` argument to `add_route` would mean a match if the range in the argument has not been *entirely* ruled out by the header, e.g. if the header is `text/html, text/other;q=0` and the `accept` argument to `add_route` is `text/*`, then that would be a match?

Wouldn't not specifying the `accept` argument to `add_route` at all get us the same result?

And if there is a case where this is needed, would it not be something very much Pyramid-specific, and belongs in Pyramid rather than WebOb?

pyramid/predicates.py
whiteroses: You can pass in `text/*` I think, but you'd get a `ValueError` from the predicate calling `__contains__` if the header was not present in the request or was the empty string &mdash; you can see it in [this test](https://github.com/whiteroses/pyramid/blob/before-webob-changes/pyramid/tests/test_integration.py#L895) from the [before-webob-changes](https://github.com/whiteroses/pyramid/tree/before-webob-changes) branch from [here](https://github.com/Pylons/pyramid/pull/3326#issuecomment-414161922).

mmerickel:
Ok thanks for your patience @whiteroses, my motivation has been to try and do what makes "sense" while also trying to figure out and maintain compat with older versions of Pyramid.

- The Pyramid docs clearly accept media ranges, but I tried going all the way back to webob 1.0.8 and I still get the same error `ValueError: The application should offer specific types, got 'text/*'` in all versions of webob PRIOR TO 1.6.
- In versions 1.6 and 1.7 media ranges are accepted so long as the client passes a valid accept header, otherwise they will raise.
- In webob 1.8 they also are accepted and treated as media types instead of media ranges.

That being said, I have found some other odd things going on with webob's accept parsing in both old versions and the new version. For example, behaviors are quite different on "invalid" media types when passing them as offers to invalid/missing/valid accept headers. Some raise a `ValueError` and some are pass through and this inconsistency is not good if we want people to use this feature in robust apps. If the app offers an invalid media type then it should raise consistently in each scenario. I opened the below issues to get these fixed in webob.

https://github.com/Pylons/webob/issues/365
https://github.com/Pylons/webob/issues/366

From here, I think I now have to balance the fact that the Pyramid docs were always wrong and magically started being correct (sometimes - based on the client's value of the accept header) in webob 1.6/1.7 and 1.8 preserved that behavior by treating them as media types instead of ranges (this should raise imo).

My proposal is to fix these issues I've opened in webob and then continue with my above solution with the following tweak:

- I think we should just start raising in Pyramid at config-time to make things consistent and predictable. Media ranges were not supported prior to webob 1.6 and I think basically everyone seems to agree that they should not be supported. So let's just do that.

With this tweak, we will be resilient to whatever webob decides to do in 1.8+ (right now it treats ranges as media types and doesn't raise, and I propose it gets fixed to raise) - but regardless Pyramid will raise.

mmerickel:
If people could look at the Webob issues I've opened and weigh in it would help me greatly in determining how to integrate this into Pyramid. The invalid headers not causing a fallback to the view with no accept predicate I think is a pressing issue right now to the extent that if Webob doesn't want to fix it then I'll probably have to check manually if it's an invalid header and handle it on my own versus relying on `acceptable_offers`.

mmerickel:
After talking to @bertjwregeer I think we're happy with the invalid headers being treated as "accept anything" just like a missing header (which is defined by the RFC to accept anything). This basically means that any view with `accept=None` will have lowest priority and will only execute if all other `accept=foo` views fail on other predicates.

With that being said, I think the implementation here is working and correct and I just need to fix up the tests.

mmerickel:
Well, I just saw https://github.com/Pylons/pyramid/issues/1407 which makes me question again whether this is a good idea to limit ourselves to media types only as the `best_match` behavior in webob had already been changed once to support media ranges, and now acceptable offers has gone backward. We cannot keep flip flopping on this behavior.

mmerickel:
ping @ltvolks as well, as the brains behind #1407.

mmerickel:
Hey @mcdonc this last commit (762bd9c) adds support for a list of accept predicates but to do so it messes with view config in a fundamental way... a single predicate value registers multiple views in the app. Thoughts?

docs/narr/viewconfig.rst
```diff
@@ -1028,6 +1029,76 @@ these values.
...
+        config.add_accept_view_order(
+            'application/json',
+            weighs_more_than='text/html',
+        )
+        config.scan()
+        return config.make_wsgi_app()
+
+In this case, the ``application/json`` view should always be selected in cases where it is otherwise ambiguous.
+
+Default Accept Ordering
```
stevepiercy: Insert ``.. _default-accept-ordering:`` for indexing.

docs/narr/viewconfig.rst
```diff
@@ -1028,6 +1029,76 @@ these values.
...
+
+The appropriate view is selected here when the client specifies an unambiguous header such as ``Accept: text/*`` or ``Accept: application/json``.
+Similarly, if the client specifies a media type that no view is registered to handle, such as ``Accept: text/plain``, it will fall through to ``myview_unacceptable`` and raise ``406 Not Acceptable``.
+There are a few cases in which the client may specify ambiguous constraints:
+
+- ``Accept: */*``.
+- A missing ``Accept`` header.
+- An invalid ``Accept`` header.
+
+In these cases the preferred view is not clearly defined (see :rfc:`7231#section-5.3.2`) and :app:`Pyramid` will select one randomly.
```
stevepiercy: Is it truly "randomly" or "that may be unexpected"?

docs/narr/viewconfig.rst
```diff
@@ -1028,6 +1029,76 @@ these values.
...
+- ``text/html``
+- ``application/xhtml+xml``
+- ``application/xml``
+- ``text/xml``
+- ``text/plain``
+- ``application/json``
+
+API clients tend to be able to specify their desired headers with more control than web browsers, and can specify the correct ``Accept`` value, if necessary.
+Therefore, the motivation for this ordering is to optimize for readability.
+Media types that are not listed above are ordered randomly during :term:`view lookup` between otherwise-similar views.
```
stevepiercy: Another instance of "randomly", where I'm not sure that's the most accurate term.

pyramid/config/views.py
```diff
@@ -1222,6 +1259,71 @@ def add_default_view_predicates(self):
...
+        weighs_more_than=None,
+        weighs_less_than=None,
+    ):
+        """
+        Specify an ordering preference for the ``accept`` view option used
+        during :term:`view lookup`.
+
+        By default, if two views have different ``accept`` options and a
+        request specifies ``Accept: */*`` or omits the header entirely then
+        it is random which view will be selected. This method provides a way
```
stevepiercy: Another instance of "random", where I'm not sure that's the most accurate term.

docs/narr/viewconfig.rst
mmerickel: It's random. It is iterating through a python `set` object.

docs/narr/viewconfig.rst
stevepiercy: I merged two thoughts together in my original comment. One is the reference which gets converted to an id and may be linked to, the other is adding an index entry for new sections. Sorry about that.

Sphinx will generate ids for all levels of section headings, whether or not we add a reference. It's not required, but we can control it. But more importantly, without a reference, you can't do this in reST:

```
:ref:`default-accept-ordering`
```

...and instead hyperlinks would have to be grabbed from the rendered HTML and inserted into the reST. That's always bothered me, especially when using intersphinx, so I tend to add references for every section. You never know who will link back to a given section.

As far as index entries, that's up to you if you think they're useful here. I think we've lately been slacking off on them, and I haven't been paying much attention to them. I guess a good rule of thumb is to ask oneself, "Would index entries help make this section easy to find?"

mmerickel: I've never once used the index directive and I've never deep linked to subsections. At the very least I don't really want to talk about any of this until I can get some actual code review on the substance of this PR because I have already re-written the docs section multiple times. If you want me to add some of these directives for the index then please paste examples because I just don't have much motivation for them. Sorry :-(

stevepiercy: Please let me know when you want me to review docs again.

mmerickel: @stevepiercy please review the docs again when you get a chance

docs/narr/viewconfig.rst
```diff
@@ -1027,7 +1027,8 @@ See :ref:`environment_chapter` for more information about how, and where to set
 these values.
 
 .. index::
-   single: HTTP caching
+   single: Accept
+   single: Accept content negotation
```
stevepiercy: negotiation

docs/narr/viewconfig.rst
```diff
@@ -1082,9 +1084,30 @@ For example:
 
 In this case, the ``application/json`` view should always be selected in cases where it is otherwise ambiguous.
 
+.. _default_accept_ordering:
```
stevepiercy: Above this line, suggest an optional index entry:

```
.. index::
    single: default accept ordering
```

stevepiercy: I'm not sure whether we need to use a `.. versionchanged:: 1.10` or `.. versionadded:: 1.10` somewhere in these changes, too.

pyramid/util.py
```diff
@@ -649,3 +652,79 @@ def loads(self, bstruct):
...
+
+    Supported offers are of the following forms, ordered by specificity
+    (higher to lower):
+
+    - ``type/subtype;params`` and ``type/subtype``
+    - ``type/*``
+    - ``*/*``
+
+    :param offers: A list of offers to be sorted.
+    :param order: A weighted list of offer where items closer to the start of
```
stevepiercy: should it be plural? offers

mmerickel:
@whiteroses I've copied your test cases into this branch verbatim and they all pass now.

For everyone listening, this PR is officially ready for review, contingent on webob merging https://github.com/Pylons/webob/pull/376 and getting a new release for all the range support.

pyramid/predicates.py
mmerickel: this works properly now, contingent on webob master + https://github.com/Pylons/webob/pull/376

tox.ini
```diff
@@ -25,6 +25,9 @@ commands =
 extras =
     testing
 
+deps =
+    -egit+https://github.com/mmerickel/webob.git@accept-parse-offer#egg=webob
```
mmerickel: remove this before merging

pyramid/tests/test_integration.py
```diff
@@ -712,7 +712,234 @@ def test_wildcards(self):
...
+            '/', headers={'Accept': 'text/*;q=0, text/html'}, status=200,
+        )
+        self.assertEqual(res.content_type, 'text/plain')
+
+    def test_header_ruled_out_by_all_range_q0(self):
+        res = self.testapp.get(
+            '/', headers={'Accept': '*/*;q=0, text/html'}, status=200,
+        )
+        self.assertEqual(res.content_type, 'text/plain')
+
```
whiteroses: @mmerickel These tests were written to show the behaviour of Pyramid at the time ([original comment](https://github.com/Pylons/pyramid/pull/3326#issuecomment-414161922)), not what I thought Pyramid should do &mdash; in particular, `test_header_ruled_out_by_specific_media_type_q0`, `test_header_ruled_out_by_type_range_q0`, and `test_header_ruled_out_by_all_range_q0` were intended to show that allowing use of media ranges in the `accept` argument meant that the view would return a content type (`text/plain`) that the client has explicitly said it doesn't accept (`text/plain;q=0, */*`, `text/*;q=0, text/html` and `*/*;q=0, text/html`). You can see this in the same tests for all four test classes &mdash; I was trying to show the inherent  problem in allowing ranges in the `accept` argument.

mmerickel: @whiteroses I agree with you that I think those cases should fail and I incorrectly assumed that they were written to show what you thought should be correct, and then we would see which ones passed or failed on various combinations of pyramid/webob. I'd encourage you to submit either fixes to webob or at least updated test cases that demonstrate your intentions.

No comments were added to https://github.com/Pylons/webob/pull/370 so I'm not sure if you're concerned about Pyramid's usage here, or the media range support I added to Webob. I'm assuming that it's Webob itself and if these things are fixed in Webob then the Pyramid tests will start properly failing as you expect.

You should review the state of Webob with respect to https://github.com/Pylons/webob/pull/376.

mmerickel: With respect to the q=0 issues, I cannot agree that things are really broken right now. There is no information in the content negotiation that declares that `text/plain` is the content type that's going to be returned. We are not obligated to bend over backward to satisfy the client. In these tests the view is for `*/*` which means it will accept anything - and it does. The only way this behavior could be fixed in Pyramid is by adding some checks in the renderer itself, or in something on egress. As far as ingress is concerned, things seem to work to me and I'm ok with the current assertions in `test_header_ruled_out_by_all_range_q0`, `test_header_ruled_out_by_type_range_q0`, `test_header_ruled_out_by_specific_media_type_q0`.

mmerickel: I think my main takeaway from this is that if you want to put any text in the docs that discourages people from using media ranges, I'm fine with that - but I don't think Pyramid is doing the wrong thing here if you do specify media ranges. It's a best-effort match on the range, and if you want to check a specific type then you are responsible for doing that in view code or somewhere else.

whiteroses: If you are ok with discouraging people from using media ranges as offers in the docs, then may I ask, *why support media ranges in the first place?* So much of the complexity is just to accommodate media ranges in offers.

I didn't comment on https://github.com/Pylons/webob/pull/370 because I'd written so many times already on the problems with ranges in offers that I thought you knew my view already; I was also hoping that maybe someone else might speak up, or that you might see from the lack of response to [this](https://github.com/Pylons/pyramid/pull/3326#issuecomment-417131147) and [this](https://github.com/Pylons/pyramid/pull/3326#issuecomment-417415052) that nobody seems to care about using media ranges in offers. (We can see, from the PR that originally changed WebOb to accept ranges in offers, that it "Fixes [#155](https://github.com/Pylons/webob/issues/155)". Click through to the WebOb issue, and we can see that @ltvolks himself notes that WebOb not accepting ranges in offers "is not insensible", but conflicts with Pyramid's documentation. Which is what in my view has been happening since: that everyone has been trying to make the behaviour match what was a mistake in the documentation.)

It is important to note, that we've established that media ranges in offers weren't (and aren't) actually working, because of [this](https://github.com/Pylons/pyramid/pull/3326#discussion_r213148720) in Pyramid (as well as the `_check_offer`s in `NilAccept`'s `__contains__` and `best_match` in WebOb).

The problem shown in those three tests above are present not only when the offer is `*/*`, but also when the offer is e.g. `text/*` (as seen in [these three tests](https://github.com/Pylons/pyramid/pull/3326/files#diff-31c901158fb3dca6c7e67ac2216cb057R853)). Yes, "There is no information in the content negotiation that declares that `text/plain` is the content type that's going to be returned" &mdash; but there is only no information when the offer is a range; when the offer is a media type, the offer *is* the content type. Doesn't this seem much more straightforward, and in line with what is described in the RFC? [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.3.2) is clear on offers being media types: we can see this for example in its discussion of "media type quality factors".

If there are no media ranges in offers, there would be no need for things like `sort_accept_offers`, no need to invent new algorithms around offers, and no need to worry about how these new algorithms interact with what is specified in the RFC, and whether we've thought of everything and accounted for all the cases. It would save you so much work and complexity, from not having to essentially design and maintain an extension to content negotiation as specified by the RFC; and it'd allow us to focus on switching to `acceptable_offers` (which without media ranges in offers should be a lot more straightforward), and dealing with the randomness in accept handling. If you're already at the point where you see the need to discourage people from using media ranges as offers in the docs, could I persuade you to not support media ranges in offers at all?

mmerickel: At the very least I cannot talk myself out of supporting ranges in add_route as it is the gateway to more fine-grained content negotiation between views. As far as views go I agree that it’s way simpler to drop support for ranges due to the complexity in `sort_accept_offers` but unfortunately I’m not willing to break compatibility when I see a path forward to maintain it. Which I feel I have accomplished in these PRs. Especially when it’s clear that at least some people were depending on that behavior. I take compatibility seriously. 

I’d break it if I couldn’t support it. But at the end of the day the code works within reason imo. What would you have me do? Remove a bunch of code that I toiled over and have already written to maintain compatibility? I’d still argue to keep the range support in webob for add_route purposes even if we dropped ranges in `sort_accept_offers`.

Finally. `sort_accept_offers` still needs to exist to sort between params and non-params media types in a hierarchy. It just wouldn’t need to deal with the extra level of sorting ranges. It’s not insanely simpler.

One way or another as I said above I think matching against ranges is legitimately useful in add_route.

mmerickel: I chatted with @bertjwregeer for a while this morning and we decided that, while things work right now and we could probably ship it, the `q=0` argument with ranges is compelling enough to us that ranges-as-offers probably shouldn't be supported in webob or in add_view. With this being decided, the implementation details of bringing range support to add_route (which I still think is a fine feature) is probably going to be too great and since I also added list-of-offer support to add_route already I'll probably just point people at that instead of supporting ranges there. Finally, I'll probably remove support for ranges in the `order` argument of `sort_accept_offers` (meaning removing ranges from `add_accept_view_order`). This doesn't need to be done, but I think it'll just be annoying to support without being able to rely on some of the parsing code in webob that it currently utilizes.

I'm going to work on making these changes in the next few days.

CHANGES.rst
```diff
@@ -57,6 +57,16 @@ Features
 - Add support for Python 3.7. Add testing on Python 3.8 with allowed failures.
   See https://github.com/Pylons/pyramid/pull/3333
 
+- Added the ``pyramid.config.Configurator.add_accept_view_order`` directive,
+  allowing users to specify media type preferences in ambiguous situations
+  such as when several views match. A default ordering is defined for media
+  types that prefers human-readable html/text responses over JSON.
+  See https://github.com/Pylons/pyramid/pull/3326
+
+- Support a list of media types in the ``accept`` predicate used in
```
bertjwregeer: AFAIK we aren't supporting a list anymore, right?

docs/narr/viewconfig.rst
```diff
@@ -344,6 +344,20 @@ configured view.
...
+  If this value is specified, it must be a specific media type, such as ``text/html``.
+  If the media type is acceptable by the ``Accept`` header of the request, or if the ``Accept`` header isn't set at all in the request, this predicate will match.
+  If this does not match the ``Accept`` header of the request, view matching continues.
+
+  If ``accept`` is not specified, the ``HTTP_ACCEPT`` HTTP header is not taken into consideration when deciding whether or not to invoke the associated view callable.
+
+  See :ref:`accept_content_negotiation` for more information.
+
+  .. versionchanged:: 1.10
+      Media ranges such as ``text/*`` will now raise :class:`pyramid.exceptions.ConfigurationError`.
```
bertjwregeer: :+1:

docs/narr/viewconfig.rst
```diff
@@ -1025,6 +1026,105 @@ and the browser representation of the error will include the same information.
...
+.. index::
+    single: default accept ordering
+
+.. _default_accept_ordering:
+
+Default Accept Ordering
+~~~~~~~~~~~~~~~~~~~~~~~
+
+:app:`Pyramid` will always sort multiple views with the same ``(name, context, route_name)`` first by the specificity of the ``accept`` offer.
+This means that ``text/plain`` will always be offered before ``text/*``.
```
bertjwregeer: Are we still accepting media ranges for add_route? But not add_view right? Does this need clarification?

pyramid/config/views.py
```diff
@@ -677,6 +673,22 @@ def wrapper(context, request):
           represents a header name or a header name/value pair, the
           case of the header name is not significant.
 
+        accept
```
bertjwregeer: This was moved in https://github.com/Pylons/pyramid/pull/1487... I just apparently missed the narrative docs. Its not a proper predicate so you can't use `not_` on it.

Although `not_` really is a misfeature...

----

This is going to be confusing since for routes, it is still a predicate. *sigh*

pyramid/config/views.py
```diff
@@ -804,22 +816,21 @@ def wrapper(context, request):
                 stacklevel=4,
                 )
 
+        if accept is not None:
+            if is_nonstr_iter(accept):
+                raise ConfigurationError(
+                    'A list is not supported in the "accept" view predicate.',
+                )
+            accept = accept.lower()
```
bertjwregeer: Could we instead run it through the normalizer in WebOb? This may be a stupid question...

pyramid/predicates.py
```diff
@@ -131,15 +131,17 @@ def __call__(self, context, request):
 
 class AcceptPredicate(object):
     def __init__(self, val, config):
-        self.val = val
+        if not is_nonstr_iter(val):
+            val = (val,)
```
bertjwregeer: Why make this a tuple? Instead of a list or a set? I guess it doesn't matter...

pyramid/tests/test_config/test_util.py
```diff
@@ -431,6 +431,65 @@ def test_it(self):
             from pyramid.config.predicates import XHRPredicate
             self.assertEqual(len(w), 1)
 
+class Test_sort_accept_offers(unittest.TestCase):
```
bertjwregeer: These tests would look 😍😍😍 in pytest!

tox.ini
bertjwregeer: ☝️

pyramid/config/views.py
bertjwregeer: @stevepiercy the use of the word random here is fine. Ultimately it ends up depending on the order in which your views are added, and the position of the sun and the `PYTHONHASHSEED`.

docs/narr/viewconfig.rst
mmerickel: I am still working on the changes but the intention is that ranges are supported in both right now but deprecated. As far as that example, it's getting fixed once I push the changes.

pyramid/predicates.py
mmerickel: there's no difference for this purpose and I like immutable objects

pyramid/config/views.py
mmerickel: yep, it will be once I push changes

mmerickel: Well to be clear I'm running them through `parse_offer` to check for validity but that api does not normalize. Webob doesn't expose an api right now to normalize offers so we still do `.lower()` in pyramid.

mmerickel: I looked briefly into making `parse_offer` do the normalization but it was a pain to do correctly because it requires reconstructing an offer with the params, which requires re-quoting things properly. If that were added to `webob.acceptparse.AcceptOffer.__str__` then I would use it.

CHANGES.rst
```diff
@@ -103,6 +113,17 @@ Deprecations
   implementation if you're still using these features.
   See https://github.com/Pylons/pyramid/pull/3353
 
+- Media ranges are deprecated in the ``accept`` argument of
```
bertjwregeer: :+1:

CHANGES.rst
```diff
@@ -103,6 +113,17 @@ Deprecations
...
+  ``pyramid.config.Configurator.add_route``. Use a list of explicit
+  media types to ``add_route`` to support multiple types.
+
+- Media ranges are deprecated in the ``accept`` argument of
+  ``pyramid.config.Configurator.add_view``.  There is no replacement for
+  ranges to ``add_view``, but after much discussion the workflow is
+  fundamentally ambiguous in the face of various client-supplied values for
```
bertjwregeer: s/fundamentally ambiguous/fundamentally ambiguous and impossible to support sanely/

CHANGES.rst
mmerickel: it's supported in add_route, not add_view

pyramid/config/views.py
```diff
@@ -804,22 +823,31 @@ def wrapper(context, request):
...
+            if '*' in accept:
+                warnings.warn(
+                    ('Passing a media range to the "accept" argument of '
+                     'Configurator.add_view is deprecated as of Pyramid 1.10. '
+                     'Use explicit media types to avoid ambiguities in '
+                     'content negotiation that may impact your users.'),
+                    DeprecationWarning,
+                    stacklevel=4,
+                    )
+            # XXX when media ranges are gone, switch verify=True
```
bertjwregeer: Triple X the new "TODO"?

pyramid/predicates.py
```diff
@@ -130,16 +130,26 @@ def __call__(self, context, request):
...
 
     def text(self):
-        return 'accept = %s' % (self.val,)
+        return 'accept = %s' % (', '.join(self.values),)
 
     phash = text
 
     def __call__(self, context, request):
-        return self.val in request.accept
+        if self._is_using_deprecated_ranges:
```
bertjwregeer: Knowing that this will warn a `DeprecatedWarning`, should we filter that here knowing that we already told the developer during config time?

CHANGES.rst
mmerickel: My intention was that the word fundamental was trying to convey that sentiment.

pyramid/config/views.py
stevepiercy: Ha, I just saw this comment after reviewing. Forget mine. I have a broader meaning for TODO.

Is there an issue in the tracker attached to a milestone for a future release that covers this?

mmerickel: I added https://github.com/Pylons/pyramid/issues/3383

pyramid/config/views.py
mmerickel: last night we updated `AcceptOffer` to normalize the offers so this is fixed now and pyramid no longer does the `lower()` itself.

pyramid/config/views.py
mmerickel: I updated the docs for both indicating which ones support it and which don't.

docs/narr/viewconfig.rst
```diff
@@ -1025,6 +1029,106 @@ and the browser representation of the error will include the same information.
...
+            'name': request.GET.get('name', 'bob'),
+        }
+
+    @view_config()
+    def myview_unacceptable(request):
+        raise HTTPNotAcceptable
+
+The appropriate view is selected here when the client specifies an unambiguous header such as ``Accept: text/*`` or ``Accept: application/json``.
+Similarly, if the client specifies a media type that no view is registered to handle, such as ``Accept: text/plain``, it will fall through to ``myview_unacceptable`` and raise ``406 Not Acceptable``.
+There are a few cases in which the client may specify ambiguous constraints:
```
whiteroses: In L1061 and L1063, the header is not ambiguous or unambiguous, it's just that a header could be specified such that more than one view matches? (So the ambiguity is in the view matching, rather than in the header itself.) Might it be worth rephrasing for people who are less familiar with Accept headers?

mmerickel: I just pushed some changes that I think make this more clear.

wichert:
I was hoping that this PR would introduce some general logic that we can use for language negotiation as well. Currently WebOb has deprecated that functionality, but nothing has replaced it as far as I can see.

mmerickel:
@wichert I can only imagine you missed the massive number of new features in webob 1.8 done by @whiteroses for GSOC last year that resulted in deprecatuing the old functionality. Is there something you were hoping to see in Pyramid specifically?

bertjwregeer:
@wichert 

Deprecated?

https://docs.pylonsproject.org/projects/webob/en/stable/api/webob.html#webob.acceptparse.create_accept_language_header

There's two algorithms for matching as specified in the RFC's:

https://docs.pylonsproject.org/projects/webob/en/stable/api/webob.html#webob.acceptparse.AcceptLanguageValidHeader.basic_filtering
https://docs.pylonsproject.org/projects/webob/en/stable/api/webob.html#webob.acceptparse.AcceptLanguageValidHeader.lookup

You can create and register a predicate to do language negotiation for any of your views.

stevepiercy:
Is issue #2807 related to language negotiation?

mmerickel:
@wichert I see, I missed the language part of your comment. It'd be interesting to explore algorithms to combine language, mimetype, charset, etc but that is not this PR. This PR is fixing long-standing complaints with just mimetype negotiation. Webob itself certainly introduced some nicer features around language that Pyramid is not using in any way by default right now.

wichert:
@bertjwregeer  The relevant method is `best_match()`, which WebOb has deprecated without a useful replacement. The documentation is inconsistent with the reality in that it says the method will be deprecated, but if you use it you already do get a deprecation warning in your output.

@mmerickel I haven't looked at the details, but I was expecting it to be possible to have a single matching function which can be used for language, mime type, transfer encoding, etc. which could act as a basis for this work. That might have been naive from me - I certainly haven't looked at the relevant RFCs.

mmerickel:
@wichert `request.accept_language.lookup` is documented as the replacement for best_match. If that api isn’t sufficient you should open an issue in webob

With respect to a full blown matching algorithm that combines all the accept headers you should also open an issue in webob but afaik there is not a well defined algorithm to do that. 

Regardless, please try to keep this issue and review focused on the work done with the accept header.

bertjwregeer:
There's no defined ordering/matching for all of the relevant Accept headers... Not sure how that would even work, because language matching has no relation to the mime-type of the returned document.

I am not even sure how Accept-Encoding would play into this, your application really shouldn't care about what transfer encoding is accepted by the remote client (unless you have pre-compressed assets on disk, maybe then).

Open an issue in WebOb if you want, but it will sit there forever, unfixed as there is no way to do what you want it to do.

----

As for `best_match` being deprecated, that's the way we deprecate things. We don't just docs deprecate it, we make it warn about deprecation and make a note of it in the documentation. It will be removed entirely in a future version of WebOb. There's two useful (and much better) replacements, I linked to the documentation on both of them.

bertjwregeer:
WebOb 1.8.3 has been released that contains the fixes necessary for this work to be released :-).

bertjwregeer:
One of the problems is that the title of this PR doesn't make it clear that this is only for dealing with the `Accept` header, not all `Accept-*` headers.

Sadly instead of calling it `Accept-Type` or something along those lines they just called it `Accept`.

mmerickel:
@wichert I think all of the accept headers affect different parts of the stack, it's unlikely you want one content negotiation step to rule them all.

- `Accept-Encoding` is usually handled by middleware.
- `Accept-Language` is usually handled by the template and anything using `request.localizer`.
- `Accept-Charset` is stupid imo and is handled better by just accepting specific media types via `Accept: text/html;charset=utf8` etc since it only really applies to specific text-based media types anyway.
- `Accept` is tied to which content type to return (as well as `Accept-Charset`) and thus belongs within the renderer itself, or the in the context of this pull request, used to select which view/renderer to use.

My point is that I think there's just not a situation where we would want to look at multiple accept headers at the same time and certainly it's not clear enough to framework it.

PR #3328, opened 2018-08-17
notfound view: use HTTPTemporaryRedirect as default

domenkozar:
So that it will redirect POST requests without failing out.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -99,6 +99,11 @@ Backward Incompatibilities
   longer override those set by the ``setup`` function.
   See https://github.com/Pylons/pyramid/pull/3318
 
+- ``pyramid.config.Configurator.add_notfound_view`` uses default redirect
+  class exception ``pyramid.httpexceptions.HTTPTemporaryRedirect`` instead
+  of previous ``pyramid.httpexceptions.HTTPFound``.
+  See https://github.com/Pylons/pyramid/pull/3328
+
 Documentation Changes
 ---------------------
 
--- a/pyramid/config/views.py
+++ b/pyramid/config/views.py
@@ -1631,8 +1631,8 @@ class ViewsConfiguratorMixin(object):
         instead implements :class:`~pyramid.interfaces.IResponse`, the
         append_slash logic will behave as if ``append_slash=True`` was passed,
         but the provided class will be used as the response class instead of
-        the default :class:`~pyramid.httpexceptions.HTTPFound` response class
-        when a redirect is performed.  For example:
+        the default :class:`~pyramid.httpexceptions.HTTPTemporaryRedirect`
+        response class when a redirect is performed.  For example:
 
           .. code-block:: python
 
@@ -1640,9 +1640,14 @@ class ViewsConfiguratorMixin(object):
             config.add_notfound_view(append_slash=HTTPMovedPermanently)
 
         The above means that a redirect to a slash-appended route will be
-        attempted, but instead of :class:`~pyramid.httpexceptions.HTTPFound`
+        attempted, but instead of :class:`~pyramid.httpexceptions.HTTPTemporaryRedirect`
         being used, :class:`~pyramid.httpexceptions.HTTPMovedPermanently will
         be used` for the redirect response if a slash-appended route is found.
+
+        :class:`~pyramid.httpexceptions.HTTPTemporaryRedirect` class is used
+        as default response, which is equivalent to
+        :class:`~pyramid.httpexceptions.HTTPFound` with addition of redirecting
+        with the same HTTP method (useful when doing POST requests).
 
         .. versionadded:: 1.3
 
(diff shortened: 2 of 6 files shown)


mcdonc:
I suspect we should bring back the concept of a generic NotFound and a generic Forbidden, and figure out some way to de-transition people from using the specific HTTP exceptions back to using the framework-generic ones.  I don't think we should change the default notfound response code until we do this.

domenkozar:
Alternatively, I wrote a small documentation patch to guide readers: https://github.com/domenkozar/pyramid/commit/b341a9d6fbaf00761469f140e5f51e3dad85360f

bertjwregeer:
Ehm... `HTTPNotFound` is a 404. If you raise a `HTTPNotFound` I am not sure how you would end up with a `307` in that code that is changed...

The change here looks to be safe and will change what used to be a 302 to a 307.

pyramid/config/views.py
```diff
@@ -1597,8 +1597,8 @@ def add_notfound_view(
         ):
         """ Add a default :term:`Not Found View` to the current configuration
         state. The view will be called when Pyramid or application code raises
-        an :exc:`pyramid.httpexceptions.HTTPNotFound` exception (e.g., when a
-        view cannot be found for the request).  The simplest example is:
+        an :exc:`pyramid.httpexceptions.HTTPTemporaryRedirect` exception (e.g.,
```
bertjwregeer: This is wrong... this should not be changed from HTTPNotFound.

pyramid/view.py
```diff
@@ -292,7 +292,7 @@ def notfound_view(context, request): return HTTPNotFound('nope')
     .. deprecated:: 1.3
 
     """
-    def __init__(self, notfound_view=None, redirect_class=HTTPFound):
+    def __init__(self, notfound_view=None, redirect_class=HTTPTemporaryRedirect):
```
bertjwregeer: This change is fine and won't cause any issues.

pyramid/config/views.py
```diff
@@ -1655,6 +1660,10 @@ def notfound(request):
         .. versionchanged:: 1.8
 
            The view is created using ``exception_only=True``.
+
+        .. versionchanged: 1.10
+
+           Default
```
bertjwregeer: Eh?

pyramid/config/views.py
```diff
@@ -1627,20 +1627,25 @@ def notfound(request):
         Pyramid will return the result of the view callable provided as
         ``view``, as normal.
 
+        :class:`~pyramid.httpexceptions.HTTPTemporaryRedirect` class is used
```
bertjwregeer: This paragraph feels weird, it should likely be documented under the `AppendSlashNotFoundViewFactory`.

Or maybe it should reference the fact that this is related to `append_slash`. Maybe remove the newline and continue the paragraph so it seems related? Not sure, but this currently reads weird to me.

bertjwregeer: Just a little further down, in the next paragraph we already explain which default we are using... maybe add the info about redirecting with the same HTTP method as a comment there?

pyramid/config/views.py
bertjwregeer: This is missing information.

pyramid/config/views.py
mmerickel: As I understand it this PR should only be changing the default for the `append_slash=True` behavior by modifying the `AppendSlashNotFoundViewFactory`'s `redirect_class` arg. This docs change does not make sense.

mcdonc:
Duh. Bert's right.  I still regret that 2011 commit but I retract my hold.

pyramid/config/views.py
domenkozar: You're obviously right, it was too late on Friday :)

pyramid/config/views.py
domenkozar: I didn't want to document it in `AppendSlashNotFoundViewFactory` since it's deprecated and I'd imagine 99% folks will be looking at docstring of `add_notfound_view`.

I'll move it below the next paragraph :)

domenkozar:
Thanks guys, I think this is ready for seconds pass (and includes test fixes).

One not so obvious breakage here is that anyone relying in their tests on 302 status code will break. An acceptabe nuisance?

bertjwregeer:
I would say we need to add a note to the CHANGES and to the release notes, but other than that I think this change is good to go.

domenkozar:
@bertjwregeer added item to  `CHANGES.rst`, but `docs/whatsnew-1.10.rst` doesn't exist.  Anything I missed?

CHANGES.rst
```diff
@@ -99,6 +99,11 @@ Backward Incompatibilities
   longer override those set by the ``setup`` function.
   See https://github.com/Pylons/pyramid/pull/3318
 
+- ``pyramid.config.Configurator.add_notfound_view`` uses default redirect
+  class exception ``pyramid.httpexceptions.HTTPTemporaryRedirect`` instead
+  of previous ``pyramid.httpexceptions.HTTPNotFound``.
```
mmerickel: the previous class was not HTTPNotFound

bertjwregeer:
@domenkozar `docs/whatsnew-1.10.rst` is created by the person doing the release, and until we are ready to release that document doesn't exist.

CHANGES.rst
bertjwregeer: Good catch!

PR #3327, opened 2018-08-16
Docs style guide

stevepiercy:
See #2880.

DO NOT MERGE.

For ease of viewing, visit the rendered docs at:
https://docs.pylonsproject.org/projects/docs-style-guide/

This PR is intended only for review, discussion, and gain exposure for what will become a new, separate project.

## Todos
- [x] Update https://pylonsproject.org/community-how-to-contribute.html with a new sentence at the end of the first paragraph: "Documentation contributions should follow the [Pylons Project Documentation Style Guide](https://docs.pylonsproject.org/projects/docs-style-guide/)."
- [x] Update https://pylonsproject.org/community-coding-style-standards.html with "Documentation Style Guide" heading and text: "Documentation contributions should follow the [Pylons Project Documentation Style Guide](https://docs.pylonsproject.org/projects/docs-style-guide/)."
- [x] Update first sentence in Pyramid's and docs-style-guide's [contributing.md](https://github.com/Pylons/pyramid/blob/master/contributing.md) with "How to Contribute, Coding Style and Standards, and [Pylons Project Documentation Style Guide](https://docs.pylonsproject.org/projects/docs-style-guide/).
- [x] Add issue templates to Pyramid, using those from [docs-style-guide](https://github.com/Pylons/docs-style-guide/tree/master/.github/ISSUE_TEMPLATE) as examples.

Issue #3263, opened 2018-04-23
Command line documentation style guide


stevepiercy:
## Question
Should we retain or remove command prompts from our documentation?

## Current practice
[Typographical Conventions](https://docs.pylonsproject.org/projects/pyramid/en/latest/typographical-conventions.html#syntax-highlighting) specifies our current practice, using command prompts to denote that we are using either a *nix-based OS (`$`) or a Windows OS (`c:>\`). Code block examples exist as reference for [*nix](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html#installing-pyramid-on-a-unix-system) and [Windows](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html#installing-pyramid-on-a-windows-system). In our tutorials where most of the commands appear, we state that we use the [*nix version of commands](http://localhost:63342/hack-on-pyramid/docs/_build/html/quick_tutorial/requirements.html):

> Commands in this tutorial use UNIX syntax and paths. Windows users should adjust commands accordingly.

And we use headings or similar indicator preceding the commands, but we could do a little better.

## Issues
This has a few unintended consequences.

* PowerShell is the new default shell in Windows 10 and later.
* Syntax highlighting for the Pygments lexer `doscon` is poor.
* Syntax highlighting will fail for the Pygments lexer `powershell` when a command prompt is used in the code block.
* *nix shell prompts are ignored.
* Command prompts are superfluous and get in the way of copy pasting.

## Proposal
The proposal is to do the following:

* Precede sections of commands with a heading or similar clear statement in which OS the command should be run.
* Remove command prompts from code-blocks.

See also #2880

stevepiercy:
Closed by #3330

PR #3330, opened 2018-08-19
Docs code style

stevepiercy:
See #3263 and https://pyramid-stevepiercy.readthedocs.io/en/docs-style-guide/documentation-style-guide.html#source-code from #3327


mmerickel:
@stevepiercy have you investigated any options for enforcing linting/style guides on the rst?

stevepiercy:
I had not considered adding some automated checking, but that's a good idea. I'd create a separate PR for this new feature.

I follow what [Plone](https://github.com/plone/documentation/issues/757) and [WriteTheDocs](http://www.writethedocs.org/guide/tools/testing/) do.

Some stuff can be automated without too much difficulty, and would be helpful and unobtrusive:
* [spellcheck](https://sphinxcontrib-spelling.readthedocs.io/en/latest/)
* [linkcheck](http://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder)
* [reST linter](https://pypi.org/project/restructuredtext_lint/)

There's also dozens of other projects that look for passive voice, "weasel words", and other linguistic violations, but I feel that's overkill and beyond our resources as an open source project to enforce. I think it would turn off contributors to have a Language Police robot.

If I were to implement any new CI stuff, I would probably create another environment, `docs-opt`, and allow failures until we have a good feel for whether it should be a hard fail, just for info, or dropped from the CI run.

I think that would be a good start because it's better than our simple `make doctest html epub` Sphinx builders.

Let me know your thoughts.

mmerickel:
@stevepiercy my interest would be in maintaining syntax consistency more than anything related to spelling/grammar.

PR #3336, opened 2018-08-27
Improve instructions of how to run p-scripts

stevepiercy:
- clarify two supported ways to run p-scripts
- add link in each script usage to the instructions. See https://github.com/Microsoft/vscode-python/issues/2427#issuecomment-416330081
- Use explicit lexer for syntax highlighting.
- Add name of script to each heading

### DIFF
--- a/docs/narr/commandline.rst
+++ b/docs/narr/commandline.rst
@@ -43,7 +43,7 @@ Each of Pyramid's console scripts may be run by its name.  For example:
 
 .. note:: ``$VENV/bin/`` is a convention we use to simplify Pyramid documentation.
     It represents the ``bin`` directory in a virtual environment, where ``$VENV`` is an environment variable representing its path.
-    See :ref:`installing_unix` and :ref:`venv-bin-pip-vs-source-bin-activate` for more information.
+    See :ref:`installing_unix` and :ref:`Why use $VENV/bin/pip instead of source bin/activate, then pip <venv-bin-pip-vs-source-bin-activate>` for more information.
 
 
 .. index::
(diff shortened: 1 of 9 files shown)


docs/narr/commandline.rst
```diff
@@ -6,15 +6,72 @@ Command-Line Pyramid
...
+
+.. _running-pscripts:
+
+Running p-scripts
+-----------------
+
+All of the Pyramid console scripts may be run either:
+
+*   by its name
+*   through a Python interpreter
```
mmerickel: "as a python module"

stevepiercy: Thanks. Fixed in d7dfe77.

docs/narr/commandline.rst
```diff
@@ -6,15 +6,72 @@ Command-Line Pyramid
...
+Running by ``p*`` script name
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Each of Pyramid's console scripts may be run by its name.  For example:
+
+.. code-block:: bash
+
+    $VENV/bin/pserve development.ini --reload
+
+.. note:: ``$VENV/bin/`` represents the ``bin`` directory in a virtual environment, where ``$VENV`` is an environment variable representing its path.
```
brettcannon: FYI `$VIRTUAL_ENV` is actually set as an environment variable when you are in an activated shell, so if you change it to that then `$VIRTUAL_ENV/bin/pserve` would be copy-and-pastable in UNIX.

brettcannon:
Thanks, @stevepiercy !

docs/narr/commandline.rst
stevepiercy: Thanks for the review, @brettcannon!

I clarified the note. `$VENV` is a convention we use throughout Pyramid documentation.

PR #3340, opened 2018-09-04
clarify session factory warnings

stloma:
I added a warning for the possibility of remote code execution, if your secret key is compromised, to the existing session warning.

With this addition, I felt it needed to be broken up into bullet points.

I'm open to feedback and changes if you all have a different idea of how it should be composed!

### DIFF
--- a/docs/narr/sessions.rst
+++ b/docs/narr/sessions.rst
@@ -69,7 +69,7 @@ using the :meth:`pyramid.config.Configurator.set_session_factory` method.
      session data is stored unencrypted in a cookie, it will also be easier for
      evildoers to obtain the current user's cross-site scripting token.
 
-   - The default serialization method, while replacable with something like
+   - The default serialization method, while replaceable with something like
      JSON, is implemented using pickle which can lead to remote code execution
      if your secret key is compromised.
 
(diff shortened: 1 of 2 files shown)


docs/narr/sessions.rst
```diff
@@ -57,18 +57,27 @@ using the :meth:`pyramid.config.Configurator.set_session_factory` method.
...
+     sensitive information in the session object, as the information can be
+     easily read by both users of your application and third parties who have
+     access to your users' network traffic.
+
+   - If you use this sessioning implementation, and you inadvertently create a
+     cross-site scripting vulnerability in your application, because the
+     session data is stored unencrypted in a cookie, it will also be easier for
+     evildoers to obtain the current user's cross-site scripting token.
+
+   - The default serialization method, while replacable with something like
```
mmerickel: `replaceable`

stevepiercy:
Thank you!

bertjwregeer:
If you declare the cookie as HTTP only then a cross site scripting vulnerability won't lead to the compromise of the cookie and won't lead to the compromise of the CSRF token contained within.

Also, generally cross site scripting will have full access to the page in question, so if there is a form on that same page which has the CSRF token as a form component they can already steal it, so they don't need the cookie.

stevepiercy:
@bertjwregeer should I revert this merge, or somehow integrate your remarks? I'm not sure what, if anything, should be changed in the docs.

bertjwregeer:
I think we can leave it for now, if you want to create a new PR I'd be happy to review the wording with my remarks added.

The wording on the first bullet point should also be revised, best practices dictate running a site using TLS, so people can't read the cookie from the wire.

PR #3344, opened 2018-09-08
Update setup.py

fuboki10:


fuboki10:
to make the code clearer

stevepiercy:
IMO, the names are clear already. Comments should add value or clarity where it is lacking. These comments are redundant.

PR #3358, opened 2018-09-18
Add registry arg to pyramid.renderers.get_renderer

### LINKED ISSUE #3354: add a way for get_renderer to avoid threadlocals
It should be as simple as adding a `registry=None` arg to `get_renderer` and forwarding that value to the `RendererHelper` and adding a test to exercise it.

cdunklau:
Fixes #3354 

This *seems* pretty straightforward, but I wonder if it can/should be tested more extensively?


cdunklau:
@mmerickel How does this look?

mmerickel:
LGTM, thanks again @cdunklau !

PR #3376, opened 2018-10-06
Issue 1747 add test quick tutorial 21

tomascorrea:
Add tests and documentation to quick tutorial.


docs/quick_tutorial/authorization/setup.py
```diff
@@ -2,11 +2,12 @@
 
 requires = [
     'bcrypt',
-    'pyramid',
+    'pyramid[testing]',
```
mmerickel: This should be removed, external apps should not depend on "whatever pyramid uses to run its own test suite". If you want to define some testing dependencies for this authorization app then normally we'd do this by specifying a "testing" extras requirement. See how it is done in the starter cookiecutter for an example. Specifically there is a `tests_require` defined in `setup.py` and instructions to run `pip install -e ".[testing]"` to install the requirements.

https://github.com/Pylons/pyramid-cookiecutter-starter/tree/latest/%7B%7Bcookiecutter.repo_name%7D%7D

docs/quick_tutorial/authorization/setup.py
```diff
@@ -7,8 +7,15 @@
     'waitress',
 ]
 
+tests_require = [
+    'pyramid[testing]'
```
mmerickel: no this needs to list out the dependencies you want to install... such as webtest. Please remove `pyramid[testing]`.

mmerickel: also as I said in the previous comment you need to update the instructions above to say `pip install -e ".[testing]"', not just `pip install -e .` so that these dependencies get installed.

stevepiercy: This is a common gotcha with the Quick Tutorial: most steps are based on previous steps' code and manually installed dependencies, yet we don't consistently update the `setup.py`'s dependencies to match the manual installation.

Tonight I will audit when dependencies are added manually against added to `setup.py` to ensure that confusion does not happen. I might also insert a step about how to manage dependencies manually without tools (because controversy is fun).

stevepiercy: I've completed my audit in PR #3379 and merged it into `master`.

@tomascorrea would you please compare these files against `master`, and either rebase or update, then push your changes to this PR?

docs/quick_tutorial/authorization.rst
```diff
@@ -74,6 +74,17 @@ Steps
 #. Visit http://localhost:6543/howdy in a browser. You should be asked to
    login.
 
+#. Lets test if the ``editor`` can access howdy and the the ``viewer`` can't:
```
stevepiercy: Two grammar corrections:

- Let's
- remove extra "the"

stevepiercy: @tomascorrea would also please sign CONTRIBUTORS.txt, if you have not done so? Thank you! This is good work!

mmerickel:
I'm going to close this soon due to inactivity.

PR #3378, opened 2018-10-06
Fixes issue #1428  - make partial functions work as view callables

### LINKED ISSUE #1428: partials as view callables behave differently under python 2.7 and python 3.4
Compare

``` python
Python 2.7.3 (default, Aug  1 2012, 05:14:39) 
[GCC 4.6.3] on linux2
>>> from pyramid.config.util import takes_one_arg
>>> from functools import partial
>>> def v(a, b, c):
...   return
... 
>>> takes_one_arg(partial(v, 1))
False
>>> takes_one_arg(partial(v, 1, 2))
False
>>> takes_one_arg(partial(v, 1, 2, 3))
False
```

with

``` python
Python 3.4.1 (default, Jul  7 2014, 09:21:41) 
[GCC 4.6.3] on linux
>>> from pyramid.config.util import takes_one_arg
>>> from functools import partial
>>> def v(a, b, c):
...   return
... 
>>> takes_one_arg(partial(v, 1))
True
>>> takes_one_arg(partial(v, 1, 2))
True
>>> takes_one_arg(partial(v, 1, 2, 3))
True
```

I was using a partial accepting two positional arguments as callable and my tests failed on python 3.4. So I think either partials should be disallowed as view callables, or `takes_one_arg` should be fixed to exhibit the same behaviour under both python 2.7 and python 3.4.

jsbueno:
What: Fixes #1428

How: Treat objectes decorated with functools.partial as special cases in `pyramid.util.takes_one_arg`.

(This PR is part of the "Pyramid Samba Sprint" taking place at Geru, in Sao Paulo)

### DIFF
--- a/pyramid/tests/test_util.py
+++ b/pyramid/tests/test_util.py
@@ -1096,6 +1096,31 @@ class Test_takes_one_arg(unittest.TestCase):
         getattr(foo, '__annotations__', {}).update({'bar': 'baz'})
         self.assertTrue(self._callFUT(foo))
 
+    def test_partial_decorated_func(self):
+        from functools import partial
+
+        def func(a, b, c):
+            pass
+
+        self.assertFalse(self._callFUT(partial(func)))
+        self.assertFalse(self._callFUT(partial(func, 1)))
+        self.assertTrue(self._callFUT(partial(func, 1, 2)))
+        self.assertFalse(self._callFUT(partial(func, 1, 2, 3)))
+
+    def test_partial_decorated_method(self):
+        from functools import partial
+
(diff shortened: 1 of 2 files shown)


stevepiercy:
@jsbueno tests do not pass yet. Suggest running tests locally using `tox -e py37` (or whatever python you have installed) as described in HACKING.txt.

Also would you please sign CONTRIBUTORS.txt and push that to this PR?

Once tests pass and we get the signature, then we can merge your work. Thank you!

mmerickel:
I'm going to close this soon due to inactivity.

PR #3379, opened 2018-10-07
Docs quick tutorial update

stevepiercy:
This PR (hopefully) eliminates the common gotcha where users jump into later steps without installing requirements in earlier steps. 

- `setup.py` includes all requirements as needed.
- Narrative text no longer mentions manual installation of dependencies.
- Standardize formatting with Documentation Style Guide.
- Setuptools updates for its glossary term.


mmerickel:
I would have really preferred if you had used `testing` as the name of the extra to be consistent with every repository we maintain as well as the pattern set forth in the cookiecutters.

stevepiercy:
No problem. I'll make changes in another PR, along with moving the pyramid_debugtoolbar into `dev` in extras.

docs/quick_tutorial/functional_testing.rst
```diff
@@ -31,31 +31,40 @@ Objectives
...
+        :emphasize-lines: 14
 
-   Be sure this file is not executable, or ``pytest`` may not include your
-   tests.
+#.  Install our project and its newly added dependency.
+    Note that we use the extra specifier ``[test]`` to install testing requirements.
+
+    .. code-block:: bash
+
+        $VENV/bin/pip install -e .[test]
```
mmerickel: This should use quotes, such as `pip install -e ".[test]"`.

docs/quick_tutorial/jinja2/setup.py
```diff
@@ -2,13 +2,22 @@
...
+              'webtest',
+          ],
+      },
       entry_points="""\
       [paste.app_factory]
       main = tutorial:main
       """,
-)
\ No newline at end of file
+      )
```
mmerickel: Why are all these closing parenthesis indented? They should not be.

stevepiercy: PyCharm has an inspection for PEP8. Either indentation form is acceptable.
https://www.python.org/dev/peps/pep-0008/#indentation

stevepiercy:
Revisions are now in PR #3380.

PR #3380, opened 2018-10-08
Revise Quick Tutorial `setup.py`s

stevepiercy:


stevepiercy:
FWIW, I have no problem doing the renaming.

I would much rather have a well-named extra than one that makes me go "wtf? pyramid_debugtoolbar is neither a production nor a testing requirement! I'm just trying to develop my first Pyramid app!"

blaflamme:
I think `tests` or `testing` still makes more sense than `dev` regarding tests dependencies. I agree that while you're developing you normally need tests dependencies but `dev` is more a state of a project than `tests` for declaring a set of specific dependencies. As an example it makes more sense to install `prod` deps + `tests` deps for the CI before deploying to production than installing `dev` deps that hides `tests` deps for production. I still prefer to be explicit than implicit here.

mmerickel:
What I have seen from the packaging community is that there has been a real push for applications to focus only on dev and prod requirements and to not split it up into more fine-grained extras like testing, linting, publishing, etc. This is different for libraries imo, but for apps the stance has been pretty clear that the two categories satisfies most users needs. Bundler, NPM, and Cargo also have clearly gotten pretty far using that strategy. Note that both pipenv and poetry has taken this stance as well, although poetry does support extras as well as an escape hatch if needed.

blaflamme:
I also agree about the difference between library and application. While I still prefer a fine-grained approach myself I can see the benefit of sharing what others are doing, less brain shifting. I have not started using `pipenv`, I'm still not sure about it bringing the best of all packaging worlds, `poetry` seems to be a good alternative.

bertjwregeer:
I personally have started using `poetry` for new projects and am all for dropping the extra distinction, it makes things cleaner. On CI I install develop, on production I install the final package without any extras.

I think just `dev` instead of `testing` is a great idea. 👍

stevepiercy:
Is this the consensus?

## Example before:

```
requires = [
    'bcrypt',
    'pyramid',
    'pyramid_chameleon',
    'pyramid_debugtoolbar',
    'waitress',
]

tests_require = [
    'pytest',
    'webtest',
]

 setup(
    name='tutorial',
    install_requires=requires,
    extras_require={
        'testing': tests_require,
    },
)
```

## Example after:

```
requires = [
    'bcrypt',
    'pyramid',
    'pyramid_chameleon',
    'waitress',
]

dev_requires = [
    'pyramid_debugtoolbar',
    'pytest',
    'webtest',
]

 setup(
    name='tutorial',
    install_requires=requires,
    extras_require={
        'dev': dev_requires,
    },
)
```

mmerickel:
LGTM

bertjwregeer:
:+1:

stevepiercy:
Changes made. Ready for final review.

PR #3397, opened 2018-10-19
refactor configurator actions and predicates into mixins

mmerickel:
basically this just gets rid of `pyramid.config.util` and moves a bunch of code from `pyramid.config` into `pyramid.config.actions` and `pyramid.config.predicates`

I left an `ActionState` shim in `pyramid.config` to avoid breaking `pyramid_zcml`.


bertjwregeer:
👍

PR #3402, opened 2018-10-27
Throw 401 for security denial for unauthenticated

bostrick:
The security framework currently throws HTTPForbidden for security failures, regardless of user context.

To better align with security frameworks, such as repoze.who, and internet convention, pyramid should throw 401 for unauthenticated users, to give them the opportunity to log in and try again....

### DIFF
--- a/src/pyramid/viewderivers.py
+++ b/src/pyramid/viewderivers.py
@@ -2,7 +2,7 @@ import inspect
 
 from zope.interface import implementer, provider
 
-from pyramid.security import NO_PERMISSION_REQUIRED
+from pyramid.security import NO_PERMISSION_REQUIRED, Authenticated
 from pyramid.csrf import check_csrf_origin, check_csrf_token
 from pyramid.response import Response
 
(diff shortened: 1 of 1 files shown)


bostrick:
... or should this be handled somewhere further up by catching and rethrowing?  user principles are in hand here...

mmerickel:
We cannot, in general, return a `401 Unauthorized` because the spec dictates

```
The server generating a 401 response MUST send
a WWW-Authenticate header field (Section 4.1) containing at least one
challenge applicable to the target resource.
```

We cannot, in general, return an appropriate `WWW-Authenticate` header as it's only really defined for HTTP Basic authentication workflows.

The proper way to handle this in your application is to register a forbidden view (see `config.add_forbidden_view`) and return a 401 error instead of a 403 if you wish.

```
@forbidden_view_config()
def forbidden_view(request):
    if request.authenticated_userid is not None:
        return HTTPUnauthorized()
    return HTTPForbidden()
```

bostrick:
understood... thanks for background and perspective!  --b

Issue #3383, opened 2018-10-10
remove support for media ranges in accept predicates


mmerickel:
This feature was decided to be impossible to implement without corner cases via discussions in https://github.com/Pylons/pyramid/pull/3326#discussion_r223197461 and were deprecated in 1.10, removed in 2.0.

- [ ] remove from `add_route`
- [ ] remove from `add_view`
- [ ] remove deprecated behavior from `AcceptPredicate`
- [ ] remove `allow_range` from `normalize_accept_offer`

PR #3411, opened 2018-11-02
remove deprecated media range support from add_view and add_route

mmerickel:
fixes https://github.com/Pylons/pyramid/issues/3383


stevepiercy:
I poked AppVeyor, and it got over the network issue. Looking good so far.

PR #3413, opened 2018-11-03
change to use JSONSerializer for SignedCookieSessionFactory

### LINKED ISSUE #2709: 2.0 feature request: Require that sessions are JSON serializable
Currently the `ISession` interface requires that that keys and values of a session must be pickleable.

My proposal is to drop that requirement and simply state that all keys and values should be JSON serializeable. We should still give users the option to use the pickle serializer, but pick a more safe default.

---

Cookie based sessions right now by default use a PickleSerializer and then the cookie is protected using a secret and HMAC hashes. However if an attacker was able to get the the secret, they could potentially use the unpickle methods to execute code in the context of the running Python process.

mmerickel:
fixes #2709

### DIFF
--- a/docs/narr/sessions.rst
+++ b/docs/narr/sessions.rst
@@ -25,7 +25,7 @@ Using the Default Session Factory
 In order to use sessions, you must set up a :term:`session factory` during your
 :app:`Pyramid` configuration.
 
-A very basic, insecure sample session factory implementation is provided in the
+A very basic session factory implementation is provided in the
 :app:`Pyramid` core.  It uses a cookie to store session information.  This
 implementation has the following limitations:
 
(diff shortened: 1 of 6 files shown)


docs/narr/sessions.rst
```diff
@@ -38,8 +38,8 @@ implementation has the following limitations:
   of the session is fewer than 4000.  This is suitable only for very small data
   sets.
 
-It is digitally signed, however, and thus its data cannot easily be tampered
-with.
+It is digitally signed, however, and thus a client cannot easily tamper with
+the content without compromising the secret key.
```
bertjwregeer:
```suggestion
the content without having access to the secret key.
```

bertjwregeer:
:+1:

PR #3415, opened 2018-11-04
support mounting routes without a trailing slash

### LINKED ISSUE #406: route_prefix doesn't allow the prefix pattern to match without a trailing slash

mmerickel:
This PR replaces #1639. Closes #406.

The gist is this snippet that was added to the docs.

```python

def users_include(config):
    config.add_route('show_users', '')

def main(global_config, **settings):
    config = Configurator()
    config.include(users_include, route_prefix='/users')

The above configuration will match ``/users`` instead of ``/users/`` if a slash
had been supplied to the :meth:`pyramid.config.Configurator.add_route` function.
```

This is backward-incompatible for anyone mounting routes and using an empty string.

Alternative approaches that could maintain compatibility:

1) Support a sentinel value as the pattern. For example:

```python
from pyramid.config import NO_APPEND_SLASH

def users_include(config):
    config.add_route('show_users', NO_APPEND_SLASH)
```

This is somewhat cumbersome.

2) Add an option to `add_route` that a user could opt into.

```python
def users_include(config):
    config.add_route('show_users', '', route_prefix_append_slash=False)
```

I'm open to something like this and am curious what people think. Pyramid has a somewhat ingrained history of supporting `'' == '/'` and this makes them different, but only in a specific use case. When used from main or without a route_prefix, they are still equivalent and result in a trailing slash. I'm not convinced it's worth the incompatibility.

ping @nek4life

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,12 @@ Features
   documentation for more information about why this change was made.
   See https://github.com/Pylons/pyramid/pull/3413
 
+- It is now possible to declare a route to not contain a trailing slash
+  when it is mounted via ``config.include(..., route_prefix=...)`` or via
+  ``with config.route_prefix_context(...)`` by specifying a ``pattern`` of
+  ``''``. This change is backward incompatible, see the notes below.
+  See https://github.com/Pylons/pyramid/pull/3414
+
 Bug Fixes
 ---------
 
@@ -67,6 +73,15 @@ Backward Incompatibilities
   "Changes to ISession in Pyramid 2.0" in the "Sessions" chapter of the
   documentation for more information about why this change was made.
   See https://github.com/Pylons/pyramid/pull/3413
+
+- Changed the URL generation and matching for a route with a ``pattern`` of
+  ``''`` when the route is mounted with a ``route_prefix`` via either
+  ``config.include(..., route_prefix=...)`` or
+  ``with config.route_prefix_context(...)``. This route will now match a bare
+  URL without a trailing slash. To preserve the old behavior, set the
+  ``pattern`` to ``'/'`` instead of the empty string and the URL will contain
+  a trailing slash. This only affects mounted routes using ``route_prefix``.
+  See https://github.com/Pylons/pyramid/pull/3414
 
 Documentation Changes
 ---------------------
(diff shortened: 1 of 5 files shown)


stevepiercy:
Are you talking about only prefix-based routes, only routes without prefixes, or both?

I define routes with leading and trailing slashes to avoid any guesswork, so I don't have an opinion about this feature.

mmerickel:
closing in favor of #3420

PR #3420, opened 2018-11-12
support inherit_slash=True on add_route

mmerickel:
This is an alternative to https://github.com/Pylons/pyramid/pull/3415 which is more explicit and backward compatible. To opt-out of the default and to get a bare url without a trailing slash, specify `inherit_slash=True` to `add_route`. This triggers an empty pattern to inherit the trailing slash from the route prefix. This means it will have a trailing slash or not depending on the route prefix, giving control to the integrator to decide if the route has a slash or not.


src/pyramid/config/routes.py
```diff
@@ -139,6 +140,26 @@ def add_route(
...
           .. versionadded:: 1.1
 
+        inherit_slash
+
+          This argument can only be used when the ``pattern`` is an empty
+          string (``''``). By default, the composed route pattern will always
+          includes a trailing slash, but this argument provides a way to
+          opt-out if both, you and the integrator (the developer setting the
+          :term:`route prefix`), agree that the pattern should not contain
+          a trailing slash. For example:
```
stevepiercy: I have a hard time parsing this paragraph so I simplified it. I assume that I am both "you" and "the integrator", but maybe that's deliberate. I also made some grammar fixes.

```
... include a trailing slash.
However this argument provides a way to opt out of the default and instead make the pattern not contain a trailing slash.
For example:
```

CHANGES.rst
```diff
@@ -12,6 +12,23 @@ Features
   documentation for more information about why this change was made.
   See https://github.com/Pylons/pyramid/pull/3413
 
+- It is now possible to control whether a route pattern contains a trailing
+  slash when it is composed with a route prefix using
+  ``config.include(..., route_prefix=...)`` or
+  ``with config.route_prefix_context(...)``. This can be done by specifying
+  an empty pattern and setting the new argument
+  ``inherit_slash=True``. For example::
```
stevepiercy: If possible, please use the following to explicitly set the Pygments lexer to Python:

```
  ... For example:

  .. code-block:: python
```

src/pyramid/config/routes.py
```diff
@@ -139,6 +140,26 @@ def add_route(
...
+
+          .. code-block:: python
+
+              with config.route_prefix_context('/users'):
+                  config.add_route('users', '', inherit_slash=True)
+
+          In this example, the resulting route pattern will be ``/users``.
+          Alternatively, if the route prefix were ``/users/`` then the
```
stevepiercy: Comma placement:

```
Alternatively if the route prefix were ``/users/``, then the
```

CHANGES.rst
mmerickel: We have a history of only using bare-minimum syntax in the changes file so that it always renders ok on PyPI. If you want to check and see if PyPI supports code-block extensions now be my guest but I'll leave it this way until then.

bertjwregeer:
:+1:

src/pyramid/config/routes.py
mmerickel: I kept the comma after alternatively, I'm pretty sure it's valid here.

src/pyramid/config/routes.py
mmerickel: I tried to improve this. Your proposed changes are not correct, opting into inheritance means there are two sides and I have to delineate them.

src/pyramid/config/routes.py
stevepiercy: Both usages are valid. It's a modern style preference in English to use fewer commas. Commas should be used as "speed bumps", where you want the reader to pause. It's similar to the [breath mark in music](https://www.symbols.com/symbol/breath-mark).

CHANGES.rst
stevepiercy: It does. See https://raw.githubusercontent.com/Pylons/pylons-sphinx-themes/master/README.rst and https://pypi.org/project/pylons-sphinx-themes/

mmerickel: sounds good, fixed

src/pyramid/config/routes.py
mmerickel: ok, I'm going to keep it the way it is

PR #3429, opened 2018-11-24
Clarify startup and logging related docs

kpinc:
Split long sentences.
Shorten and clarify sentences by using links.
Add index entries.

Please see if the right words are used in the index entries.
Thanks.


docs/narr/startup.rst
```diff
@@ -47,10 +47,15 @@ Here's a high-level time-ordered overview of what happens when you press
...
    the format.
 
+.. _ini_section_discovery:
+
 #. The :term:`PasteDeploy` finds a section named either ``[app:main]``,
-   ``[pipeline:main]``, or ``[composite:main]`` in the ``.ini`` file.  This
-   section represents the configuration of a :term:`WSGI` application that will
-   be served.  If you're using a simple application (e.g., ``[app:main]``), the
+   ``[pipeline:main]``, or ``[composite:main]`` in the ``.ini`` file.
+   This section becomes the ``ini`` section for the :term:`WSGI` webserver portion of your application -- the part ``Pyramid`` was built to help with.
```
stevepiercy: I think the original sentence is clearer and more concise.

p.s. -- If you use ` -- `, please instead use an em-dash `—` without spaces around it.

docs/narr/startup.rst
```diff
@@ -63,17 +68,20 @@ Here's a high-level time-ordered overview of what happens when you press
...
    file and uses it to configure the Python standard library logging system for
-   this application.  See :ref:`logging_config` for more information.
+   the application.  See :ref:`logging_config` for more information.
 
-#. The application's *constructor* named by the entry point referenced on the
-   ``use=`` line of the section representing your :app:`Pyramid` application is
-   passed the key/value parameters mentioned within the section in which it's
-   defined.  The constructor is meant to return a :term:`router` instance,
-   which is a :term:`WSGI` application.
+#. The application's entry point :ref:`found above <entry_point_discovery>`, usually the entry point referenced on the above mentioned ``use=`` line, is the application's *constructor*.
```
stevepiercy: Remove redundancy. Fix grammar.

```
The application's entry point, usually the entry point referenced on the above mentioned ``use=`` line, is the application's *constructor*.
```

stevepiercy: We don't define a term for "constructor" in the Glossary. It would be good to add one.

mmerickel:
This is fine, it's not strictly correct but it's accurate for 99% of use cases. In reality `pyramid.paster.setup_logging` just invokes `plaster.setup_logging` and thus does whatever the loader says, which is usually `plaster_pastedeploy` for `ini` files, and `plaster_pastedeploy` calls `logging.config.fileConfig`. Describing all the levels of indirection going on is something we should do somewhere in the future and just point to that but for now this is fine.

PR #3430, opened 2018-11-24
Fix REST indentation so numbered list does not restart counting

kpinc:
Sorry about letting this bug sneak into the last pull request.


stevepiercy:
Sneaky reST. Thank you!

PR #3436, opened 2018-11-27
Remove obsolete intersphinx references

stevepiercy:


stevepiercy:
s/obsolete/unused so that we don't download the `.inv` files, speeding up docs builds

PR #3442, opened 2018-12-26
changed  principals_allowed_by_permission to act differently if permissive is True or False

Estartu:
principals_allowed_by_permission of the DummySecurityPolicy should return an empty list if permissive is set to False as no Principal is allowed


mmerickel:
Thanks! I'm leaving this open until we backport this to 1.10-branch as I consider it a bug fix.

PR #3421, opened 2018-11-13
initial work to remove py2 from the codebase

mmerickel:
- [X] remove references to PY2
- [X] remove conditional imports
- [x] figure out what to do with `pyramid.compat` going forward (more on this below)
- [X] remove api docs for `pyramid.compat`
- [X] remove usages of `pyramid.compat.native_`
- [x] remove usages of `string_types`, `integer_types`, and other `foo_types` from `pyramid.compat`
- [X] replace usages of `pyramid.compat` in codebase
- [X] fix starter  cookiecutter
- [x] fix docs for supported versions
- [x] fix wiki tutorial
- [x] fix wiki2 tutorial
- [x] fix quick tutorial
- [x] fix quick tour docs
- [x] fix docstrings / api docs for affected methods (track down references to bytes, unicode or python versions)

One note is that it's important to remember to keep using `text_` and `bytes_` where necessary for apis that may legitimately accept bytes or strings.

**what to do about pyramid.compat**

`pyramid.compat` is a public api and addons rely on it for some compatibility features.

If we want to remove it, we should push a bugfix release in pyramid that deprecates the module to help people when upgrading.

closes https://github.com/Pylons/pyramid/issues/2903

PR #3443, opened 2018-12-29
Restore compat.py in master/2.X

mcdonc:
Yeah it was never an API.  But we used it a lot, and apps that depend on it can't immediately migrate wholesale to py3 only.  It's maybe a bit fundamentalist to remove it.  Maybe it should be an API.


mmerickel:
I did at little research before deleting it and at least the main pyramid addons I use didn't depend on it. I'm sure it broke substanced though. With respect to "yeah it was never an API" it actually was an api and removing it was part of the bw-incompat changes in 2.x. We just forgot to update the changelog (for any of the changes, even dropping py2) in https://github.com/Pylons/pyramid/pull/3421. I opened https://github.com/Pylons/pyramid/issues/3444 to remind me to do that before any release.

I'm not going to put up much fight against this but I will say that I think it shouldn't be here. At the very least this module should raise a deprecation warning with an intent to remove at some point. Either that or please update Pyramid's codebase to use the functions in this module and remove the ones I added to `pyramid.util`. IMO addons should vendor in their own compat or use six or whatever.

bertjwregeer:
I'm a :-1: on bringing `pyramid.compat` back, and I am saying that as someone that uses it quite a bit in his own code bases. Although so far the migration has been pretty simple because removing py2 compat means not needing this module anymore.

mcdonc:
> I'm a  on bringing `pyramid.compat` back, and I am saying that as someone that uses it quite a bit in his own code bases. Although so far the migration has been pretty simple because removing py2 compat means not needing this module anymore.

Understood for personal codebases but frameworky bits like pyramid_debugtoolbar can't really just up and ditch Python 2.X unless we declare some sort of flag day.  Alternately, wee could change it so it had its own compat stuff. But it's just easier for people (and us) if it doesn't break when they use it with 3.X or 2.X so we don't need to make noise Pyramid 2.X compat releases of things when it would continue to work just fine if compat was there.

mcdonc:
> I'm not going to put up much fight against this but I will say that I think it shouldn't be here. At the very least this module should raise a deprecation warning with an intent to remove at some point. Either that or please update Pyramid's codebase to use the functions in this module and remove the copies I added to `pyramid.util`. IMO addons should vendor in their own compat or use six or whatever. I really don't want to have code in Pyramid that Pyramid doesn't use and doesn't have great tests, but if they are deprecated I'm less worried about it.

Deprecation warnings sound like a good idea to me.  But as should be clear by me not just checking this in on master, I'm willing to defer too.

mmerickel:
> Understood for personal codebases but frameworky bits like pyramid_debugtoolbar can't really just up and ditch Python 2.X unless we declare some sort of flag day.

Answering this part in a vacuum, unrelated to this PR: this is less of a problem than you might think thanks to the `python_requires` metadata. We could technically (of course it's a fairly irresponsible) drop py2 on a minor version and it wouldn't affect anyone using a remotely modern (8-ish) version of pip where it simply wouldn't install the new release. Dropping it on a major version though shouldn't be an issue and is something I'd plan to do to addons as they get maintained.

As far as this PR goes, I'm +0 merging this with deprecation warnings.

bertjwregeer:
`pyramid_debugtoolbar` can be updated to support only Python 3.x too, so it is imho less of an issue, or it could even for its next release vendor its own version of compat.py into an existing point release and then the next major version is Py3 only.

It feels really weird to me to keep this around when it is no longer internally used by Pyramid, especially when we are doing a major version jump specifically because we are breaking backwards compatibility by removing Python 2 support.

mcdonc:
> It feels really weird to me to keep this around when it is no longer internally used by Pyramid, especially when we are doing a major version jump specifically because we are breaking backwards compatibility by removing Python 2 support.

But the intent of supporting only Py 3 isn't to explicitly break Py 2 compatibility.  It's not being done as a political statement, it's being done to decrease maintainer burden.  If we could move to Py 3 without breaking bw compat in any way, we would.  The burden of maintaining compat.py is close to zero, and if having it around keeps things working for people who want to support Py 2 and Py 3 versions of their software that depends on Pyramid, I see no practical reason to remove it. IMO, the license to break bw compat should be relatively conservative even at major version bumps, esp. when it doesn't hurt us in any way.

mmerickel:
If you want to keep it that's fine. Please either 1) add the deprecation warning and pick some version where you're cool with removing it, or 2) bring it back to the standard it was at before I removed it in #3421. It had api docs and full tests before which this PR omits. In either scenario, a decision needs to be made where the code lives. Either in `pyramid.compat` or in `pyramid.util`. Maintaining copies in both places isn't going to fly. Personally I'd say that `pyramid.compat` should be a thin shim just importing the functions from `pyramid.util` and a deprecation warning added to `pyramid.compat`.

mmerickel:
I found at least that pyramid_mako is using pyramid.compat. https://github.com/Pylons/pyramid_mako/issues/41

bertjwregeer:
I haven't given `pyramid_mako` much love.

mmerickel:
@mcdonc what's the plan?

mcdonc:
Sorry for the late reply.  What's attached now to this PR is I think the only stuff that was  there before (and is required) with some updates to the docs.

mmerickel:
Awesome. We now have some duplicate methods defined in `pyramid.util`. The last step in this PR is that I think we should re-export those from `pyramid.compat` instead of maintaining these copies.

tests/test_compat.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class OldStyle:
+    def run(self):  # pragma: no cover
+        return 'OK'
+
+
+class NewStyle(object):
+    def run(self):  # pragma: no cover
+        return 'OK'
```
mmerickel: Python 3 only has new-style classes, some of these tests can be removed.

src/pyramid/compat.py
```diff
@@ -0,0 +1,188 @@
...
+binary_type = bytes
+long = int
+
+
+def text_(s, encoding='latin-1', errors='strict'):  # pragma: no cover
+    """ If ``s`` is an instance of ``binary_type``, return
+    ``s.decode(encoding, errors)``, otherwise return ``s``"""
+    if isinstance(s, binary_type):
+        return s.decode(encoding, errors)
+    return s
```
mmerickel: These functions are already defined in `pyramid.util`. Please import them from there and re-export them instead of re-defining them here.

mmerickel:
reopen this if you feel like addressing comments

Issue #3422, opened 2018-11-18
Auth post-mortem (Pyramid 2.0)


luhn:
Hi guys!  I'm interested in taking a crack at reworking the auth system for Pyramid 2.0, based on the [Auth API Post-Mortem](http://web.archive.org/web/20170201093432/plope.com/pyramid_auth_design_api_postmortem).  Would you be open to starting a discussion?

I like Pyramid a lot and would love to see a new auth API in 2.0, as mentioned in #2362.  Hopefully I'm not being too brash, barging through the front door as a first-time contributor and asking to remodel kitchen.  I want to be respectful of your time; once we hash out the details I'm confident I can make it happen independently with minimal time investment from you.

bertjwregeer:
We'd be more than happy to see someone take it on and try it. The only concerns that we have from core is that we want to make sure that it is backwards compatible and that may be an issue since the API's don't map well to each other.

It's something that has been discussed quite a bit on IRC, and in private, but there are some other open issues around authn/authz, mostly related to being able to access the request object.

luhn:
Hmm... Backwards compatibility will be a bit awkward, but I think doable.  I'll see if I can sketch something up.

luhn:
Okay, here's a quick sketch:  https://github.com/Pylons/pyramid/compare/master...luhn:auth-sketch

- Old-style auth interfaces are kept the same
- New-style auth interfaces are `INewAuthenticationPolicy` and `INewAuthorizationPolicy` (awkward, but inevitable for strict bw compat)
- Secure view deriver exclusively uses new-style auth
- New-style compatibility policies are implemented (and would be the default) which shims the old-style policies
- The final product would include an ACL helper as discussed in the post-mortem.

The new interfaces deviate from the post-mortem, since I figured this sketch was a chance to exercise some creativity/opinions.  The sketch could be adjusted to better match the post-mortem if that's what the core wants.

- I did not drop AuthenticationPolicy in favor of IdentityPolicy, it helps minimize the differences between the new and old auth.
- I merged `unauthenticated_userid` and `authenticated_userid` into `get_user`.  Not sure how this is going to be received... I personally have never found unauthenticated identity (i.e. `unauthenticated_userid`) useful.  Plus, it's the authentication API, isn't unauthed anything out of scope?
- I used `get_user` instead of `get_userid` because whether the auth policy returns an ID or a full user object is IMO an implementation detail.  (Really just semantics, it doesn't influence the implementation)
[...]

PR #3454, opened 2019-01-21
[WIP] [IDEA] Auth sketch 3

bertjwregeer:
Creating a PR of the auth sketch as set out in #3422 to be able to leave comments/get feedback on the code changes and see what shakes out.


src/pyramid/interfaces.py
```diff
@@ -482,8 +482,40 @@ def __call__(self, **kw):
         """
 
 
+class ISecurityPolicy(Interface):
+    def identify(request):
+        """ Return an object identifying a trusted and verified user. """
+
+    def permits(context, request, identity, permission):
```
mmerickel: I think for consistency I'd reverse the signature and put request before context.

src/pyramid/security.py
```diff
@@ -363,3 +363,50 @@ def has_permission(self, permission, context=None):
             )  # should never happen
         principals = authn_policy.effective_principals(self)
         return authz_policy.permits(context, principals, permission)
+
+
+@implementer(ISecurityPolicy)
+class CompatibilitySecurityPolicy(object):
```
mmerickel: A naming convention nit: `LegacySessionCSRFStoragePolicy` exists to make csrf storage work like it used to. Probably use `Legacy` instead of `Compatibility` here as well.

src/pyramid/security.py
```diff
@@ -363,3 +363,50 @@ def has_permission(self, permission, context=None):
...
+        """ Remove the stored userid from the session."""
+        if self.identity_key in request.session:
+            del request.session[self.identity_key]
+        return []
+
+    def permits(self, context, request, identity, permission):
+        raise NotImplementedError(
+            '`SessionSecurityPolicy` does not provide an implementation for '
+            '`permits`.  You may wish to make your own subclass.'
+        )
```
mmerickel: Instead of shipping `SessionSecurityPolicy` and other variants like `AuthTktSecurityPolicy` I think the goal should be to convert those policies to things closer to the `AuthTktCookieHelper` that can be used by people to write their own security policies.

luhn:
Thanks for the feedback guys, I've made the adjustments.

Anything else?  Should I start on making this a real implementation?

mmerickel:
@luhn I'm +1 on pursuing it myself. IMO there are still other discussions about caching, privilege escalation (session invalidation / caching), vary headers, remember/forget, etc but if it helps to work on this incrementally then you should go for it.

bertjwregeer:
I would love to see this as a real implementation, and I love the ideas presented here. It makes things much simpler and does away with the dual interface stuff that made it difficult to really understand where and what actions were being taken, and where 1 part had access to the request but the other did not.

I think that session invalidation/caching/privilege escalation are all important and we should discuss them, but I don't think it should stand in the way of this improvement.

mmerickel:
It’s worth voicing one more time. I’m concerned about dropping some api to acquire a userid. It’s something addons use in their logging systems right now. For example sentry_sdk and pyramid_tm and I’m sure many others. It’s also used in tutorials as a way to check if a user is logged in which the identity is a fine replacement for but the point remains that it’s a feature that sees some legitimate use that I’m not sure should be left out of the new api.

bertjwregeer:
Just as a note: 

- `pyramid_tm` uses `unauthenticated_userid`: https://github.com/Pylons/pyramid_tm/blob/8579fbee520a5653d4f36559ed09d641093b4400/pyramid_tm/__init__.py#L128
- `sentry_sdk` uses `authorized_userid`: https://github.com/getsentry/sentry-python/blob/1c41dfbfcdb3e58a7467b73af80bebf0356821e3/sentry_sdk/integrations/pyramid.py#L155 and https://github.com/getsentry/sentry-python/blob/1c41dfbfcdb3e58a7467b73af80bebf0356821e3/sentry_sdk/integrations/pyramid.py#L21

The former is almost meaningless IMHO, since if it is unauthenticated it may or may not be valid, its a nice to have. The latter makes sense, but why could one not use `identity` for this?

I would expect that anything that is using it directly should wrap it with `str()` or maybe even `repr()` so that the `identity` can decide how to present itself to the outside world (and thus Sentry and or pyramid_tm).

I will admit that `unauthenticated_userid` in all of the policies I've written has always returned `None`, so I am ambivalent about that use case.

Having said that, if the API contained a way to return a `userid` would that need to be a string? Could I for instance return the `repr()` of an `identity` in its stead? Does it have any uniqueness requirements?

mmerickel:
I’m not against using `identity.__str__` for this so long as it’s clearly documented as being used for logging purposes.

luhn:
`identity.__str__` sounds good to me, I'll document it.

I'm starting the real implementation over at #3465.

mcdonc:
I mentioned this over in #3465 but I am not a fan of gluing authentication and authorization together.  I think the separation we gave them was right, and I see no reason to change that bit of design.  I did have concerns, but my concern was with the implementation of the actual authentication and authorization policies using composition (e.g. for groupfinder callback) instead of just telling people to subclass the built-in ones.  And since I wrote that screed I've come to think that changing things as described in that mea culpa post would be worse than not changing things, as we'd just be moving around deck chairs without any significant improvement.  It's just a hard problem and moving machinery around between plugin classes does not make it any less hard.

mmerickel:
The largest significant improvement is getting away from principal strings as the transport data between authn and authz which is limiting compared to an opaque object such as macaroons or other data. The new policy will also encourage the usage of composition via helper objects instead of subclassing.

mcdonc:
I'll review the stuff over in #3422 which seems to be the genesis of the desire for the agglomeration before commenting more.

PR #3459, opened 2019-01-29
do not configure logging in the master process when using the reloader

### LINKED ISSUE #3455: pserve configures logging twice when using --reload
pserve invokes `loader.configure_logging()` before invoking `hupper.start_reloader` which configures logging in the monitor process **and** then later configures it again in the child/worker/app process.

We should, instead, invoke `basicConfig` if `--reload` is set and hupper is not active yet such that the parent process just logs to stderr instead of using the logging config intended for the app itself.

See https://groups.google.com/d/msgid/pylons-discuss/098b16e7-1955-4d21-9f4c-6a32d82fe103%40googlegroups.com

mmerickel:
fixes #3455

### DIFF
--- a/src/pyramid/scripts/pserve.py
+++ b/src/pyramid/scripts/pserve.py
@@ -182,7 +182,11 @@ class PServeCommand(object):
         app_name = self.args.app_name
 
         loader = self._get_config_loader(config_uri)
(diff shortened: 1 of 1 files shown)


mitymarcus:
@mmerickel wow, thank you so much!!!

PR #3464, opened 2019-02-16
Add ignore_files to pserve configuration section

ericatkin:


mmerickel:
An inconsistency here is that changes to `[pserve] watch_files` will work at runtime, however `[pserve] ignore_files` will not be reflected until pserve is completely restarted. I'm not sure what to do about that but maybe we could detect it and at least emit a warning.

ericatkin:
Also, I don't know how to do this, but shouldn't pyramid be dependent on the new hupper feature? How do we do this before hupper is tagged with a new version?

mmerickel:
I won't merge this until hupper is released and tests pass on that new version. For now maybe just add a checklist to the issue to indicate the missing step as we'll want to pin to that version in `setup.py`.

mmerickel:
I just released hupper 1.5.

ericatkin:
I'll have to look at the rest of it another time.

setup.py
```diff
@@ -26,7 +26,7 @@ def readfile(name):
 VERSION = '1.10.2'
 
 install_requires = [
-    'hupper',
+    'hupper >= 1.5',
```
mmerickel: please add a comment about why it’s pinned. Something like “ignore_files support” is enough.

mmerickel:
@ericatkin Is this where you're leaving things? Your earlier comment doesn't make it clear what you're planning to finish before you consider this PR complete.

As far as tracking the ignore_files, it basically requires computing a hash before start_reloader() and passing it to the child as a worker_kwarg. Then later around https://github.com/Pylons/pyramid/blob/fc7c597ce0e305acf3e0b4dd32ae70add4d4bef1/src/pyramid/scripts/pserve.py#L258 you can see if the current hash is the same as the one from the parent, if they aren't then just emit a message along the lines of "Detected a change to ignore_files but it will not take effect until pserve is restarted."

ericatkin:
Sorry for the ambiguity. I didn't have time last week to look in to the ignore_files warning and wasn't sure when I'd be get back to it. I decided I shouldn't leave it hanging though and so I'll push a commit shortly that implements what you've suggested (I think).

ericatkin:
Hmm, not sure what to do about this. Py2 only has assertNotRegexpMatches and Py34 doesn't have it? I use the assertRegex* because on failure the exception shows the two values vs something like `assertTrue(val1.startswith(val2))`

ericatkin:
BTW, is there some reason the lint section in tox.ini needs to be pinned to python3.6? It seems to work fine with 3.7. If it were just `python3` that would make running tests easier on people (like me) who don't have multiple versions of python 3 installed.

mmerickel:
> BTW, is there some reason the lint section in tox.ini needs to be pinned to python3.6? It seems to work fine with 3.7. If it were just `python3` that would make running tests easier on people (like me) who don't have multiple versions of python 3 installed.

I'm not against changing it but it's pinned to 3.6 because it requires 3.6 or greater and tox doesn't have a way to specify that. `python3` is just any version of python3 which will not work but I think technically you can do something like `tox -e lint-py37` to override it. Could be wrong tho.

mmerickel:
This looks great. I'm not sure about the word `extant` just because I had to look it up which makes it less readable but maybe that's just me. `original` or `pinned` or something might be easier to consume for people with no vocabulary. :-) If you wouldn't mind changing that I'd appreciate it. Either way I won't merge this until I or you or someone gets around to forward porting this to master as well. I might have time this weekend but I'm not sure.

ericatkin:
Ok, thanks for all your help! I made it `original_ignore_files`.

I don't know what "forward port to master" means in the context of a github PR. I'm pretty new to this. Would that be just a `git merge master` on my branch and fix any issues? I used the 1.10 tag because master was throwing exceptions when I tried to start my webapp, but I didn't spend much time looking at it. I won't have time for that until next week at the earliest.

Also, I feel like `test_original_ignore_files` is more complicated than it needs to be, but I didn't really understand _makeOne, DummyApp, DummyLoader, etc. You may want to take a peek. It works, but could perhaps be simplified a little by someone who understands those other pieces.

mmerickel:
It means do whatever needs to be done to get this feature working on master. That usually means cherry picking the commits over there as they are divergent branches. You cannot just merge across our release branches.

PR #3465, opened 2019-02-17
Security policy implementation

### LINKED ISSUE #3422: Auth post-mortem (Pyramid 2.0)

luhn:
- [x] `ISecurityPolicy`
- [x] `Configurator.set_security_policy`
- [x] bwcompat security policy
- [x] Request properties
- [x] Security view deriver
- [x] Built-in policies and helpers
- [x] Narrative docs
- [x] `DeprecationWarning` / `zope.deprecation`

Closes #3422


stevepiercy:
Please include a signed [`CONTRIBUTORS.txt`](https://github.com/Pylons/pyramid/blob/master/CONTRIBUTORS.txt). Otherwise LGTM so far.

luhn:
Security policies are working!  Please review.  Tests are passing locally, but it doesn't look like CI got triggered?

Test coverage is 100% except for [this line](https://github.com/Pylons/pyramid/blob/8a1f8d4acc77cd9fe57405384b5c20e9fa3fb078/src/pyramid/security.py#L43].  Looks like it's from the repoze.bfg days.  Still necessary in Pyramid 2.0, or can I delete it?

I think I should add an integration test for security policies and legacy policies.  Also update the other integration tests to use security policies.

I still need to implement built-in policies and helpers, which hasn't been discussed much:

* Existing authentication policies could be made into mixins providing `identify`, `remember`, and `forget`.
* Implement an ACL helper, as discussed in the post-mortem.
* I'd like to see a `BasicSessionSecurityPolicy` which remembers to the session and permits all or nothing based on the presence of an identity.  I think this would be a good plug-in-and-play default for a lot of smaller projects.

src/pyramid/config/security.py
```diff
@@ -21,6 +23,38 @@ class SecurityConfiguratorMixin(object):
...
+
+        intr = self.introspectable(
+            'security policy',
+            None,
+            self.object_description(policy),
+            'security policy',
+        )
+        intr['policy'] = policy
+        # authentication policy used by view config (phase 3)
+        self.action(ISecurityPolicy, register, introspectables=(intr,))
```
mmerickel: This needs to be done in phase 1 or phase2 similar to the old policies so that they are configured prior to add_view in phase 3.

src/pyramid/config/security.py
```diff
@@ -21,6 +23,38 @@ class SecurityConfiguratorMixin(object):
...
+
+    def _set_security_policy(self, policy):
+        policy = self.maybe_dotted(policy)
+        self.registry.registerUtility(policy, ISecurityPolicy)
```
mmerickel: I'm not sure what the point of this pattern is but the code should just be done in the register function. Not by adding a hidden method to the configurator. I assume it is a hook for tests but you aren't using it so it shouldn't be done.

src/pyramid/config/security.py
```diff
@@ -64,6 +99,15 @@ def _set_authentication_policy(self, policy):
...
         self.registry.registerUtility(policy, IAuthenticationPolicy)
 
+    def _set_legacy_policy(self):
+        if self.registry.queryUtility(ISecurityPolicy) is not None:
+            raise ConfigurationError(
+                'Cannot configure an authentication and authorization policy '
+                'with a configured security policy.'
+            )
+        policy = LegacySecurityPolicy()
+        self.registry.registerUtility(policy, ISecurityPolicy)
```
mmerickel: similarly just do this work in register

src/pyramid/security.py
```diff
@@ -280,19 +287,68 @@ class ACLAllowed(ACLPermitsResult, Allowed):
     """
 
 
+class SecurityAPIMixin(object):
+    @property
+    def identity(self):
```
mmerickel: What about `authorized_user`?

src/pyramid/config/__init__.py
```diff
@@ -139,13 +139,17 @@ class Configurator(
...
     If ``authentication_policy`` is passed, it should be an instance
     of an :term:`authentication policy` or a :term:`dotted Python
-    name` to the same.
+    name` to the same.  (Deprecated as of Pyramid 2.0 in favor of
+    ``security_policy``.)
 
     If ``authorization_policy`` is passed, it should be an instance of
     an :term:`authorization policy` or a :term:`dotted Python name` to
-    the same.
+    the same.  (Deprecated as of Pyramid 2.0 in favor of ``security_policy``.)
```
mmerickel: versionchanged directives should be added to the bottom of the docstring noting the deprecations and the additions.

mmerickel:
> I'd like to see a BasicSessionSecurityPolicy which remembers to the session and permits all or nothing based on the presence of an identity. I think this would be a good plug-in-and-play default for a lot of smaller projects.

The issue with shipping security policy classes is that they tie together a lot of decisions into one class. We should not add any without first having the building blocks and at least testing out the class in the existing tutorial (probably in the wiki2 tutorial like https://docs.pylonsproject.org/projects/pyramid/en/1.10-branch/tutorials/wiki2/authentication.html).

mmerickel:
> Existing authentication policies could be made into mixins providing identify, remember, and forget.

Mixins work for some things but we're actively trying to reduce the usage of things like the `CallbackAuthenticationPolicy` and prefer a composition approach instead. It's come up several times but helper objects would be a better way to start, mixins could be built using them, but the helper objects need to exist first.

mmerickel:
I added a bullet for deprecations. Anywhere that you're putting into the docs that something is deprecated should be followed by calls to `zope.deprecation` in the code to trigger `DeprecationWarning` warnings. There's a lot of code that needs to be deprecated, but it can come after things settle down.

src/pyramid/interfaces.py
```diff
@@ -482,8 +482,40 @@ def __call__(self, **kw):
...
+        """
+
+
 class IAuthenticationPolicy(Interface):
-    """ An object representing a Pyramid authentication policy. """
+    """ An object representing a Pyramid authentication policy.
+
+    .. deprecated:: 2.0
+
+        Use :class:`ISecurityPolicy`.
```
mmerickel: I think this needs to be `.ISecurityPolicy` with a leading dot to reference the object in the same file, but I could be wrong. Normally we'd use the dot or an absolute reference like `pyramid.interfaces.ISecurityPolicy`.

mmerickel: We'll need to link to some section in the docs from all these locations for some sort of narrative about the change / upgrading code. I'd expect something like `Use :class:``.ISecurityPolicy``. See :ref:``security_policy_upgrade`` for more information.`.

src/pyramid/config/__init__.py
stevepiercy: @mmerickel do you mean `versionchanged` or `deprecated` directives? `glossary.rst` uses `deprecated`.

mmerickel: I think if I were doing it myself I would've used versionchanged in cases where a feature (like a particular argument) is deprecated in a function but not we didn't deprecate the whole function. People may disagree with me though, that's just my opinion. Thoughts?

docs/glossary.rst
```diff
@@ -306,18 +306,36 @@ Glossary
...
+
    authentication policy
      An authentication policy in :app:`Pyramid` terms is a bit of
      code which has an API which determines the current
      :term:`principal` (or principals) associated with a request.
 
+     .. deprecated:: 2.0
+       Authentication policies have been deprecated in favor of a
+       :term:`security policy`.
+
```
mcdonc: I don't want to add stop energy here but I will :)  I am a big -1 on gluing together authentication and authorization into a single policy object.  Thoughts?

mmerickel: I think you need to expand on why you are concerned about this. A -1 with no reasoning isn’t very helpful. There is a lot of discussion in https://github.com/Pylons/pyramid/issues/3422#issuecomment-457861409 about why combining them provides a lot of benefits and discusses issues people are running into where 1) the only way to pass data between authn and authz policies right now is via principal strings as well as 2) subclassing of authn policies.

src/pyramid/config/__init__.py
stevepiercy: I'd follow the intent of the directives as explained in their documentation, [versionchanged](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) and [deprecated](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated). IOW, the feature has greater weight than the code (function, method, arguments) when deciding which directive is most appropriate.

In this case, one feature is deprecated in favor of another.

docs/glossary.rst
bertjwregeer: I've been thinking about this change a lot, and the positives and improvements that this change brings outweigh the negative effects of gluing the two together.

For many of my own projects these changes will provide greater flexibility to better match the way permissions work in my app that are not strictly ACL and/or based upon principals.

src/pyramid/security.py
luhn: My understanding was we were just doing `identity`.

mmerickel: Well it's an opaque object, yeah, but on the request I would prefer that it to be something more security-oriented than just `request.identity` because that feels fairly meaningless. For example, we have other systems like the debugtoolbar or tracing libraries that assign a `request.id` to a request. It's just an overloaded term and some form of namespacing would help. So `authorized_user`, `authorized_userid`, `authorized_identity`, `security_identity` or something would be better IMO. Definitely open to bike-shedding some ideas but I don't think `request.identity` is clear enough.

src/pyramid/config/security.py
luhn: Roger that.

Question:  Shouldn't policy be resolved (`policy = self.maybe_dotted(policy)`) in the method body, rather than in the config action?  That way the introspectable always gets the actual policy class.  Hesitant to change it since I assume I'm missing something.

mmerickel: Yeah in other actions that's done right away at the top of the method.

luhn: Thanks!  Made the requested changes to the configurator.

luhn:
So I'm going through the authentication policies to see what helpers I need to implement:

* CallbackAuthenticationPolicy — Obviously don't want this.
* RepozeWho1AuthenticationPolicy — Is this still relevant?  Happy to implement if it's something people want.
* RemoteUserAuthenticationPolicy — There's not much functionality here, not sure it merits a helper.
* AuthTktAuthenticationPolicy — There's already AuthTktHelper, and it looks to be what we want.
* SessionAuthenticationPolicy — Pretty minimal, does it merit a helper?
* BasicAuthAuthentictionPolicy — Already have extract_http_basic_credentials, which is basically all the functionaltiy.

So that leaves me with... nothing.  Thoughts?

src/pyramid/security.py
luhn: I don't find `identity` to be a problem.  If we are going for something more explicit, I'd vote for `authenticated_user` (too close to deprecated `authenticated_userid`?) or `authenticated_identity` or `secure_identity`.  I personally don't like `authorized_*` because it implies it's somehow involved with the authorization mechanisms, but it's strictly in the realm of identity/authentication.

luhn:
Started work on ACL helper, which is basically just moving ACLAuthorizationPolicy into standalone functions.  I think it makes sense to implement both `permits` and `principals_allowed_by_permission`.  Not sure best way to organize it:

* Put `acl_helper` and `principals_allowed_by_permission_helper` into `pyramid.security` (`pyramid.security.principals_allowed_by_permission` is already in use.)
* Put `ACLHelper` in `pyramid.security` with `permits` and `principals_allowed_by_permission` methods (Personally don't like, since the methods are stateless the class is superfluous.)
* Create new `pyramid.acl` module.  (Personal preference)

Thoughts?

mmerickel:
> RepozeWho1AuthenticationPolicy — Is this still relevant? Happy to implement if it's something people want.

I don't have a strong opinion on this one.

> RemoteUserAuthenticationPolicy — There's not much functionality here, not sure it merits a helper.

Agreed.

> AuthTktAuthenticationPolicy — There's already AuthTktHelper, and it looks to be what we want.

Yep.

> SessionAuthenticationPolicy — Pretty minimal, does it merit a helper?

I think it's worth adding a helper as it and auth_tkt are probably the most commonly used patterns in apps.

> BasicAuthAuthentictionPolicy — Already have extract_http_basic_credentials, which is basically all the functionaltiy.

The remember/forget parts are not done by `extract_http_basic_credentials`, so I'd argue a helper is useful here to contain that functionality.

> Started work on ACL helper, which is basically just moving ACLAuthorizationPolicy into standalone functions. I think it makes sense to implement both permits and principals_allowed_by_permission. Not sure best way to organize it:

Good question, my preference in the short term is an ACLHelper in pyramid.security just to help namespace the code. It's at least a good starting point.

luhn:
Alright, implemented the desired helpers and migrated `AuthTktCookieHelper` to `pyramid.security`.  That's pretty much it for the code, except for deprecation warnings and renaming `request.identity` to whatever we decide on.

luhn:
Any further comments?  Before I start updating the docs, is everybody on board with this implementation?

bertjwregeer:
I am very happy with the direction of this PR. I am looking forward to what this new implementation will allow implementers to do in the space of authentication/authorization.

src/pyramid/security.py
```diff
@@ -280,19 +296,68 @@ class ACLAllowed(ACLPermitsResult, Allowed):
...
+
         """
-        policy = _get_authentication_policy(self)
-        if policy is None:
+        authn = _get_authentication_policy(self)
+        security = _get_security_policy(self)
+        if authn is not None:
+            return authn.authenticated_userid(self)
+        elif security is not None:
+            return security.identify(self)
```
mmerickel: should this be `str(security.identify(self))` ?

src/pyramid/security.py
```diff
@@ -318,48 +390,740 @@ def effective_principals(self):
...
+        return []
+
+    def forget(self, request):
+        """ Remove the stored userid from the session."""
+        if self.userid_key in request.session:
+            del request.session[self.userid_key]
+        return []
+
+    def identify(self, request):
+        return request.session.get(self.userid_key)
```
mmerickel: can the `pyramid.authentication.SessionAuthenticationPolicy` be updated to use this to avoid the duplication?

mmerickel:
I'm not sold on moving all the authentication helpers into `pyramid.security`. Similar feelings with the `ACLHelper` but not as strong.

luhn:
Yeah, that's fair.  Should I put them in `pyramid.authentication` and `pyramid.authorization` as appropriate?

mmerickel:
> Yeah, that's fair. Should I put them in pyramid.authentication and pyramid.authorization as appropriate?

Yeah that'd be my preference.

src/pyramid/security.py
luhn: Good catch!  Yeah, I think you're right.

bertjwregeer:
I've been mulling over the name of `identity`. I know that @mmerickel has been too. We want it to include something like `secure_` or `authenticated_`.

We don't want to include something like "user" because we want it to be possible for it to be an opaque object that doesn't necessarily have to be a user.

We want to make it explicit that it is related to security/authentication/authorization, and `identity` is too all-encompassing.

luhn:
Personally I lean towards `authenticated_*`, I think "secure" is too vague.  `authenticated_identity`?

bertjwregeer:
Having taken some time to think about it, `authenticated_identity` would be okay, `auth_identity` may also work just to reduce the length of the name.

stevepiercy:
`auth_identity` is ambiguous: `authenticated_identity` or `authorized_identity`?

mmerickel:
authenticated_identity wins over authorized_identity due to the semantics of authorization

bertjwregeer:
> `auth_identity` is ambiguous: `authenticated_identity` or `authorized_identity`?

Since it's both part of the security policy and it deals with both authentication and authorization in one, I don't think it is a huge problem.

But I am okay with `authenticated_identity`.

luhn:
I've written the first draft of the docs.  I'm not very experienced writing narrative docs, so hopefully I did a decent job.  Feedback welcome!

I have a question about organization:  I've written the upgrade instructions in `upgrade-2.0.rst`.  There's already `narr/upgrading.rst`, should I put the content in there or does the 2.0 upgrade merit it's own page?

stevepiercy:
tl;dr: update `/CHANGES.rst` with specific instructions to upgrade from 1.10 to 2.0.

`narr/upgrading.rst` is a general procedure. Specific details of how to upgrade from a previous version to the current should be in the change log, which is `docs/changes.rst`, which pulls in `/CHANGES.rst` and `/HISTORY.rst`. When we do a release (see `/RELEASING.txt`), we use `/CHANGES.rst` as the source from which we update other files, including `/HISTORY.rst` and `docs/whatsnew-1.10.rst`.

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+upcoming versions.
+
+The new security policy should implement
+:interface:`pyramid.interfaces.ISecurityPolicy` and can be set via the
+``security_policy`` argument of :class:`pyramid.config.Configurator` or
+:meth:`pyramid.config.Configurator.set_security_policy`.
+
+The new security policy merges ``unauthenticated_userid`` and
+``authenticated_userid`` into an :term:`identity` object.  This object can be
+of any shape, such as a simple ID string or an ORM object, but should   The
```
stevepiercy: Something not right here. Copy-pasta error?

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+match the ACL pattern.  If you were previously using
+:class:`pyramid.authorization.ACLAuthorizationPolicy`, you can achieve the same
+results by writing your own ``permits`` method using
+:class:`pyraid.authorization.ACLHelper`.  For more details on implementing an
+ACL, see :ref:`assigning_acls`.
+
+Pyramid does not provide any built-in security policies.  Similiar
+functionality of the authentication and authorization policies is now provided
+by helpers, which can be utilized to easily implement your own security policy.
+The functionality of the legacy authencation policies roughly correspond to the
```
stevepiercy: s/authencation/authentication

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+flexibility in authorization implementations, especially those that do not
+match the ACL pattern.  If you were previously using
+:class:`pyramid.authorization.ACLAuthorizationPolicy`, you can achieve the same
+results by writing your own ``permits`` method using
+:class:`pyraid.authorization.ACLHelper`.  For more details on implementing an
+ACL, see :ref:`assigning_acls`.
+
+Pyramid does not provide any built-in security policies.  Similiar
+functionality of the authentication and authorization policies is now provided
+by helpers, which can be utilized to easily implement your own security policy.
```
stevepiercy: remove "easily". Auth is hard.

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+The functionality of the legacy authencation policies roughly correspond to the
+following helpers
```
stevepiercy: add period

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+  :class:`pyramid.authentication.SessionAuthenticationHelper`
+* :class:`pyramid.authentication.AuthTktAuthenticationPolicy`:
+  :class:`pyramid.authentication.AuthTktCookieHelper`
+* :class:`pyramid.authentication.BasicAuthAuthenticationPolicy`:
+  Use :func:`pyramid.authentication.extract_http_basic_credentials` to retrieve
+  credentials.
+* :class:`pyramid.authentication.RemoteUserAuthenticationPolicy`:
+  ``REMOTE_USER`` can be accessed with ``request.environ.get('REMOTE_USER')``.
+* :class:`pyramid.authentication.RepozeWho1AuthenticationPolicy`:
+  No equivalent.
```
stevepiercy: I'd prefer a table to a list, with headings of "Legacy" and "Pyramid 2.x". It's easier to visually scan tabular information.

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+
+For further documentation on implementing security policies, see
+:ref:`writing_security_policy`.
+
```
stevepiercy: Add an index entry such as the following or whatever makes sense:

```
.. _behavior-of-the-legacy-system:
```

docs/upgrading-2.0.rst
```diff
@@ -0,0 +1,74 @@
...
+Behavior of the Legacy System
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Legacy authentication and authorization policies will continue to function as
+normal, as well as all related :class:`pyramid.request.Request` properties.
+The new :prop:`pyramid.request.Request.authenticated_identity` property will
```
stevepiercy: `:prop:` is a new one to me! It looks like `:prop:` requires Sphinx 2.1+. If that assumption is true, then we would need to bump the version minimum version in `setup.py` accordingly.

It would also be good to verify that it builds on RTD, too, before merging this PR. To do so, you can setup your fork on RTD and build the specific branch.

stevepiercy:
Also to save some effort, I'd suggest getting `/CHANGES.rst` into a good final state, then copy-paste portions of it into `docs/narr/security.rst` and modify it accordingly.

docs/upgrading-2.0.rst
luhn: Whoops, didn't realize it was 2.1+.  As of a couple months ago, RTD was running Sphinx 1.8, so I think we'll have to get rid of it.

What's the recommended way to reference properties in older versions of Sphinx?

mmerickel:
Like @stevepiercy said we usually put general guidance into the changelog directly and more comprehensive info in whatsnew-2.0. Look at historical ones like 1.5 for examples of major changes when upgrading.

docs/upgrading-2.0.rst
stevepiercy: If memory serves me correctly, we can force RTD to use a more current version of Sphinx via the use of:

* modify `setup.py`
* invoke via `rtd.txt`'s `-e .[docs]`, which in turn is invoked by...
* configuring RTD under _Advanced Settings_ to use `rtd.txt` as a requirements file and install into a virtual environment (see screenshot).
Google Chrome007

I can grant you access to the RTD configuration as an admin if you need to compare setup if you provide me with your RTD username. I might have omitted a detail, but I think that covers it all.

luhn:
Okay, I'll add entries to `CHANGES.rst`.

@stevepiercy, I'm not sure what you mean about copying `CHANGES.rst` to `narr/security.rst`, I've already updated `narr/security.rst`.  Are my narrative updates missing something?

My `upgrade-2.0.rst` was intended to address @mmerickel's [comment](https://github.com/Pylons/pyramid/pull/3465#discussion_r264017483) about having an upgrade guide to reference in deprecation notices.  Does `whatsnew-2.0.rst` serve that purpose, or do we have a different section for that?

mmerickel:
You can leave the upgrade guide anywhere for now but I think its final home will be in whatsnew-2.0.

stevepiercy:
> I'm not sure what you mean about copying CHANGES.rst to narr/security.rst, I've already updated narr/security.rst. Are my narrative updates missing something?

Oops, sorry. I forgot you had done that already.

The content of `upgrade-2.0.rst` should eventually go into `whatsnew-2.0.rst`.

luhn:
If it's alright with you guys, I'll get a whatsnew-2.0 started now so we don't have this temporary upgrade-2.0 floating around.

mmerickel:
That’s fine by me. Go for it.

docs/upgrading-2.0.rst
luhn: Looks like `:attr:` does what I'm looking for and that works in 1.8.  I'm not even sure where I found `:prop:`, I don't see it in the Sphinx documentation.

IMO best to stick with the RTD Sphinx version if we can help it.

luhn:
I've added deprecation warnings.

I skipped `set_authorization_policy` because it necessitates `set_authentication_policy`, which will throw a warning.

I skipped `authenticated_userid` and `unauthenticated_userid` because they are very common, so I figured adding a warning would result in a useless flood of deprecation warnings.

Actually, now that I think about it, perhaps we shouldn't deprecate `authenticated_userid` at all.  Since it maps to `str(request.authenticated_identity)`, which is documented as something an identity object should implement, it's still somewhat meaningful and useful.

luhn:
Related to my question above, I realize `remember()` and `forget()` are still dealing with `userid` as well.  I think that makes sense, since most applications will just remember a user ID rather than a full identity object.  So do we want to keep around the concept of "userid" around and separate from identity?

Other than that, I've reviewed the code and everything looks good AFAICT.  Please review yourselves and let me know if anything needs to change.

bertjwregeer:
I need this PR to land and a new release of Pyramid, like yesterday ;-) This would solve some problems I am currently working on.

I'll review in a second, but changing "userid" to "identity" may make sense, but I'd like @mmerickel's opinion on that.

mmerickel:
I think the api for remember/forget should remain a userid and not an opaque object. The identity object is opaque and is meant for exposing data to permits but there's no standard way to load one outside of the policy in order to pass it in. I think `authenticated_identity` is basically replacing `effective_principals` but there's still use cases for userid.

luhn:
Since we're still using userid, should I undo the deprecation of `authenticated_userid`?

mmerickel:
@luhn that’s my feeling right now. Definitely open to alternative opinions though.

luhn:
Done!

mmerickel:
I confirmed over the weekend that some $work apps continue to function unchanged using this branch which is great. I'm planning to get into the rest of the details more this week and try to get this merged.

docs/narr/security.rst
```diff
@@ -37,89 +32,144 @@ how it works at a high level:
...
+                return Denied('User is not signed in.')
+
+        def remember(request, userid, **kw):
+            return self.helper.remember(request, userid, **kw)
+
+        def forget(request):
+            return self.helper.forget(request)
+
+Helpers are intended to be used with application-specific code, so perhaps your
+authentication also queries to database to ensure the identity is valid.
```
mmerickel:
```suggestion
authentication also queries the database to ensure the identity is valid.
```

docs/glossary.rst
```diff
@@ -298,13 +298,20 @@ Glossary
...
-     :term:`principals <principal>`. In the authentication policies which
-     :app:`Pyramid` provides, the default behavior returns the user's userid as
-     a principal, but this is not strictly necessary in custom policies that
-     define their principals differently.
+     A *userid* is the string representation of an :term:`identity`.  Just like
+     the identity, it should identify the user associated with the current
+     request.  Oftentimes this is the ID of the user object in a database.
+
+   identity
+     An identity is an object identify the user associated with the
```
mmerickel:
```suggestion
     An identity is an object identifying the user associated with the
```

mmerickel:
@luhn can you do me a tiny favor and rename this PR to remove the WIP just to make sure you're happy with things before I merge?

luhn:
👍

luhn:
Excited to see this merged!  Thanks to everyone for working with me on this.

Cheers! 🍻

docs/glossary.rst
```diff
@@ -298,13 +298,20 @@ Glossary
...
-     a principal, but this is not strictly necessary in custom policies that
-     define their principals differently.
+     A *userid* is the string representation of an :term:`identity`.  Just like
+     the identity, it should identify the user associated with the current
+     request.  Oftentimes this is the ID of the user object in a database.
+
+   identity
+     An identity is an object identifying the user associated with the
+     current request.  The identity can be any object, but should implement a
+     ``__str__`` method that outputs a corresponding :term:`userid`.
```
merwok: Hi!  Sorry to comment after this is merged, but I was following the issue and not aware of this PR.

I don’t understand the reason for using `__str__` as the method to convert Identity to user ID.
- `__str__` is meant to control what happens with `print`, `%s` formatting, etc.
- IMO it is good practice to have `__str__` fast, without side effects, without possibility of failure (so no parsing, casting, etc)
- `repr` uses `__str__` if the class doesn’t define `__repr__`
- REPL, logging, others use repr or str
- depending on the system, user ID may not be a string
- calling `str(identity)` instead of `identity.some_explicit_method()` seems a bit obscure (not discoverable)

If people agree with the disadvantages here, could this API decision be reconsidered?

docs/whatsnew-2.0.rst
```diff
@@ -0,0 +1,103 @@
...
+The new security policy should implement
+:class:`pyramid.interfaces.ISecurityPolicy` and can be set via the
+``security_policy`` argument of :class:`pyramid.config.Configurator` or
+:meth:`pyramid.config.Configurator.set_security_policy`.
+
+The new security policy adds the concept of an :term:`identity`, which is an
+object representing the user associated with the current request.  The identity
+can be accessed via :attr:`pyramid.request.Request.authenticated_identity`.
+The object can be of any shape, such as a simple ID string or an ORM object,
+but should implement a ``__str__`` method that outputs a string identifying the
```
merwok: Minor: using `outputs` here (and other parts of this PR) could be misleading (does it mean `print`?), compared to `returns`

docs/glossary.rst
mmerickel: This is a discussion I'm happy to have. I'm not in love with `__str__` myself. I'd ask that since this is merged we open a new issue to talk about it though, instead of here.

ztane: There are 2 good options that come to my mind: either return `storable_id, IdentityClass` tuple or have another function in the policy that derives the serializable id from the identity class...

luhn: There's been some discussion on this, I dug up what I could find:

https://github.com/Pylons/pyramid/issues/3422#issuecomment-443840114
https://github.com/Pylons/pyramid/pull/3465#issuecomment-511140572

luhn: Opened an issue over at #3520

src/pyramid/security.py
```diff
@@ -280,86 +293,155 @@ class ACLAllowed(ACLPermitsResult, Allowed):
...
+            When using the new security system, this property outputs the
+            string representation of the :term:`identity`.
+
+        """
+        authn = _get_authentication_policy(self)
+        security = _get_security_policy(self)
+        if authn is not None:
+            return authn.authenticated_userid(self)
+        elif security is not None:
+            return str(security.identify(self))
```
merwok: @luhn can I ask why this method checks old-style authn policy before security policy?
The other methods in the mixin only work with security policy, without falling back on (or checking first!) an authentication policy.

Is there a goal to support a config with security policy + authentication policy?

luhn: Usually we just let LegacySecurityPolicy handle the bw compat, but in this case we need to circumvent the `str(security.identify(self))`, otherwise we'd be changing the behavior of the legacy method.

Now that `str(identity)` is out and authenticated_userid is back, we should be able to just reference the security policy and let LegacySecurityPolicy work its magic.

merwok: Great! :sparkles:

src/pyramid/authentication.py
```diff
@@ -823,10 +682,150 @@ def encode_ip_timestamp(ip, timestamp):
...
+    ``domain``
+
+       Default: ``None``. If provided the auth_tkt cookie will only be
+       set for this domain. This option is not compatible with ``wild_domain``
+       and ``parent_domain``.
+       Optional.
+
+    ``hashalg``
+
+       Default: ``sha512`` (the literal string).
```
merwok: Noticed today that the actual default is `md5`: https://github.com/Pylons/pyramid/pull/3465/files#diff-00f789cd007443672848b9eb23973a86R861

Ping @mmerickel this could be changed in your doc PR too?  Fix the default to match the doc (as this is a new class for Pyramid 2), then the docs example can remove the explicit `hashalg` parameter.

For the backward-compatible authn policy, its constructor could be kept to `md5` (if that is indeed the previous behaviour).

mmerickel: Unfortunately this class is not new.

mmerickel: I can't believe we forgot to fix it when we changed it on the policy. I'm changing it now and we'll just release a deprecation in 1.10.5.

merwok: Alright then I should send a doc PR for 1.x / stable branch?
(not sure how many versions of Pyramid are maintained and how backports work)

mmerickel: I did it yesterday in #3558

invisibleroads:
These are significant changes!

stevepiercy:
@invisibleroads you can continue to use existing auth policy stuff, too. See the [What's New in Pyramid 2.0 document](https://docs.pylonsproject.org/projects/pyramid/en/2.0-branch/whatsnew-2.0.html#upgrading-authentication-authorization) for details in the blue box.

invisibleroads:
Pyramid 2.0 is big news. I am very happy to see that Pyramid is still very alive after 10+ years thanks to lots of work from all of you.

PR #3480, opened 2019-03-30
Fix docs build on Sphinx 2.0.0

luhn:
Sphinx 2.0.0 was released, which triggered some warnings when building
the docs and caused the build to fail.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ tests_require = [
 
 
 docs_extras = [
-    'Sphinx == 1.8.5',  # Current version for Read the Docs
+    'Sphinx >= 1.8.1',  # Unicode characters in tree diagrams
     'docutils',
     'pylons-sphinx-themes >= 1.0.8',  # Ethical Ads
     'pylons_sphinx_latesturl',
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ tests_require = [
 
 
 docs_extras = [
-    'Sphinx >= 1.8.1',  # Unicode characters in tree diagrams
+    'Sphinx == 1.8.5',  # Current version for Read the Docs
     'docutils',
     'pylons-sphinx-themes >= 1.0.8',  # Ethical Ads
     'pylons_sphinx_latesturl',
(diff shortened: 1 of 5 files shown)


luhn:
Hmmm, have some errors in the epub build I can't quite puzzle out.

luhn:
There we are!

stevepiercy:
Read The Docs is currently on Sphinx 1.8.5. We might need to pin Sphinx, or account for two versions, if there is a version incompatibility.

luhn:
Yeah, works on both 1.8.5 and 2.0.0.  I've pinned it to 1.8.5 so Travis tests against same sphinx as RTD, but this PR will at least future-proof the docs.

Where are you getting that RTD is Sphinx 1.8.5?  I can't find a reference for that.

setup.py
```diff
@@ -44,7 +44,7 @@ def readfile(name):
 
 
 docs_extras = [
-    'Sphinx >= 1.8.1',  # Unicode characters in tree diagrams
+    'Sphinx == 1.8.5',  # Current version for Read the Docs
```
stevepiercy: Is this pin necessary? RTD will install Sphinx 1.8.5 without it. We want to upgrade to 2.x whenever RTD gets around to it. They rarely make such an announcement. It's better to have the flexibility to build with either 1.8.5 on RTD or 2.0 locally (and whenever RTD upgrades).

luhn: Happy to revert it if you want.

stevepiercy: Yes, please. Then we can forget about it.

Here's the requirement:
https://github.com/rtfd/readthedocs.org/blob/master/requirements/pip.txt#L13

luhn: Reverted.

Ignorant question:  Could we put `sphinx >= 2.0.0` in the doc extras to force RTD to use 2.0?

stevepiercy: _Good_ question! Nope, I tried it a while ago. They pin Sphinx to avoid breaking customer docs.

stevepiercy:
Oops, approved too soon. There should also be a changelog entry under https://github.com/Pylons/pyramid/blame/master/CHANGES.rst#L91

luhn:
Will do.

Should I backport to `1.x-master` as well?

stevepiercy:
> Should I backport to `1.10-branch` as well? `1.9-branch`?

Yes, please.

Issue #3471, opened 2019-03-20
Fix Python 3.8 test failure


stevepiercy:
Fix the test failure on Python 3.8 for Pyramid 2.0:
https://travis-ci.org/Pylons/pyramid/jobs/508850032#L223

luhn:
I took a look at this and it went a lot deeper than I was expecting.

The failing test is checking for `ConfigurationConflictError` with conflicting `@view_config`.  It checks the offending config actions by looking at `ActionInfo.src`, the value of which changes in 3.8 causing the tests to fail.  Following the trail, that value comes from `inspect.getframeinfo().code_context[0]`.  https://github.com/Pylons/venusian/blob/master/venusian/advice.py#L53

It looks like in Python 3.8 the code context for decorators has changed from the decorator itself to the decorated function, e.g. `@view_config()` becomes `def view()`.  You can see that in action here:  https://gist.github.com/luhn/d1dac5eff4d6fd0018ae12eac48e94e0  If you dig even deeper, this is happening because `sys._getframe().f_lineno` changes.

I've considered whether this is a bug in Python, but imo the code context given by 3.8 is correct.

Anyways, so as far as fixing this test, our options:

- If we consider this a bug in Python, we could file a bug report and hope it gets fixed before release.
- We could check the Python version and change the assertions based on that.  This seems bad practice.
- We could assert against something else in the `ActionInfo`, but unfortunately there doesn't seem to be anything useful there.
- We can remove the offending assertions.  This is what I would pick, because `sys._getframe()` is an implementation detail and I don't think it should be considered a stable API.  I don't think we lose much by removing them.

bertjwregeer:
Went and looked at when the change was made, and it looks like it is related to this bug:

https://bugs.python.org/issue34876
https://github.com/python/cpython/pull/9731

The fix makes sense, and looking at the output from trace it makes sense why this change was made.

What I'd really like to see us do though is return the line number of the decorator where the issue lies so that it is easy to find the error in your source code.

Right now with stacking decorators I don't think we do the right thing in venusian to get that information, and we will just get `def view()`.

luhn:
I don't think Venusian would be able to get that, because it's at the mercy of `sys._getframe()` within `venusian.attach`, but we should be able to get a useful frame within `pyramid.view.view_config`.  I'll see what I can do.

bertjwregeer:
[...]

PR #3483, opened 2019-04-02
Improve debugging info from view_config decorator

luhn:
Grab a frame when instanciating view_config, rather than pulling info from
venusian's AttachInfo.  See discussion at
https://github.com/Pylons/pyramid/issues/3471

This fixes the build for Python 3.8.

### DIFF
--- a/src/pyramid/view.py
+++ b/src/pyramid/view.py
@@ -220,7 +220,8 @@ class view_config(object):
         self._get_info()
 
     def _get_info(self):
-        frameinfo = inspect.stack()[2]
+        depth = self.__dict__.get('_depth', 0)
+        frameinfo = inspect.stack()[depth + 2]
         sourceline = frameinfo[4][0].strip()
         self._info = frameinfo[1], frameinfo[2], frameinfo[3], sourceline
 
--- a/src/pyramid/view.py
+++ b/src/pyramid/view.py
@@ -220,10 +220,9 @@ class view_config(object):
         self._get_info()
 
     def _get_info(self):
-        frame = sys._getframe(2)
-        frameinfo = inspect.getframeinfo(frame)
-        sourceline = frameinfo[3][0].strip()
-        self._info = frameinfo[0], frameinfo[1], frameinfo[2], sourceline
+        frameinfo = inspect.stack()[2]
+        sourceline = frameinfo[4][0].strip()
+        self._info = frameinfo[1], frameinfo[2], frameinfo[3], sourceline
 
     def __call__(self, wrapped):
         settings = self.__dict__.copy()
--- a/src/pyramid/view.py
+++ b/src/pyramid/view.py
@@ -1,5 +1,6 @@
 import itertools
 import sys
+import inspect
 
 import venusian
 
@@ -216,6 +217,13 @@ class view_config(object):
             if settings.get('context') is None:
                 settings['context'] = settings['for_']
         self.__dict__.update(settings)
+        self._get_info()
+
+    def _get_info(self):
+        frame = sys._getframe(2)
+        frameinfo = inspect.getframeinfo(frame)
+        sourceline = frameinfo[3][0].strip()
+        self._info = frameinfo[0], frameinfo[1], frameinfo[2], sourceline
 
     def __call__(self, wrapped):
         settings = self.__dict__.copy()
@@ -237,14 +245,13 @@ class view_config(object):
             if settings.get('attr') is None:
                 settings['attr'] = wrapped.__name__
 
-        settings['_info'] = info.codeinfo  # fbo "action_method"
         return wrapped
 
 
 bfg_view = view_config  # bw compat (forever)
 
 
-class view_defaults(view_config):
+def view_defaults(**settings):
     """ A class :term:`decorator` which, when applied to a class, will
     provide defaults for all view configurations that use the class.  This
     decorator accepts all the arguments accepted by
@@ -253,9 +260,11 @@ class view_defaults(view_config):
     See :ref:`view_defaults` for more information.
     """
 
-    def __call__(self, wrapped):
-        wrapped.__view_defaults__ = self.__dict__.copy()
+    def wrap(wrapped):
+        wrapped.__view_defaults__ = settings
         return wrapped
+
+    return wrap
 
 
 class AppendSlashNotFoundViewFactory(object):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,9 @@ Features
   provided by WebOb. This allows the attribute to now be settable.
   See https://github.com/Pylons/pyramid/pull/3447
 
+- Improve debugging info from ``pyramid.view.view_config`` decorator.
+  See https://github.com/Pylons/pyramid/pull/3483
+
 Deprecations
 ------------
 
--- a/docs/api/view.rst
+++ b/docs/api/view.rst
@@ -14,8 +14,7 @@
   .. autoclass:: view_config
      :members:
 
-  .. autoclass:: view_defaults
-     :members:
+  .. autofunction:: view_defaults
 
   .. autoclass:: notfound_view_config
      :members:
--- a/tests/test_view.py
+++ b/tests/test_view.py
@@ -507,7 +507,14 @@ class TestViewConfigDecorator(unittest.TestCase):
 
     def test_create_defaults(self):
         decorator = self._makeOne()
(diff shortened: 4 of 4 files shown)


luhn:
One possible breaking change:  I changed `view_defaults` from a class based on `view_config` to a simple function.  This should be functionally identical except that `for_` will no longer replace `context`.  This behavior has no tests and is not documented.  Nor is it documented for `view_config`, so I'm not sure its purpose.

luhn:
Now that the tests are passing, do we remove 3.8 from allowed failures, or do we wait until the stable release?

mmerickel:
> Now that the tests are passing, do we remove 3.8 from allowed failures, or do we wait until the stable release?

I haven't reviewed anything but I can at least answer that question. It stays in allowed failures until we officially support it which will be after it has at least a beta release, probably a final release. It's not in allowed failures just because the tests were failing.

bertjwregeer:
@luhn 

There is a failure in the documentation because it is no longer a class:

https://travis-ci.org/Pylons/pyramid/jobs/514447775#L650

Please update the docs for this change too.

src/pyramid/view.py
```diff
@@ -216,6 +217,13 @@ def __init__(self, **settings):
             if settings.get('context') is None:
                 settings['context'] = settings['for_']
         self.__dict__.update(settings)
+        self._get_info()
+
+    def _get_info(self):
+        frame = sys._getframe(2)
```
bertjwregeer: I'm wondering if instead of using `sys._getframe(2)` directly if we shouldn't instead use

```
inspect.stack()[2]
```

instead to grab the frame, then there is no reason to use an underscore method on `sys`.

src/pyramid/view.py
```diff
@@ -237,14 +245,13 @@ def callback(context, name, ob):
             if settings.get('attr') is None:
                 settings['attr'] = wrapped.__name__
 
-        settings['_info'] = info.codeinfo  # fbo "action_method"
```
bertjwregeer: The only benefit I see from also using this `info.codeinfo` is that it can tell us where the view is actually located which we are wrapping.

For example if you import a view function from another module, and then "decorate" it and store it in the local module, we could point the user at where their view code is located, as well as where the `view_config` decorator is being applied.

May not be worth it though, so if there are no objections to anything else in this changeset, ignore this comment.

luhn:
I've fix the docs build.  You mention updating the docs... what needs to be changed?

bertjwregeer:
Nope, you got what I meant!

src/pyramid/view.py
```diff
@@ -216,6 +217,12 @@ def __init__(self, **settings):
             if settings.get('context') is None:
                 settings['context'] = settings['for_']
         self.__dict__.update(settings)
+        self._get_info()
+
+    def _get_info(self):
+        frameinfo = inspect.stack()[2]
+        sourceline = frameinfo[4][0].strip()
+        self._info = frameinfo[1], frameinfo[2], frameinfo[3], sourceline
```
mmerickel: should this use the `_depth` argument at all? The `[2]` in the stack may not be what you want in all situations but I'm not sure? I suspect this will always point at the location of the view_config decorator, but if view_config is wrapped in something else, you'd want to point there?

bertjwregeer: Ah,. that's a good point. It should use depth + 1.

bertjwregeer:
One thing I will mention, is that we likely should update the documentation on `_depth` to specify that if you are wrapping `view_config` you NEED to set the `_depth` to 1 or more (depending on how far into your call stack you go).

That `_depth` variable is used by venusian to figure out how far up the stack it most go to find the `class` `f_locals` to be able to attach the information to the `class`, if that is off, then `venusian`'s `scan` won't find it because the information was attached to the wrong location.

mmerickel:
What is completely unclear to me, and what venusian needs to document, is how to use that variable. It's unclear in which scenarios it matters. For example if I use `view_config` in a decorator, or partial, or class with `__call__` etc. It is **very** hard to decipher right now. I'd almost say that venusian should error if the decorator is used in certain circumstances but it at least should document how to get it right.

bertjwregeer:
venusian can't know if it is being used in the wrong way. What venusian needs to know is how far up the stack it needs to walk to find the `class` declaration it is inside. If you get the depth wrong, it attaches information to the wrong location (but only if the decorator is being used on a class method)

bertjwregeer:
Venusian attaching to functions inside a class is basically black magic :/

mmerickel:
So that's why I don't usually see it matter is because I mostly decorate module-level functions?

mmerickel:
I thought it was about the depth of the actual decorator with respect to the function it's decorating. So the more function calls between the func and the call to view_config, the more depth was needed. The way you described it seems more about attaching the info to the class versus the method?

mmerickel:
Sorry to derail this thread but I bet that everyone has these questions... or at least I have for years and never bothered to get answers!

bertjwregeer:
> I thought it was about the depth of the actual decorator with respect to the function it's decorating.

It is. It's both. It only matters if you are decorating functions defined inside a class though, because if it doesn't detect it, it will add an attribute on the class itself, if you are just decorating module level functions and never functions/methods inside a class, then you can completely ignore the depth argument and it doesn't matter.

----

If you have:

```
import venusian

def myattach(func):
    venusian.attach(func, ...)

class Klass:
    @myattach
    def somefunc():
        pass
```

Then Venusian will walk up the stack enough to find `Klass` to be able to attach information to it so that during a `venusian.scan()` it will correctly pick up that `somefunc()` was decorated.

If however like `view_config` you nest it into `__call__` you need to explicitly add to `_depth` because `venusian.attach` now needs to move further up the stack to find `Klass`.

If someone else comes along and wraps `view_config` they need to explicitly increase `_depth`, especially if your new replacement for `view_config` is used to decorate functions inside a class, or else `venusian` may mis-categorize and attach the venusian information to the wrong location.

```
from .attacher import attach

class Klass:
    @attach(kwarg=True)
    def somefunc():
        pass
```

`attacher.py`:

```
import venusian

def someattacher(func):
    venusian.attach(func, ...)

def attach(kwarg=False):
    def _attacher(func):
        someattacher(func)
    return _attacher
```

Will have `venusian` look up the stack, and find `_attacher`, which is not a class and not a package/module, and thus it will attach the information to `somefunc()` inside of `Klass`.

It does this by looking at some heuristics of whether or not the `f_locals` and `f_globals` for that particular stack frame it is looking at contains certain attributes/variables. If you correctly use `depth` though, `venusian` will use it's heuristics to say, yes, this stack frame is a class, thus we need to attach the information to `Klass` so that it is at module level so that a `.scan()` can pick it up.

So someone that is using `view_config` in their code needs to increase `_depth` to match if they want users to be able to use their new decorator on functions inside of a class.

bertjwregeer:
Specifically, if the `depth` is wrong, then the `scope` here will not equal `class`:

https://github.com/Pylons/venusian/blob/master/venusian/__init__.py#L313-L328

mmerickel:
So it's about how many stackframes to skip before looking for a class or module scoped frame?

bertjwregeer:
Yes. Because Python at decorator time doesn't provide a way to know if the function is defined in a class or not :-( That is only available at call time.

mmerickel:
Is there anything wrong with my conclusion? If it's that simple to explain then it should be in the docs in several places.

bertjwregeer:
No, there is nothing wrong with your conclusion. Yes, it should be in the docs. You were able to boil down a long lengthy comment to just a few words because you are smart like that.

mmerickel:
Haha it's just because I'm the one asking the question I'm willing to be wrong and show my ignorance. You're in a position of trying to answer me correctly and thus you're more naturally afraid to be wrong. :-)

mmerickel:
I tagged this for backporting to 1.10-branch, will wait to merge this until I get around to it or someone else provides that PR.

luhn:
Yeah, I can do it.

mmerickel:
Good stuff, thank you @luhn!

mmerickel:
Hey @luhn I didn't realize until after, but could you please add your name to CONTRIBUTORS.txt? Sorry but you'll have to do it in a separate PR now :-(

luhn:
Already in CONTRIBUTORS.

https://github.com/Pylons/pyramid/blob/master/CONTRIBUTORS.txt#L344

mmerickel:
Oops I was looking at 1.10-branch on accident. That's good enough for me that it's on master.

PR #3486, opened 2019-04-12
Refactor base_version to major_version.

stevepiercy:
- base_version includes the minor version number, and we want the major version which is only the first two bits of the version number.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,11 +54,13 @@ docs_extras = [
 
 testing_extras = tests_require + ['coverage', 'nose']
 
-base_version = parse_version(VERSION).base_version
+major_version = ".".join(
+    str(v) for v in parse_version(VERSION)._version.release[:2]
+)
 
 # black is refusing to make anything under 80 chars so just splitting it up
 docs_fmt = 'https://docs.pylonsproject.org/projects/pyramid/en/{}-branch/'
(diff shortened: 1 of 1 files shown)


stevepiercy:
I found this when testing out the project_urls on PyPI. I also backported it to 1.10 in #3487, where we started using them.

setup.py
```diff
@@ -54,11 +54,13 @@ def readfile(name):
 
 testing_extras = tests_require + ['coverage', 'nose']
 
-base_version = parse_version(VERSION).base_version
+major_version = ".".join(
+    str(v) for v in parse_version(VERSION)._version.release[:2]
+)
```
mmerickel: Instead of using this private api I'd rather just use `'.'.join(VERSION.split('.')[:2])`.

setup.py
```diff
@@ -54,11 +54,11 @@ def readfile(name):
 
 testing_extras = tests_require + ['coverage', 'nose']
 
-base_version = parse_version(VERSION).base_version
+major_version = ".".join(VERSION.split(".")[:2])
```
mmerickel: just call it root_version, major already has meaning and base does too according to pkg_resources

mmerickel: either that or branch_version.

mmerickel:
@stevepiercy fix the lint

Issue #3489, opened 2019-04-15
Pyramid 1.10.3 is much slower to initiate


Deimos:
Just updated to Pyramid 1.10.3 and it's now considerably slower to initiate. The main reason I notice this is because I'm using [WebTest](https://docs.pylonsproject.org/projects/webtest/en/latest/index.html) and my first test that uses it is now much slower than it used to be.

On Pyramid 1.10.2, the setup time for my first WebTest test averages about 0.7 seconds. After updating to Pyramid 1.10.3 (and changing nothing else), the setup time for the same test increases to about 3.8 seconds.

Similarly, starting up `pshell` is noticeably slower. I haven't tried to time it properly, but I can now easily count past 5 while waiting for it to start up, but on Pyramid 1.10.2 I can't even get to 3.

I can try to help investigate where the additional time is being spent if needed, but I figured you might already know which of the 1.10.3 changes could be the culprit.

mmerickel:
You should check that other libraries are at the same versions as you test and then finally if that's all constant then check if the changes made to view_config to improve stackframe captures [1]. The diff in this release is ultra tiny, affecting only pserve (but you said you're using bootstrap, not pserve) and view_config.

https://github.com/Pylons/pyramid/compare/1.10.2...1.10.3

Deimos:
Sorry, I shouldn't have used the word "bootstrap". I'll try to do some profiling and see if I can figure out where the extra time is being spent.

mmerickel:
Gotcha, well I suspect it's because of the stackframe code in the diff if it is slower but it'd help if you could narrow down that it's at least not related to any third party dependency issues in your own project first.

luhn:
Hmm, I figured there would be a performance regression with the new view_config, but nothing so dramatic.  venusian already inspects the frame, so I figured worst-case would be 2x slowdown.

Just spitballing, it could be that inspect.stack() calls getframeinfo for every frame in the stack, whereas venusian only calls getframeinfo for a single frame.

Chad, would you mind testing against https://github.com/luhn/pyramid/tree/performance-regression? <https://github.com/luhn/pyramid/tree/performance-regression?>  (`pip install --force https://github.com/luhn/pyramid/archive/performance-regression.zip` <https://github.com/luhn/pyramid/archive/performance-regression.zip%60>)

— Theron

Deimos:
[...]

PR #3490, opened 2019-04-15
Fix performance regression in view_code changes (1.10)

luhn:
As discusssed in #3489

This uses `sys._getframe`, which was previously avoided because it's a cpython implementation detail (however, it is [supported in pypy](https://www.pypy.org/performance.html#micro-tuning-tips)).  It is used by venusian as well, so we already depend on the `sys._getframe`.

I'll port it to 2.0 if everybody is on board.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,11 @@
 .. _changes_1.10.3:
 
+Unreleased
+==========
+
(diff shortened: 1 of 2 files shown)


bertjwregeer:
Please port to 2.0, thank you!

Issue #3441, opened 2018-12-05
Wrong URL in "Creating Your First Pyramid Application"


padomu:
## Get Support
To get help or technical support, see [Get Support](https://pylonsproject.org/community-support.html).

## Documentation Suggestion

Please [search the issue tracker](https://github.com/Pylons/pyramid/issues) for similar issues before submitting a new issue.

**Describe the issue**
At https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/firstapp.html#firstapp-chapter it says that one should call "localhost:6543/hello/world" to run the simple application but the application just dispatches "/", so a user gets a 404.

Apparently the commit who introduced the bug: https://github.com/Pylons/pyramid/commit/4b84f6e97b321309544513b0697e8b378a57bafc#diff-656e2e498075a842eb2fecb4c3ce6d72

The older application: https://github.com/Pylons/pyramid/blob/b6937e32be59a214bac5a8097bf5b200eb7b5ac8/docs/narr/helloworld.py

**Describe the improvement**
Just change "If your application is running on your local system, using http://localhost:6543/hello/world in a browser will show this result." to "If your application is running on your local system, using http://localhost:6543/ in a browser will show this result."

stevepiercy:
I'd accept a PR with this change, as well as in the earlier sentence:

```
When port 6543 is visited by a browser on the URL ``/hello/world``
```

s/b

```
When port 6543 is visited by a browser on the URL ``/``
```

PR #3492, opened 2019-04-28
Fix for issue #3441

Myrheimb:
This fixes the issue described in #3441.

Changed the URL for the 'hello' route in "Creating Your First Pyramid Application" so that it doesn't give a 404.


stevepiercy:
Thanks for the fix. I missed this reference in the narrative docs when updating the code to which it refers. I'll also forward port it on `master`.

Issue #3468, opened 2019-02-27
Response callbacks not always invoked after exception view


lrowe:
In [Using Hooks](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#using-response-callbacks) it is stated that "Response callbacks are, however, invoked when a exception view is rendered successfully." While this is true when the exception view is invoked by the excview_tween, it may also be invoked by the execution policy in which case the response callbacks are not invoked.

When ab exception is raised under excview_tween the following are executed in order:
```
request.invoke_exception_view(exc_info)
request._process_response_callbacks(response)
notify(NewResponse(request, response))
request._process_finished_callbacks()
```

When an exception is raised over excview_tween it is:
```
request._process_finished_callbacks()
request.invoke_exception_view(reraise=True)
```

In my case I had an exception view registered that called `request.session.changed()` to refresh the session cookie. BaseCookieSessionFactory.changed sets the cookie in a response callback so this no longer works.

I'm not sure if this a documentation or a behavioral issue.

bertjwregeer:
At least in `pyramid_retry`, the only execution policy we currently have, it is a last ditch effort to try and render the exception. That would fall under the:

> No response callback is called if an unhandled exception happens in application code, or if the response object returned by a view callable is invalid.

Part of the docs, since the exception is not a valid response object that is returned. I am not sure it is a good idea for `pyramid_retry` to call the response call backs since at that point you are far outside of the usual request cycle.

I would question why your exception is falling through to the execution policy of `pyramid_retry`, instead of being handled by the excview tween.

bertjwregeer:
The same statements stand for the default execution policy: https://github.com/Pylons/pyramid/blob/master/src/pyramid/router.py#L274

mmerickel:
I think in hindsight it was probably a mistake to have the execution policy call `invoke_exception_view`. I did know it was a thing Pyramid didn't do before, but it seemed safe enough to try and get a response out. This type of inconsistency though feels not great. In general exception views need to be really careful about what they expect [1], since they can be called at weird spots in the lifecycle. It's unclear what a good solution is though.

[1] For example, pyramid_tm adds a `tm_active` predicate which exception views can use to ensure they only match if the transaction manager is still active. This type of predicate isn't supported for other things like `request.session` though.

mmerickel:
[...]

PR #3496, opened 2019-06-07
do not reinvoke exception views if an exception hits the execution policy

mmerickel:
fixes https://github.com/Pylons/pyramid/issues/3468


bertjwregeer:
:+1:

PR #3512, opened 2019-09-20
Add allow_no_origin option to CSRF

### LINKED ISSUE #3508: Allow disabling the Origin check in CSRF protection
## Feature Request

**Is your feature request related to an issue? Please describe.**
A decent number of users use privacy-oriented settings or extensions that can prevent their browser from sending referrer info. Lynx-based browsers seem to do this by default as well. Because of this, Pyramid's CSRF protection prevents those users from being able to use any views with CSRF protection, since the Origin check for CSRF always fails.

**Describe the solution you'd like**
I'd like an easy way to disable the Origin part of the check (maybe an argument on [config.Configurator.set_default_csrf_options](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.set_default_csrf_options)?). The Origin check is recommended but not truly necessary, and between the token check and `SameSite` cookies I'm comfortable disabling it.

**Describe alternatives you've considered**
I *think* I could probably effectively disable the check by finding somewhere to catch the `BadCSRFOrigin` exception and ignore it, but it seems a little tricky and a way to just disable the check cleanly would be nicer.

I haven't dug through the relevant Pyramid code much, so if there's already an easy way to ignore this check that I missed, please let me know.

luhn:
Fixes #3508


luhn:
Things to discuss:

- `allow_no_origin` — Name okay?
- `allow_no_origin` argument to `set_default_csrf_options` and `check_csrf_origin` — Technically not backwards compatible.

mmerickel:
I think the name is fine. I thought about `allow_unknown_origin` or `allow_missing_origin` as well. This all seems good to me so far.

luhn:
Great!  Wrote up some docs.

docs/narr/security.rst
```diff
@@ -944,7 +944,9 @@ that it matches one of the trusted origins. By default the only trusted origin
 is the current host, however additional origins may be configured by setting
 ``pyramid.csrf_trusted_origins`` to a list of domain names (and ports if they
 are non-standard). If a host in the list of domains starts with a ``.`` then
-that will allow all subdomains as well as the domain without the ``.``.
+that will allow all subdomains as well as the domain without the ``.``.  If no
+``Referer`` or ``Origin`` header is present in an HTTPS request, the CSRF check
+will fail unless the ``allow_no_origin`` is set.
```
Deimos: Minor nitpick: should probably remove the "the" on this line.

src/pyramid/config/security.py
```diff
@@ -221,6 +222,9 @@ def set_default_csrf_options(
         never be automatically checked for CSRF tokens.
         Default: ``('GET', 'HEAD', 'OPTIONS', TRACE')``.
 
+        ``allow_no_origin`` is a boolean.  If false, a request lacking both an
+        ``Origin`` and ``Referer`` header will fail the CSRF check.'
```
Deimos: Tiny thing again: extra single quote `'` on the end of the line.

luhn:
Thanks for catching those!  Changes made.

docs/whatsnew-2.0.rst
```diff
@@ -0,0 +1,16 @@
+What's New in Pyramid 2.0
```
bertjwregeer: This document generally gets built when we are ready to go to press... not sure we want to add it in this PR.

src/pyramid/config/security.py
```diff
@@ -236,9 +240,17 @@ def set_default_csrf_options(
         .. versionchanged:: 1.8
            Added the ``callback`` option.
 
+        .. versionchanged:: 2.0
+           Added the ``allow_no_origin`` option.
+
         """
         options = DefaultCSRFOptions(
-            require_csrf, token, header, safe_methods, callback
+            require_csrf=require_csrf,
```
bertjwregeer: Not a huge fan of using keywords here for passing variables that are not keyword arguments in the target function... but I'll defer to @mmerickel for this.

mmerickel: it's not a public api so it's a non-issue

On a side note re: Python 3 - I'm all for liberal use of keyword-only via `foo(*, arg1, arg2)` arguments in Pyramid 2.0!

mmerickel:
Honestly I feel like the default for this should be ``True`` considering an ``Origin`` header isn't set by default when on the same domain. In that case aren't we currently just falling back to the `Referer` header? Or am I wrong here?

mmerickel:
To clarify it seems to me like the current logic is:

same domain request -> no origin header sent, use referer
cross domain request -> origin header is almost always sent, fallback to referer

I think (without testing) that even if you turn off the privacy settings named by @Deimos, the Origin header will still be sent on cross domain requests, and this PR is really focused on fixing same domain requests. The common case here, to me, should be permissive as compliant user agents will probably still sent an Origin header on cross-domain requests even in private browsing situations because it hides the origin's path and only shows the domain. The spec also leaves room for "null" as the value of the Origin header, which we aren't supporting atm but I'd claim should be handled permissively if set, which it is not right now.

My proposed changes:

1) change default to True

2) handle `Origin: null` as permissive always, non-configurable

Deimos:
Personally, I'd support the check being disabled by default, at least for same-domain requests. I think the really key thing to recognize is that the token-based method is already considered strong, and it is individually good protection. The Origin/Referer check is just a second (weak) layer that isn't truly reliable and can cause issues like this.

[In the OWASP cheatsheet that I linked in the issue](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md), it says:

> We recommend token based CSRF defense (either stateful/stateless) as a primary defense to mitigate CSRF in your applications. Only for highly sensitive operations, we also recommend a user interaction based protection (either re-authentication/one-time token, detailed in section 6.5) along with token based mitigation.
>
> As a defense-in-depth measure, consider implementing one mitigation from Defense in Depth Mitigations section (you can choose the mitigation that fits your ecosystem considering the issues mentioned under them). These defense-in-depth mitigation techniques are not recommended to be used by themselves (without token based mitigation) for mitigating CSRF in your applications.

The Origin check is one of the "Defense in Depth Mitigations". It's not an essential part, and wouldn't be considered proper protection on its own. The recommendation in the cheatsheet is:

> Usually, a minor percentage of traffic does fall under above categories [for why the headers wouldn't be present] (1-2%) and no enterprise would want to lose even this minor % of traffic. One of the popular technique used across Internet to make this technique more usable is to accept the request if the Origin/referrer matches your configured list of domains "OR" a null value (Examples here. null value is to cover the edge cases mentioned above where these headers are not sent). Please note that, attackers can exploit this but people prefer to use this technique as a defense in depth measure because of minor effort involved in deploying it

mmerickel:
I opened https://github.com/Pylons/pyramid/pull/3518, please let me know what you folks think.

PR #3513, opened 2019-09-20
Add Github actions for CI

bertjwregeer:
Depends on https://github.com/Pylons/pyramid/pull/3578


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,88 @@
...
+
+on: 
+    # Only on pushes to master or one of the release branches we build on push
+    push:
+        branches:
+            - master
+            - "[0-9].[0-9]+-branch"
+        tags:
+    # Build pull requests
+    pull_request:
```
bertjwregeer, line 11: You'll like this @mmerickel

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Set the default behaviour, in case poeple don't have core.autocrlf set
```
bertjwregeer, line 1: This file is required because core.autocrfl is set to "auto" on Windows by default

.gitattributes
```diff
@@ -0,0 +1,5 @@
...
+* text=auto
+
+# Always lf, this file size is checked in a test
+tests/fixtures/static/encoded.html text eol=lf
```
bertjwregeer, line 5: And this file was getting checked out, and it's line endings were changed, however a test was counting bytes inside the file.

So here we explicitly set the `eol` to linefeed so that git doesn't change the file size accidentally.

bertjwregeer:
With the discussion on https://github.com/Pylons/pyramid/pull/3578 I've now disabled coverage on everything but `CPython` on `ubuntu-latest`.

PR #3514, opened 2019-09-20
define an IPredicateInfo instead of passing the full configurator to predicates

### LINKED ISSUE #2535: define a new IPredicateInfo to mirror IDeriverInfo
One annoyance with predicates right now is that their signature is `val, config` which doesn't make much sense. It's not particularly good practice to pass around the config. The only reason it's done right now is because the containment predicate needs to call `config.maybe_dotted`. That's it. It'd be much preferred to be something like `info, val` which could more closely mirror the new `IViewDeriverInfo`. We could then define an `IPredicateInfo` which contained the registry and package information without passing the raw `config`.

This would be a backward compatibility issue, unfortunately, so I'd save it for Pyramid 2.0.

mmerickel:
fixes #2535

- [x] fix predicate docs
- [x] add changelog

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,6 +46,10 @@ Features
   ``Referer`` header for privacy reasons.
   See https://github.com/Pylons/pyramid/pull/3512
 
+- Added ``pyramid.interfaces.IPredicateInfo`` which defines the object passed
+  to predicate factories as their second argument.
+  See https://github.com/Pylons/pyramid/pull/3514
+
 Deprecations
 ------------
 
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -46,6 +46,10 @@ Features
   ``Referer`` header for privacy reasons.
   See https://github.com/Pylons/pyramid/pull/3512
 
+- Added ``pyramid.interfaces.IPredicateInfo`` which defines the object passed
+  to predicate factories as their second argument.
+  See https://github.com/Pylons/pyramid/pull/3514
+
```
bertjwregeer: Fancy multi-line comment... to be fancy.

PR #3519, opened 2019-10-02
simplify caller_package logic

mmerickel:


bertjwregeer:
This is going to sound annoying, but could you add more information to the git message as to why this change is the same as the previous code? I am not sure I understand all the mechanics of it.

mmerickel:
I'll see what I can do to increase confidence. In general Pyramid is using some pretty old conventions here and I was looking at what I could modernize in that file via new importlib features in 3.x. The dotted name resolver is also ripe for some improvement but it honestly just isn't that hard to understand right now so I left it alone.

mmerickel:
I'll get back to this later... closing for now.

PR #3524, opened 2019-11-04
deprecate PickleSerializer

### LINKED ISSUE #3511: Feature Request- deprecate/remove PickleSerializer
This has basically already been decided to happen, but the timing is not clear.  I could not find the scheduled removal listed anywhere.

`[pyramid.session.PickleSerializer](https://github.com/Pylons/pyramid/blob/587fe72fae0efda3a860d37a1ea2449a41dab622/src/pyramid/session.py#L46-L70)` is the legacy session implementation and [not used by any actual code in Pyramid](https://github.com/Pylons/pyramid/search?q=PickleSerializer&unscoped_q=PickleSerializer).

It only exists in a test, docs and the changelog.  The code is simple enough that it could be consolidated into the narrative docs that invoke it.

If this is staying for a while to enable 3rd party libraries that require it, I suggest invoking a `warnings.warn`  on `PickleSerializer.__init__()` to have people upgrade.

If this does not need to stay, can it be removed sooner than later?   I keep nearly using it on a compatibility layer, and then I realize this will likely be gone soon.

jvanasco:
fixes #3511 

I updated the narrative docs to reflect this change:

* the fallback example recipe now reimplements the PickleSerializer logic, instead of importing the now-deprecated class.
* the fallback example uses the `appstruct`/`bstruct` terminology and docstrings of the now-deprecated PickleSerializer

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ unreleased
 Features
 --------
 
+- Deprecated ``pyramid.session.PickleSerializer``.
+
 - Changed the default ``serializer`` on
   ``pyramid.session.SignedCookieSessionFactory`` to use
   ``pyramid.session.JSONSerializer`` instead of
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Features
 --------
 
 - Deprecated ``pyramid.session.PickleSerializer``.
+  See https://github.com/pylons/pyramid/issues/2709
+  and https://github.com/pylons/pyramid/pull/3353
+  and https://github.com/pylons/pyramid/pull/3413
 
 - Changed the default ``serializer`` on
   ``pyramid.session.SignedCookieSessionFactory`` to use
--- a/docs/narr/sessions.rst
+++ b/docs/narr/sessions.rst
@@ -85,6 +85,12 @@ This is a stricter contract than the previous requirement that all objects be pi
 This is a backward-incompatible change.
 Previously, if a client-side session implementation was compromised, it left the application vulnerable to remote code execution attacks using specially-crafted sessions that execute code when deserialized.
 
+Please reference the following tickets if detailed information on these changes is needed:
+
+* `2.0 feature request: Require that sessions are JSON serializable #2709 <https://github.com/pylons/pyramid/issues/2709>`_.
+* `deprecate pickleable sessions, recommend json #3353 <https://github.com/pylons/pyramid/pull/3353>`_.
+* `change to use JSONSerializer for SignedCookieSessionFactory #3413 <https://github.com/pylons/pyramid/pull/3413>`_.
+
 For users with compatibility concerns, it's possible to craft a serializer that can handle both formats until you are satisfied that clients have had time to reasonably upgrade.
 Remember that sessions should be short-lived and thus the number of clients affected should be small (no longer than an auth token, at a maximum). An example serializer:
 
--- a/docs/narr/sessions.rst
+++ b/docs/narr/sessions.rst
@@ -91,26 +91,32 @@ Remember that sessions should be short-lived and thus the number of clients affe
 .. code-block:: python
     :linenos:
 
+    import pickle
     from pyramid.session import JSONSerializer
-    from pyramid.session import PickleSerializer
     from pyramid.session import SignedCookieSessionFactory
+
 
     class JSONSerializerWithPickleFallback(object):
         def __init__(self):
             self.json = JSONSerializer()
-            self.pickle = PickleSerializer()
 
-        def dumps(self, value):
+        def dumps(self, appstruct):
+            """Accept a Python object and return bytes."""
             # maybe catch serialization errors here and keep using pickle
             # while finding spots in your app that are not storing
             # JSON-serializable objects, falling back to pickle
-            return self.json.dumps(value)
+            return self.json.dumps(appstruct)
 
-        def loads(self, value):
+        def loads(self, bstruct):
+            """Accept bytes and return a Python object."""
             try:
-                return self.json.loads(value)
+                return self.json.loads(bstruct)
             except ValueError:
-                return self.pickle.loads(value)
+                try:
+                    return pickle.loads(bstruct)
+                # at least ValueError, AttributeError, ImportError but more to be safe
+                except Exception:
+                    raise ValueError
 
     # somewhere in your configuration code
     serializer = JSONSerializerWithPickleFallback()
--- a/docs/narr/sessions.rst
+++ b/docs/narr/sessions.rst
@@ -101,10 +101,14 @@ Remember that sessions should be short-lived and thus the number of clients affe
             self.json = JSONSerializer()
 
         def dumps(self, appstruct):
-            """Accept a Python object and return bytes."""
-            # maybe catch serialization errors here and keep using pickle
-            # while finding spots in your app that are not storing
-            # JSON-serializable objects, falling back to pickle
+            """
+            Accept a Python object and return bytes.
+
+            During a migration, you may want to catch serialization errors here,
+            and keep using pickle while finding spots in your app that are not
+            storing JSON-serializable objects. You may also want to integrate
+            a fall-back to picke serialization here as well.
+            """
             return self.json.dumps(appstruct)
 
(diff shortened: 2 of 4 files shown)


docs/narr/sessions.rst
```diff
@@ -91,26 +91,32 @@ Remember that sessions should be short-lived and thus the number of clients affe
...
         def __init__(self):
             self.json = JSONSerializer()
-            self.pickle = PickleSerializer()
 
-        def dumps(self, value):
+        def dumps(self, appstruct):
+            """Accept a Python object and return bytes."""
             # maybe catch serialization errors here and keep using pickle
             # while finding spots in your app that are not storing
             # JSON-serializable objects, falling back to pickle
```
stevepiercy: These 3 commented lines above should be updated, too, correct?

docs/narr/sessions.rst
```diff
@@ -91,26 +91,32 @@ Remember that sessions should be short-lived and thus the number of clients affe
...
+        def loads(self, bstruct):
+            """Accept bytes and return a Python object."""
             try:
-                return self.json.loads(value)
+                return self.json.loads(bstruct)
             except ValueError:
-                return self.pickle.loads(value)
+                try:
+                    return pickle.loads(bstruct)
+                # at least ValueError, AttributeError, ImportError but more to be safe
```
stevepiercy: This comment seems misplaced. Maybe indent to align with the `return` or push below within the next code block?

src/pyramid/session.py
```diff
@@ -44,10 +44,24 @@ def changed(session, *arg, **kw):
...
+    use :class:`pyramid.session.JSONSerializer`, and ``PickleSerializer`
+    has been been removed from active Pyramid code.
+
+    Pyramid will require JSON-serializable objects in :app:`Pyramid` 2.0.
+
+    Please see :ref:`pickle_session_deprecation`.
+
+    Also, please see: #2709, #3353, #3413
+
+    A serializer that uses the pickle protocol to dump Python data to bytes.
```
stevepiercy: This block needs to be indented 4 spaces to fall under the `.. warning::` reStructuredText directive.

Also I do not think that we have enabled the Sphinx extension that will convert issue or PR numbers to hyperlinks, so please use the actual links to those issues/PRs.

docs/narr/sessions.rst
jvanasco: thanks for catching these so fast. I was having some issues with rst documents.  I'll address asap

jvanasco: @stevepiercy so, I looked at this and I agree it is weird, but I copied this style exactly from the deprecated code:  https://github.com/Pylons/pyramid/blob/master/src/pyramid/session.py#L60-L66

I would be happy to nest the comment under the `except` block, as I prefer that style myself. this is up to you and Mike though, and I am happy to issue a pull on it if you'd like to switch it.

stevepiercy: Well if both of us had to say "huh?", then it is probably not the way to do it. :wink:

src/pyramid/session.py
jvanasco: ok; I'm going to migrate that info into the session docs too.

docs/narr/sessions.rst
```diff
@@ -85,32 +85,49 @@ This is a stricter contract than the previous requirement that all objects be pi
...
-            # JSON-serializable objects, falling back to pickle
-            return self.json.dumps(value)
+        def dumps(self, appstruct):
+            """
+            Accept a Python object and return bytes.
+
+            During a migration, you may want to catch serialization errors here,
+            and keep using pickle while finding spots in your app that are not
+            storing JSON-serializable objects. You may also want to integrate
+            a fall-back to picke serialization here as well.
```
stevepiercy:
```suggestion
            a fall-back to pickle serialization here as well.
```

docs/narr/sessions.rst
```diff
@@ -85,32 +85,49 @@ This is a stricter contract than the previous requirement that all objects be pi
...
             try:
-                return self.json.loads(value)
+                return self.json.loads(bstruct)
             except ValueError:
-                return self.pickle.loads(value)
+                try:
+                    return pickle.loads(bstruct)
+                except Exception:
+					# this block should catch at least:
+					# ValueError, AttributeError, ImportError; but more to be safe
```
stevepiercy: Why so many indents? Tabs?

src/pyramid/session.py
stevepiercy: I think it is also good practice to put links to the issue/PRs in the change notes instead of docstrings. I would look at change notes first when I see deprecation warnings, and I can refer to issue tracker for further information.

CHANGES.rst
```diff
@@ -4,6 +4,8 @@ unreleased
 Features
 --------
 
+- Deprecated ``pyramid.session.PickleSerializer``.
```
stevepiercy: Please add the following, but with fully qualified URLs instead of just #s.

```
See: #2709, #3353, #3413
```

src/pyramid/session.py
jvanasco: Noted for next time!

CHANGES.rst
jvanasco: thanks for catching that. updated. i also signed the contributors file, as i've only signed it in the sideprojects.

stevepiercy:
The [build fails](https://travis-ci.org/Pylons/pyramid/jobs/607797098#L362) because you have not run [Black](https://github.com/psf/black) on the code base. We now require Black. Please install and run Black on the files so that the build will pass.

jvanasco:
@stevepiercy I'm 100% willing to do this, but we should discuss first.  I'm not sure this is the right place to handle this ticket; it might be best done elsewhere and then a merge.

My branch passed on black `19.3b0` on my system. However... on Oct 28th, `19.10b0` was released, which integrated a handful of changes.  All of the travis fails are due to these changes, and affect code that is unrelated to my PR.

Perhaps we should do a new PR which specifies black 19.3b0 as the minimum, runs the new version on the codebase to fix these 7 unrelated files, then apply this PR?

stevepiercy:
Ah, yes, you are correct. Sorry about that. I created a new issue to track the matter. And at the same time you created a new PR. Jinx!

jvanasco:
ok, i'll do that.

docs/narr/sessions.rst
```diff
@@ -85,32 +85,49 @@ This is a stricter contract than the previous requirement that all objects be pi
 This is a backward-incompatible change.
 Previously, if a client-side session implementation was compromised, it left the application vulnerable to remote code execution attacks using specially-crafted sessions that execute code when deserialized.
 
+Please reference the following tickets if detailed information on these changes is needed:
```
bertjwregeer: Is this something we've done in the past, having our narrative documentation link to issue tickets?

tests/test_session.py
```diff
@@ -564,6 +564,10 @@ def test_it(self):
 
 
 class TestPickleSerializer(unittest.TestCase):
+    """
```
bertjwregeer: I don't think it is necessary to add a docstring to tests... but meh.

jvanasco: I only put it there so it would pop up faster on the list of things to delete when the actual removal takes place.

docs/narr/sessions.rst
jvanasco: I don't think I've seen that in this project per-se, but a handful of other projects do this when there is some sort of extended discussion detailing a security related issue, but it's just not worth repeating all of that stuff within the documentation itself.

stevepiercy: Bert's correct. I just scanned narrative docs, and there are no direct links to a specific issue or PR. As I consider it, it makes sense to not add a maintenance burden, especially if things change again in the future.

Let's just fall back to the entries in `CHANGES.rst` and call it soup.

tests/test_session.py
stevepiercy: When it is removed, wouldn't the test fail?

bertjwregeer: @jvanasco Yeah, I figured it was to make it simpler to grep across the codebase for stuff that is removed. It's fine :-)

PR #3525, opened 2019-11-05
black 19.10b0 formatting

jvanasco:
black 19.10b0 was released on 2019.10.28 and introduced several changes that affect Pyramid

this is just running the new `black` on master.  I could not find a reference in the Pyramid repo to specifying a version of black.

#3524

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -344,3 +344,5 @@ Contributors
 - Theron Luhn, 2019/03/30
 
 - Mandar Vaze, 2019/07/20
+
+- Jonathan Vanasco, 2019/11/05
--- a/src/pyramid/threadlocal.py
+++ b/src/pyramid/threadlocal.py
@@ -55,7 +55,7 @@ def get_current_request():
 
 
 def get_current_registry(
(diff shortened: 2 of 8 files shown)


mmerickel:
I do love me some trailing commas but I hate how black doesn't handle comments well

merwok:
Agree about comments (and not letting me decide when I want sequences with one item per line!)

I wonder if the formatting is better when the comment is inside the parens but before the code.

stevepiercy:
> I wonder if the formatting is better when the comment is inside the parens but before the code.

Before:

```python
class DummyRequest:

    application_url = (
        'http://example.com:5432'
    )  # comment after parens
    application_url_2 = (
        # comment inside parens
        'http://example.com:5432'
    )
    # comment above parens
    application_url_3 = (
        'http://example.com:5432'
    )
```

After:

```python
class DummyRequest:

    application_url = "http://example.com:5432"  # comment after parens
    application_url_2 = (
        # comment inside parens
        "http://example.com:5432"
    )
    # comment above parens
    application_url_3 = "http://example.com:5432"
```

Issue #3432, opened 2018-11-26
support precompressed static assets


mmerickel:
`config.add_static_view` and `pyramid.static.static_view` would benefit from supporting precompressed assets in some way. For example, if my build pipeline dumped out pre-gzipped `.js.gz` assets, I'd like them to get served directly. For example, through content-negotiation get a list of acceptable encodings.

```
offers = request.accept_encoding.acceptable_offers(['gzip', 'deflate', 'identity'])
```

These encodings could then be used to look in the filesystem for files with some pre-defined names. For example `foo.js` is requested, and `gzip in offers`, look for `foo.js.gz` first, and iterate through the acceptable offers until something is found. I am not asking for this feature to support finding `foo.js.gz` when only `identity` is supported and gunzipping it in the response. That could be some other future extension to this work though.

This is all very rough but would be really nice to have support for it. Right now it's necessary to write some tween that translates the requested path and is a bit tedious.

```
>>> mimetypes.guess_type('foo.js.gz')
('application/javascript', 'gzip')
```

PR #3537, opened 2019-12-05
negotiate the best static asset using supported encodings

mmerickel:
fixes https://github.com/Pylons/pyramid/issues/3432

Basically this adds support for a `content_encodings` argument in `add_static_view` and `static_view`. If supplied, then `static_view` will use the `accept-encoding` HTTP header to try and serve the appropriately-encoded file (preferring smaller files). This will also set the ``Vary`` header appropriately for files with multiple encodings.

The list of encodings comes from `mimetypes.encodings_map` and so it depends on the system. Modifying the map can be done to add support for other encodings like brotli.

```python
mimetypes.encodings_map['.br'] = 'br'
```

The locations of the files on disk are also cached by default unless `pyramid.reload_assets` is true.


src/pyramid/static.py
```diff
@@ -94,46 +143,128 @@ def __call__(self, context, request):
         if path is None:
             raise HTTPNotFound('Out of bounds: %s' % request.url)
 
+        # normalize asset spec or fs path into resource_path
         if self.package_name:  # package resource
             resource_path = '%s/%s' % (self.docroot.rstrip('/'), path)
             if resource_isdir(self.package_name, resource_path):
                 if not request.path_url.endswith('/'):
-                    self.add_slash_redirect(request)
+                    raise self.add_slash_redirect(request)
```
bertjwregeer: This changes the behaviour from previous versions of Pyramid, will this break anything?

bertjwregeer:
I say yes... but Travis says no.

src/pyramid/static.py
mmerickel: It's not different, I just moved the raise outside the method instead of inside to be more explicit.

mmerickel:
Thanks, yeah I'm writing up some narrative docs right now and changelog. I also obviously forgot to blackify.

src/pyramid/static.py
bertjwregeer: Oh, I see now. I missed it. :+1:

src/pyramid/static.py
```diff
@@ -94,46 +143,126 @@ def __call__(self, context, request):
...
+        # only cache the results if reload is disabled
+        if not self.reload:
+            self.filemap[resource_name] = result
+        return result
+
+    def find_best_match(self, request, files):
+        """ Return ``(path | None, encoding)``."""
+        # if the client did not specify encodings then assume only the
+        # identity is acceptable
+        if not request.accept_encoding:
```
mmerickel: short circuit `if len(files) < 2` here and return the identity path as well - no sense in doing the rest of the accept-header negotiation if we only have one possible result

PR #3542, opened 2019-12-09
Fixed typo. Remove extra $.

behzod:
Making a new pull request agains the master branch as requested in https://github.com/Pylons/pyramid/pull/3541

### DIFF
--- a/docs/quick_tutorial/json.rst
+++ b/docs/quick_tutorial/json.rst
@@ -63,7 +63,7 @@ Steps
 
    .. code-block:: bash
 
-    $ $VENV/bin/pserve development.ini --reload
+    $VENV/bin/pserve development.ini --reload
 
 #. Open http://localhost:6543/howdy.json in your browser and you will see the
    resulting JSON response.


docs/quick_tutorial/json.rst
```diff
@@ -63,7 +63,7 @@ Steps
 
    .. code-block:: bash
 
-    $ $VENV/bin/pserve development.ini --reload
+    $VENV/bin/pserve development.ini --reload
```
mmerickel: the initial `$` is supposed to be the bash command prompt indicator... I'm not sure how consistent we are with this through our examples. I see the bullet point above this doesn't use it. I'm pointing out the reason so we can check what our standard practice is in most of the docs before deciding which way is correct because I don't have an answer.

stevepiercy: We leave prompts off because (1) they differ across shells, (2) copy-paste errors and easy to select a line with a triple-click, (3) the command itself does not include the prompt, and (4) syntax highlighting interprets the prompt as a part of the command which is confusing.

Issue #3520, opened 2019-10-02
Change `__str__` requirement for identity


luhn:
Opening this based on @merwok's https://github.com/Pylons/pyramid/pull/3465#discussion_r330175629

> Hi! Sorry to comment after this is merged, but I was following the issue and not aware of this PR.
> 
> I don’t understand the reason for using `__str__` as the method to convert Identity to user ID.
> 
>     * `__str__` is meant to control what happens with `print`, `%s` formatting, etc.
> 
>     * IMO it is good practice to have `__str__` fast, without side effects, without possibility of failure (so no parsing, casting, etc)
> 
>     * `repr` uses `__str__` if the class doesn’t define `__repr__`
> 
>     * REPL, logging, others use repr or str, so seeing something that’s a user ID instead of custom str/repr output (like `<SomeIdentityClass id=blah source=value>`) in debug logs, etc. could make inspection harder
> 
>     * depending on the system, user ID may not be a string
> 
>     * calling `str(identity)` instead of `identity.some_explicit_method()` seems a bit obscure (not discoverable)
> 
> 
> If people agree with the disadvantages here, could this API decision be reconsidered?

Some previous discussions:

https://github.com/Pylons/pyramid/issues/3422#issuecomment-443840114
https://github.com/Pylons/pyramid/pull/3465#issuecomment-511140572

luhn:
The more I think about it, the more I think @merwok's complaint has merit.

My current favorite idea is adding a `userid()` method to the security policy.  This has a couple advantages:

- The identity can now be entirely opaque.  That was the original intention, but we strayed from that with the requirement of `__str__`.
- In the common case that userid is stored in the session and the identity object stored in a database, the userid can be fetched without needing to query the database.
- Clearer boundaries for the `SessionAuthenticationHelper`, etc.  Helpers implement remember, forget, userid.  The user implements identify and permits.

luhn:
I realize my second and third points don't address the idea of `authenticated_userid` being the userid after it's been checked against the database.

Personally I've never found the unauthenticated_userid/authenticated_userid dichotomy to be useful.  However, it does seem to be important.

Perhaps un-deprecate `unauthenticated_userid`, which naively returns the userid provided by the session or whatever, then has `authenticated_userid` return the userid *only if the identity is not null*.

luhn:
[...]

PR #3544, opened 2019-12-14
Rework security policy

merwok:
Addresses #3520 for #3422 
Follow-up to PR #3465 

- [ ] fix all tests
- [ ] fix all docs
- [ ] confirm some API points (see https://github.com/Pylons/pyramid/issues/3520#issuecomment-565742075)

### DIFF
--- a/src/pyramid/config/testing.py
+++ b/src/pyramid/config/testing.py
@@ -13,6 +13,7 @@ class TestingConfiguratorMixin(object):
     # testing API
     def testing_securitypolicy(
         self,
+        userid=None,
         identity=None,
         permissive=True,
         remember_result=None,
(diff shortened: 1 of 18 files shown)


tests/test_viewderivers.py
```diff
@@ -2086,7 +2086,10 @@ def __init__(self, permitted=True):
     def identify(self, request):
         return 123
```
merwok: Should change this for an instance of a named tuple or tiny class.

tests/test_viewderivers.py
```diff
@@ -2086,7 +2086,10 @@ def __init__(self, permitted=True):
...
 
-    def permits(self, request, context, identity, permission):
+    def authenticated_userid(self, request):
+        return 123
```
merwok: See discussion about userid being required to be a string

src/pyramid/config/testing.py
```diff
@@ -51,19 +53,21 @@ def testing_securitypolicy(
...
         :func:`pyramid.security.principals_allowed_by_permission`.
 
         .. versionadded:: 1.4
            The ``remember_result`` argument.
 
         .. versionadded:: 1.4
            The ``forget_result`` argument.
+
+        .. versionchanged:: 2.0
+           Removed ``groupids`` argument and doc about effective principals.
```
merwok: Review other changed methods to see if we need more versionchanged notes

merwok:
Replaced by #3545

PR #3545, opened 2019-12-15
Security policy changes

### LINKED ISSUE #3520: Change `__str__` requirement for identity

luhn:
In this pull request:

- Add `authenticated_userid` to `ISecurityPolicy`
- Add keyword arguments to `ISecurityPolicy.forget`
- Remove `identity` from `ISecurityPolicy.permits`

Closes #3520


luhn:
@merwok, thank you for your work!  I made a couple changes to your implementation:

- I brought `identity` back into `permits`.  I think it makes more sense to have it in there, and I don't recall any discussion around taking it out.
- `unauthenticated_userid` was not backwards compatible, so I redid the implementation.
- I undid marking `unauthenticated_userid` as deprecated.  This was an intentional omission, as it's a commonly-used function and we didn't wish to spam users.

Everything else was just tests and docs.

luhn:
@mmerickel, are you okay with the PR covering two features plus some miscellaneous formatting changes?  I can split it up if you'd like.

I wrote docs to show security policies implemented with `authenticated_userid` pulling from `identify`, so that they validate user ID exists in the database.  Let me know if I misunderstand how they're meant to be implemented.

luhn:
One concern is 0f1ef0d4885ab2fd99d1cf2ccc92886c5519f651, where some outdated tests suddenly started failing because the result of `get_current_registry` changes between master and this branch.  I don't understand why that would change, which makes me nervous.

docs/narr/security.rst
```diff
@@ -32,14 +32,11 @@ how it works at a high level:
...
-- If a :term:`security policy` is in effect, it is passed the request and
-  returns the :term:`identity` of the current user.
-
 - If a :term:`security policy` is in effect and the :term:`view
   configuration` associated with the view callable that was found has a
-  :term:`permission` associated with it, the policy is passed the
-  :term:`context`, the current :term:`identity`, and the :term:`permission`
-  associated with the view; it will allow or deny access.
+  :term:`permission` associated with it, the policy is passed :term:`request`,
+  the :term:`context`,  and the :term:`permission` associated with the view; it
```
stevepiercy: Remove extra space

```suggestion
  the :term:`context`, and the :term:`permission` associated with the view; it
```

luhn:
Okay, I changed the docs in this PR to four spaces where I could find it.

Should I do one-sentence-per-line as well?  If I start I'm not sure where I should stop.

docs/narr/security.rst
```diff
@@ -137,8 +141,11 @@ For example, our above security policy can leverage these helpers like so:
             self.helper = SessionAuthenticationHelper()
 
         def identify(self, request):
-            """ Return the user ID stored in the session. """
-            return self.helper.identify(request)
+            userid = self.helper.authenticated_userid(request)
+            return load_identity_from_db(request, userid)
+
+        def authenticated_userid(self, request):
+            return self.identify(request).id
```
mmerickel: identify can return `None`

docs/whatsnew-2.0.rst
```diff
@@ -95,9 +95,8 @@ The new :attr:`pyramid.request.Request.authenticated_identity` property will
 output the same result as :attr:`pyramid.request.Request.authenticated_userid`.
 
 If using a security policy,
-:attr:`pyramid.request.Request.unauthenticated_userid` and
-:attr:`pyramid.request.Request.authenticated_userid` will both return the
-string representation of the :term:`identity`.
+:attr:`pyramid.request.Request.authenticated_userid` will return the same value
+as :attr:`pyramid.request.Request.authenticated_userid`.
```
mmerickel: I'm not clear what this is trying to convey but it's not quite right as is.

docs/glossary.rst
```diff
@@ -304,8 +304,8 @@ Glossary
 
    identity
      An identity is an object identifying the user associated with the
-     current request.  The identity can be any object, but should implement a
-     ``__str__`` method that outputs a corresponding :term:`userid`.
+     current request.  The identity can be any object, but security policies
+     should ensure that it represents a valid user (not deleted or deactivated).
```
mmerickel: I don't think it's required here that they aren't deleted or deactivated etc but `permits` and anything else that uses it does need to be aware of whatever this opaque object represents.

luhn:
@merwok, I just saw there was discussion about dropping `identity` from `permits`.  Apologies.

> That being said, if you want to drop it from authenticated_userid then drop it from permits as well and the policy can do things in the order it wishes.

https://github.com/Pylons/pyramid/issues/3520#issuecomment-545542789

So I'll revert my commit, back to your original implementation with no `identity`.

mmerickel:
> Should I do one-sentence-per-line as well? If I start I'm not sure where I should stop.

I'd advocate new docs use one-sentence-per-line but that when editing old docs I can take it or leave it based on whatever feels good to you.

docs/glossary.rst
mmerickel: `a valid user with respect to the application (e.g. not deleted).`

docs/narr/security.rst
```diff
@@ -73,8 +70,15 @@ A simple security policy might look like the following:
...
-            return request.session.get('userid')
+            """ Return app-specific user object. """
+            userid = request.session.get('userid')
+            if userid is None:
+                return None
+            return load_identity_from_db(request, userid)
+
+        def authenticated_userid(self, request):
+            """ Return a string ID for the user. """
+            return self.identify(request).id
```
mmerickel: identify can return `None`

merwok:
Why deprecate Request.unauthenticated_userid in docs but not in the code?

luhn:
Since it’s commonly used, it may spam users with warnings. Since it just proxies authenticated_userid for new policies, which is the same thing but with stricter guarantee, there’s no great motivation to get people off of it. 

— Theron

merwok:
Do you mean: commonly used by libs/apps made for older Pyramid?
In other words, people won’t be encouraged/forced to update their authn/authz settings to a security policy?

luhn:
Both. 

Since unauthenticated_userid plays nicely with the new policy, we can be gentle with the deprecation. We’ll be keeping it around for a while.

Things like effective_principals that straight up don’t work with new policy get warnings. 

— Theron

stevepiercy:
> Should I do one-sentence-per-line as well? If I start I'm not sure where I should stop.

I agree with @mmerickel's advocacy. When I edit docs now, I usually fix only those sentences that I edit. However, because I know that I will unwrap all sentences eventually, and if I have time, I will unwrap the sentences in the paragraph or the entire file.

luhn:
As discussed, I've removed the documented guarantees for identity, it's now up to the application.

merwok:
FWIW I am testing the branch in a small project and am getting two warnings from the `effective_principals` route/view predicate, even though my code doesn’t use them:

```
.../site-packages/pyramid/config/views.py:1209: DeprecationWarning: EffectivePrincipalsPredicate: The new security policy has removed the concept of principals.  See https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html#upgrading-authentication-authorization for more information.
    ('effective_principals', p.EffectivePrincipalsPredicate),

.../site-packages/pyramid/config/routes.py:532: DeprecationWarning: EffectivePrincipalsPredicate: The new security policy has removed the concept of principals.  See https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html#upgrading-authentication-authorization for more information.
    ('effective_principals', p.EffectivePrincipalsPredicate),
```

mmerickel:
ah we probably need to place the deprecation inside the constructor or in add_view and add_route directly instead of on the actual predicate object

luhn:
Interesting, looks the framework registering `EffectivePrincipalsPredicate` triggers the warning.  Maybe I can deprecate `__init__` instead?  I'll play around with it.  Thanks for pointing that out.

merwok:
That fixes it!

docs/narr/security.rst
```diff
@@ -72,12 +69,24 @@ A simple security policy might look like the following:
...
+            if validate_userid(request, userid):
+                return userid
+            else:
+                return None
+
         def identify(self, request):
-            """ Return the user ID stored in the session. """
-            return request.session.get('userid')
+            """ Return app-specific user object. """
+            userid = self.authenticated_userid
```
merwok: This definition is circular (see line 74)

docs/narr/security.rst
```diff
@@ -73,11 +70,22 @@ A simple security policy might look like the following:
 
     class SessionSecurityPolicy:
         def identify(self, request):
-            """ Return the user ID stored in the session. """
-            return request.session.get('userid')
+            """ Return app-specific user object. """
+            userid = request.session.get('userid')
+            if userid is None:
+                return None
+            return load_identity_from_db(request, userid)
```
merwok: :rocket:

docs/narr/security.rst
```diff
@@ -73,11 +70,22 @@ A simple security policy might look like the following:
...
+            """ Return a string ID for the user. """
+            identity = request.authenticated_identity
+            if identity is None:
+                return None
+            return string(identity.id)
 
-        def permits(self, request, context, identity, permission):
+        def permits(self, request, context, permission):
             """ Allow access to everything if signed in. """
+            identity = self.identify(request)
```
merwok: When writing a security policy, we end up calling `self.identify` in all other methods.
Using `request.authenticated_identity` feels like adding a non-obvious layer of indirection when the `identify` method is right there.  But it is wasteful to process the request and fetch/create an object in every call!  So if `request.authenticated_identity` is cached, maybe this idiom (like on line 81) should be called out explicitly in the docs, with its caching benefit explained (but maybe in a later section, not here under `A simple security policy might look like the following`).

src/pyramid/interfaces.py
```diff
@@ -483,33 +483,33 @@ def __call__(self, **kw):
...
-        outputs a corresponding :term:`userid`.
+    def authenticated_userid(request, identity):
+        """ Return a :term:`userid` string identifying the trusted and
+        verified user, or ``None`` if unauthenticated.
+        """
 
+    def identify(request):
+        """ Return the :term:`identity` of the current user.  The object can be
+        of any shape, such as a simple ID string or an ORM object.
         """
```
merwok: Detail: this could be the first method defined in the interface.

src/pyramid/interfaces.py
```diff
@@ -483,33 +483,33 @@ def __call__(self, **kw):
...
         of the given identity is allowed the ``permission`` in the current
         ``context``, else return an instance of
         :class:`pyramid.security.Denied`.
-
         """
 
     def remember(request, userid, **kw):
         """ Return a set of headers suitable for 'remembering' the
         :term:`userid` named ``userid`` when set in a response.  An
         individual authentication policy and its consumers can
```
merwok:
```suggestion
        individual security policy and its consumers can
```

docs/narr/security.rst
merwok: After re-reading the ticket, I see that the caching item will be done in another discussion and PR, which seems good.  In this PR, I think the simplest thing to do is call `self.identify()` in all examples to keep things straightforward.

docs/glossary.rst
merwok: I guess my django experience showed!  The auth module there changed somewhat recently to refuse login for deactivated accounts.  I suppose it’s a convenience, but does not cover all needs: new account with unvalidated email, suspended but can still see some pages, etc.

I agree that Pyramid auth is more flexible and identify’s scope more limited: return an object or none, let other methods decide permissions.

src/pyramid/interfaces.py
luhn: I wasn't sure which way to order them.  On one hand, suggested workflow is `identify` -> `authenticated_userid`, so by that lens `identify` comes first.  On the other hand, `authenticated_userid` is the "important" one, identity is just for convenience.

If you have a strong opinion I'm happy to reorder them.

merwok: I don’t have an opinion yet; I’m still digesting the recent discussions (changes to the status and name of `identify` method).  I am installing from your branch in my project, let me see what it looks like now (I was using a dict for identity, so I will rewrite it now to rather rely on `unauthenticated_userid` in `permits`, makes more sense for this case!)

mmerickel:
Where did this end up? Is this PR complete and integrated caching is deferred to a separate PR?

luhn:
Yeah, I was planning on doing caching separately, so this PR is feature-complete.

src/pyramid/predicates.py
```diff
@@ -285,6 +285,14 @@ def __init__(self, val, config):
...
             self.val = set((val,))
 
+    __init__ = deprecated(
+        __init__,
+        'The new security policy has removed the concept of principals.  See '
+        'https://docs.pylonsproject.org/projects/pyramid/en/latest'
+        '/whatsnew-2.0.html#upgrading-authentication-authorization '
+        'for more information.',
+    )
+
```
mmerickel: We should move this deprecation into `add_view` and `add_route` such that the `stacklevel` can be set appropriately to match user code. For reference, here is the `check_csrf` deprecation which was done in 1.10-branch. https://github.com/Pylons/pyramid/blob/bfa2ca0684c0821be8d50718169adc0a3dee8480/src/pyramid/config/views.py#L823-L834

luhn:
Thanks for fixing that stacktrace issue 🍻

mmerickel:
I have a few compatibility fixes incoming but I'll make a separate PR for them.

luhn:
What compatibility issues did you run into?

mmerickel:
It's small stuff. I'm still getting into it a bit but for starters I want to raise inside `request.unauthenticated_userid` and `request.effective_principals` if not using the legacy security policy. I'll probably open a PR tomorrow. I also realized the errors are all linking to the `/latest/` docs and that concerns me a bit so I wanted to tweak them to something closer to what I put in the effective principals changes instead of an actual link that might break. Historically we haven't maintained many links to the docs inside error messages because they feel pretty fragile.

mmerickel:
#3550 opened to address comments above - I decided not to raise to better preserve bw-compat for now.

PR #3546, opened 2019-12-16
set the minimum supported version at py36

mmerickel:
I could use some help finishing this out. Need to track down other places (probably in the docs and tutorials) that python versions are referenced.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,13 @@ sudo: false
 
 matrix:
     include:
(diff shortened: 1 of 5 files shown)


stevepiercy:
Was there a reason to drop 3.5? I'm guessing it has to do with f-strings? [Py35 EOL is another 9 months off](https://devguide.python.org/#status-of-python-branches). I can pitch in, if you decide to proceed.

PR #3547, opened 2019-12-16
set the minimum supported version at py35

mmerickel:
I could use some help finishing this out. Need to track down other places (probably in the docs and tutorials) that python versions are referenced.


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     lint,
-    py34,py35,py36,py37,pypy3,
+    py36,py37,pypy3,
```
stevepiercy:
```suggestion
    py36,py37,py38,pypy3,
```

setup.py
```diff
@@ -94,7 +94,7 @@ def readfile(name):
     package_dir={'': 'src'},
     include_package_data=True,
     zip_safe=False,
-    python_requires='>=3.4',
+    python_requires='>=3.6',
```
stevepiercy: Also update the trove classifiers above.

stevepiercy:
By dropping 3.5, we need to revise our [Python support policy](https://docs.pylonsproject.org/projects/pyramid/en/master/narr/upgrading.html#python-support-policy).

But first the wording is kind of klunky.

> At the time of a Pyramid version release, each supports all versions of Python
through the end of their lifespans.

I think the intent was:

> When a major version of Pyramid is released, it is supported to work on each active version of Python at that time through the end of its lifespan.

And finally, perhaps add a sentence?

> For the Pyramid 2.0 release, an exception was made to that policy to drop support of Python 3.5 ahead of its end-of-life date.
> [Insert optional reasons here.]

stevepiercy:
I found this while scanning for 3.5. Candidate for removal?

https://github.com/Pylons/pyramid/blob/497e667e41b83193d7c5a4f74e9419320d755c46/src/pyramid/config/actions.py#L562-L567

mmerickel:
I’m fine with keeping 3.5 if we have docs around it. There aren’t 3.6 specific features yet. If we were doing more with typing then there might be good reason.

stevepiercy:
The docs don't need 3.5. There were a couple mentions that we support 3.5, and that was it. When I last updated the docs, I bumped up almost everything to 3.7.

I think the choice is whether you want to open the door to 3.6 features or have 3.5 block their implementation.

mmerickel:
> By dropping 3.5, we need to revise our Python support policy.

I don't really want to make an exception for 3.6. If our policy is to support 3.5 then we should support 3.5.

mmerickel:
the failing CI is around an unrelated issue with cookiecutter

PR #3550, opened 2019-12-24
security policy docs and legacy policy improvements

mmerickel:
- Added `set_security_policy`` to more places in the docs.
- Ensure that the authn/authz policies are not used at all if the legacy
  policy is not in effect to avoid edge cases where the code would skip
  the security policy and use the authn/authz policy on accident.
- Change deprecation warnings in code to reference the docs by name
  instead of by URL.

The only code/behavior changes are in `request.effective_principals` and `request.unauthenticated_userid`.

I am working on the tutorials in a followup PR.

cc @luhn @merwok

### DIFF
--- a/docs/narr/advconfig.rst
+++ b/docs/narr/advconfig.rst
@@ -297,22 +297,22 @@ Methods Which Provide Conflict Detection
 
 These are the methods of the configurator which provide conflict detection:
 
-:meth:`~pyramid.config.Configurator.add_view`,
-:meth:`~pyramid.config.Configurator.add_route`,
-:meth:`~pyramid.config.Configurator.add_renderer`,
-:meth:`~pyramid.config.Configurator.add_request_method`,
-:meth:`~pyramid.config.Configurator.set_request_factory`,
-:meth:`~pyramid.config.Configurator.set_session_factory`,
-:meth:`~pyramid.config.Configurator.set_root_factory`,
-:meth:`~pyramid.config.Configurator.set_view_mapper`,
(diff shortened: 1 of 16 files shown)


src/pyramid/config/routes.py
```diff
@@ -335,9 +335,10 @@ def add_route(
         if 'effective_principals' in predicates:
             warnings.warn(
                 (
-                    'The new security policy has removed the concept of '
-                    'principals. See "Upgrading Authentication/Authorization" '
-                    'in "What\'s New in Pyramid 2.0" for more information.'
+                    'The new security policy has deprecated '
+                    'effective_principals. See "Upgrading '
+                    'Authentication/Authorization" in "What\'s New in '
+                    'Pyramid 2.0" for more information.'
```
stevepiercy:
```suggestion
                    'Pyramid 2.0" in the Pyramid documentation for more information.'
```

It might not be obvious for the user to look in the documentation. Update elsewhere, too.

docs/narr/advconfig.rst
```diff
@@ -307,6 +307,7 @@ These are the methods of the configurator which provide conflict detection:
 :meth:`~pyramid.config.Configurator.set_view_mapper`,
 :meth:`~pyramid.config.Configurator.set_authentication_policy`,
 :meth:`~pyramid.config.Configurator.set_authorization_policy`,
+:meth:`~pyramid.config.Configurator.set_security_policy`,
```
merwok: This line could be before the deprecated methods above

mmerickel: I'll just re-order alphabetical.

Issue #3434, opened 2018-11-27
Migrate content from `/narr/scaffolding.rst` to `/narr/cookiecutters.rst`


stevepiercy:
Review content in `/narr/scaffolding.rst` from the `1.10-branch` and port relevant bits to `/narr/cookiecutters.rst` on `master`.

See also https://github.com/Pylons/trypyramid.com/issues/255

mmerickel:
@stevepiercy any plans to make this happen? If not I'm removing the 2.0 milestone.

stevepiercy:
Yes. Do we have a timeline for a 2.0 release?

mmerickel:
This week I’m looking at an alpha if only docs stuff is left.

stevepiercy:
Opened PR #3552. There wasn't a whole lot that can be migrated, but I expanded it a bit.

stevepiercy:
Closed by #3552

PR #3552, opened 2019-12-25
Expand cookiecutters doc to include more references, using the old sc…

stevepiercy:
…affolds.rst as an example.

- Update reST target and label for Pyramid cookiecutters.

See #3434.

### DIFF
--- a/docs/narr/cookiecutters.rst
+++ b/docs/narr/cookiecutters.rst
@@ -1,12 +1,12 @@
 .. index::
     single: cookiecutter
 
-.. _pyramid-cookiecutters:
+.. _cookiecutters:
 
 Pyramid cookiecutters
 =====================
 
-A :term:`cookiecutter` is a command-line utility that creates projects from `cookiecutters <https://cookiecutter.readthedocs.io/en/latest/>`_ (project templates).
+A :term:`cookiecutter` is a command-line utility that creates projects from `cookiecutters <https://cookiecutter.readthedocs.io/en/latest/>`__ (project templates).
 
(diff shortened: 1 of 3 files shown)


docs/narr/cookiecutters.rst
```diff
@@ -1,11 +1,20 @@
-.. _cookiecutters:
+.. index::
+    single: cookiecutter
+
+.. _pyramid-cookiecutters:
```
mmerickel: In a previous commit I removed the cookiecutter intersphinx link. I'd probably say you keep naming this `cookiecutters` instead of renaming to `pyramid-cookiecutters`.

stevepiercy: Oh, this is the problem: https://github.com/sphinx-doc/sphinx/issues/3921

I removed that second underscore on line 9 in an earlier commit, thinking it was a typo, but it actually has purpose. That in turn caused me to rename the reference to avoid these errors:

```
pickling environment... /Users/stevepiercy/projects/hack-on-pyramid/docs/narr/cookiecutters.rst:7: WARNING: Duplicate explicit target name: "cookiecutters".

...snip...

/Users/stevepiercy/projects/hack-on-pyramid/docs/narr/project.rst:76: WARNING: undefined label: cookiecutters (if the link has no caption the label must precede a section header)
/Users/stevepiercy/projects/hack-on-pyramid/docs/quick_tutorial/ini.rst:135: WARNING: undefined label: cookiecutters (if the link has no caption the label must precede a section header)
```

I restored both the name and double underscore, and all is well. Thanks!

PR #3554, opened 2019-12-26
isort

mmerickel:
This config basically does what we've done in the past and remains compatible with black. The main tweak is to group everything that isn't pyramid into one section at the top without attempting to differentiate stdlib vs third party as I think this fits the goal better in the face of things like polyfill libraries.

It changes the tox helper for black/isort to `tox -e format` and continues to maintain `tox -e lint` for checking in CI.

### DIFF
--- a/HACKING.txt
+++ b/HACKING.txt
@@ -48,7 +48,7 @@ In order to add a feature to Pyramid:
 Coding Style
 ------------
 
-- Pyramid uses Black (https://pypi.org/project/black/) and isort (https://pypi.org/project/black/) for code formatting style.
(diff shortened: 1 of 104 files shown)


HACKING.txt
```diff
@@ -48,11 +48,10 @@ In order to add a feature to Pyramid:
 Coding Style
 ------------
 
-- Pyramid uses Black for code formatting style.
-  https://pypi.org/project/black/
-  To run Black:
+- Pyramid uses Black (https://pypi.org/project/black/) and isort (https://pypi.org/project/black/) for code formatting style.
```
luhn:
```suggestion
- Pyramid uses Black (https://pypi.org/project/black/) and isort (https://pypi.org/project/isort/) for code formatting style.
```

Issue #3523, opened 2019-10-19
Update docs to sync with https://github.com/Pylons/pyramid-cookiecutter-starter/pull/71


stevepiercy:
With the PR Pylons/pyramid-cookiecutter-starter#71 we need to synchronize the files used in tutorials and docs for all steps.

stevepiercy:
To kill two birds use Python 3.8, and update the output for all tutorials.

mmerickel:
It's probably time to do this, pending the open PRs to the cookiecutter get merged. I don't see anything bw-incompat or out of date in the cookiecutter other than the pending PRs.

mmerickel:
Well - we need to do https://github.com/Pylons/pyramid/issues/3522. Do you want to do it or should I?

stevepiercy:
I can have a look tonight, and if I get stuck, I'll hand it over to you.

mmerickel:
Don’t forgo your holidays!

stevepiercy:
- The Pyramid Issue #3522 was already addressed.
- Previously open PRs in pyramid-cookiecutter-starter have been merged.
- I've opened a new [PR to drop support for 3.4 and add for 3.8](https://github.com/Pylons/pyramid-cookiecutter-starter/pull/74). If it passes, I'll merge it, then start work on this issue.

PR #3556, opened 2019-12-27
[WIP] Update docs to sync with cookiecutter master branch

stevepiercy:
- [x] `docs/quick_tour.rst`
- [x] `docs/narr/project.rst`
- [x] `docs/quick_tutorial/*`
- [x] `docs/tutorials/wiki/*`
- [x] `docs/tutorials/wiki2/*`
- [x] Any files that refer to `docs/narr/myproject`

See #3523


stevepiercy:
This monster is ready for review.

mmerickel:
Let's say hypothetically I moved the tests out of the package. On a scale from -1 to +1 how annoyed would you be to re-sync this again?

stevepiercy:
+1. Tests are the least troublesome to synch, as they are at the end of every tutorial. It'll have to wait till Monday night though.

mmerickel:
OK I will try to sort it out tomorrow so that it’s ready for you on Monday. Thank you!

stevepiercy:
If you're feeling ambitious, test coverage for both the sqlalchemy and zodb backends are low, and maybe move source into `/src/package`?

mmerickel:
I do not want to move the source into `/src/package` - I don't do it in my own apps and I think it's mostly valuable for libraries versus apps. I have made the change to move the tests but won't do anything to increase coverage atm.

stevepiercy:
@mmerickel I could use some help with one failing test.

https://github.com/Pylons/pyramid/blob/6c5f524cdb750a7a3a112087942a0eafc8e10e4d/docs/tutorials/wiki2/src/tests/tests/test_initdb.py#L12-L16

```
E       sqlite3.IntegrityError: UNIQUE constraint failed: users.name
```

The database already exists and has been populated with data from an earlier step in the tutorial when changing the models. If I change the database name in `development.ini` to `tutorialtest.sqlite` and adjust the test accordingly, it passes. Obviously that is not the right way, but I couldn't figure out the right way.

Also there are a couple of warnings, one for `imp`, which will go away in a future release, but there are a couple of other warnings:

```
tests/test_functional.py::FunctionalTests::test_anonymous_user_cannot_add
tests/test_functional.py::FunctionalTests::test_anonymous_user_cannot_add
  /Users/stevepiercy/projects/hack-on-pyramid/tutorial/env/lib/python3.7/site-packages/pkg_resources/__init__.py:1145: DeprecationWarning: Use of .. or absolute path in a resource path is not allowed and will raise exceptions in a future release.
    self, resource_name
```

I fiddled around with that for a while, but nothing obvious jumped out at me.

stevepiercy:
This PR is now ready for review, and except for the above mentioned failing test and test warnings, can be merged.

mmerickel:
The test should be either

1) removed

2) modified to just skip itself if the database already exists

in its current form, initialize_db is meant to be run only one time ever. So if the database already exists it will fail to run again. The bug here (imo) is having this test in the first place and I'd just remove it.

stevepiercy:
Thanks. Removal was my first thought. Tests should be repeatable, and modifying the script would feel like accommodating something that doesn't need to be accommodated.

stevepiercy:
I rebased this branch on master, resolved conflicts and merged, then force pushed. I guess that wasn't the proper procedure, looking at the commit history.

mmerickel:
is this ready for review then or still wip?

stevepiercy:
The only thing missing is how to resolve the warnings in https://github.com/Pylons/pyramid/pull/3556#issuecomment-570122485

I couldn't figure it out. Otherwise it is ready for review.

merwok:
I would edit the pkg_resources code in site-packages to add/change `stacklevel` parameter to `warnings.warn` in order to find the faulty code.

stevepiercy:
Thanks for the tip! I'll try that out tonight, after daylight runs out.

stevepiercy:
PyCharm caching of diff will be the death of me. Refreshing cache revealed the file diff.

Thanks to the tip to modifying `pkg_resources`, I found a few files that did not show up in PyCharm's diff that used `../` instead of `tutorial:` for the `renderer`. I had to change `msg = "Use of .. or absolute path in a resource path is not allowed."` to include `str(path.split(posixpath.sep))`, which revealed the offending files.

Changes pushed, and this PR is now complete and ready for review. Ping @mmerickel.

docs/narr/project.rst
```diff
@@ -1159,11 +1166,9 @@ inotify support
...
 
 By default ``hupper`` will poll the filesystem for changes to all Python
-code. This can be pretty inefficient in larger projects. To be nicer to your
-hard drive, you should install the
-`watchdog <https://pythonhosted.org/watchdog/>`_ package in development.
-``hupper`` will automatically use ``watchdog`` to more efficiently poll the
-filesystem.
+code. This can be pretty inefficient in larger projects.
+To be nicer to your hard drive, you should install either the `watchman <https://facebook.github.io/watchman/>`_ or `watchdog <https://pythonhosted.org/watchdog/>`_ package in development.
+``hupper`` will use, in order of preference for efficiency, if available, ``watchman``, ``watchdog``, then finally ``inotify`` to poll the filesystem.
```
mmerickel: watchman, watchdog or simple polling

mmerickel:
I pushed one small tweak to find_packages. This is great work. I'm going to merge it and will try to update the tutorials this weekend.

PR #3561, opened 2020-01-08
Request local cache

### LINKED ISSUE #3549: Cache identity
Seems like caching identity is a popular request, but has been difficult to figure out how to properly deal with it.  Opening this issue for discussion.   Some ideas off the top of my head:

- Bust cache on request.authenticated_userid change
- Bust cache on remember/forget
- Have `identify` return a two-tuple of the identity and a boolean on whether to cache or not.
- Have a `request.authenticated_identity` and `request.cached_identity`
- Have a manual `bust_identity_cache()` function

mmerickel:
fixes #3549

### DIFF
--- a/src/pyramid/request.py
+++ b/src/pyramid/request.py
@@ -378,6 +378,7 @@ class RequestLocalCache:
     .. versionadded:: 2.0
 
     """
+
     NO_VALUE = Sentinel('NO_VALUE')
 
     def __init__(self):
@@ -393,10 +394,13 @@ class RequestLocalCache:
                 return decorator(self.inst, self.__wrapped__)(*args, **kwargs)
 
             def __get__(self, instance, owner):
-                return decorator(self.inst, self.__wrapped__.__get__(instance, owner))
+                return decorator(
+                    self.inst, self.__wrapped__.__get__(instance, owner)
+                )
 
         def adapt(inst, fn):
             return FunctionOrMethodAdapter(inst, fn)
+
         return adapt
 
     @auto_adapt
@@ -405,6 +409,7 @@ class RequestLocalCache:
         Decorate a method or function.
 
         """
+
         @functools.wraps(fn)
         def wrapper(request, *args, **kwargs):
             result = self.get(request)
@@ -413,6 +418,7 @@ class RequestLocalCache:
                 self.set(request, result)
                 request.add_finished_callback(self.clear)
             return result
+
         return wrapper
 
     del auto_adapt
--- a/src/pyramid/request.py
+++ b/src/pyramid/request.py
@@ -1,4 +1,6 @@
 from collections import deque
+import functools
+import weakref
 from webob import BaseRequest
 from zope.interface import implementer
 from zope.interface.interface import InterfaceClass
@@ -17,6 +19,7 @@ from pyramid.url import URLMethodsMixin
 from pyramid.util import (
     InstancePropertyHelper,
     InstancePropertyMixin,
+    Sentinel,
     bytes_,
     text_,
 )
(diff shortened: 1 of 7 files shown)


mmerickel:
@luhn you should probably look at this - I have a PR for the security docs that I'll push once this is in.

src/pyramid/request.py
```diff
@@ -330,3 +333,117 @@ def apply_request_extensions(request, extensions=None):
...
+
+        """
+        return self.store.get(request, default)
+
+    def set(self, request, value):
+        """
+        Update the cache with a new value.
+
+        """
+        self.store[request] = value
```
luhn: Should `set()` do an `add_finished_callback`?  One the other hand, technically not necessary due to the weakref, and could result in multiple redundant callbacks.

src/pyramid/request.py
```diff
@@ -330,3 +333,117 @@ def apply_request_extensions(request, extensions=None):
...
+        return adapt
+
+    @auto_adapt
+    def __call__(self, fn):
+        """
+        Decorate a method or function.
+
+        """
+
+        @functools.wraps(fn)
```
luhn: This whole auto_adapt thing negates `wraps`, it swallows `__name__`, `__doc__`, et al.

```python
from pyramid.request import RequestLocalCache


cache = RequestLocalCache()


@cache
def myfunc(request):
    """
    Return hello.

    """
    return 'hello'


print(myfunc.__name__)  # should print "myfunc", but raises AttributeError
print(myfunc.__doc__)  # should print "Return hello", but prints "None"
```

src/pyramid/request.py
mmerickel: yes I need to move the callback registration from the decorator into this method, good catch.

The redundant callbacks is the cost of doing business and I'm not going to try to fix it. Technically only one is necessary per request but more is a minimal concern at this point.

src/pyramid/request.py
mmerickel: Yeah I couldn't come up with an approach that preserved that stuff correctly unfortunately. The cost of doing business unless someone can come up with a better approach. I looked at using the `wrapt` package initially and it returned things like `FunctionWrapper` objects, etc so I decided this wasn't really worse than that.

I'm comfortable with the current approach and am pretty certain we can fix it in the future without compatibility concerns so unless you actually know what to do, I'm gonna leave it.

luhn:
I was playing around with auto_adapt and got into a rabbit hole.  It looks like decorators that support both methods and functions is a very difficult problem.  Django struggled with this problem and ended up with decorator pairs, one for functions and one for methods.  https://groups.google.com/d/msg/django-developers/MCSxSkf1QE0/s768z_V2afMJ

src/pyramid/request.py
```diff
@@ -330,3 +333,117 @@ def apply_request_extensions(request, extensions=None):
...
+    Instantiate a cache object and use it to decorate methods or functions
+    that accept a request parameter. Any other arguments are not used as
+    cache keys so make sure they are constant across calls per-request.
+
+    Wrapping methods:
+
+    .. code-block:: python
+
+        class SecurityPolicy:
+            identity_cache = RequestLocalCache()
```
merwok: Is is required that the cache instance lives in the class namespace?
Or would this work as well:
```python
class SecurityPolicy:
    @RequestLocalCache()
    def authenticated_identity(self, request):
        ...
```

mmerickel: Yes your snippet works fine, the goal of the example is to show how to maintain a reference to the cache so you can modify it directly. For example if you wanted to change the value if `forget()` is called, you'd need to be able to reference the cache from there.

src/pyramid/request.py
mmerickel: Ok I've pushed a new API for this and got rid of the original decorator approach. The original approach suffered from some issues like sharing the cache object across instances of the policy when used at the class level, and was error-prone in that it accepted extra arguments into the function. Now it only accepts a request, always, and it's happy.

mmerickel:
Ok the new version changes the approach a bit. The API is much more flexible and much less error prone.

- Basically just a simple cache populated by a creator function.
- The cache or the creator function can be used as the primary api for users.
- The example no longer shares the cache across instances of a class. For example, there should be a new cache created per policy instance.
- There is an imperative `get_or_create` method that can be used directly.
- The finished callback is only registered once per request.

src/pyramid/request.py
```diff
@@ -330,3 +333,132 @@ def apply_request_extensions(request, extensions=None):
...
+        class SecurityPolicy:
+            def __init__(self):
+                self.identity_cache = RequestLocalCache(self.load_identity)
+
+            def load_identity(self, request):
+                result = ...  # do some expensive computations
+                return result
+
+            def authenticated_identity(self, request):
+                return self.identity_cache.get_or_create(request)
```
merwok: This is a fantastic helper.

mmerickel:
Ok, final idea - might be getting a little crazy - but I added `RequestLocalCache.instancemethod` which basically works exactly the same as `reify` but with a per-request argument. It seems to work ok to me but will likely fail in the same way django fails when stacking decorators. Fortunately this object has several workarounds if there are issues.

mmerickel:
Ok I'll remove it for now - I'm fine with less magic.

PR #3557, opened 2019-12-30
update docs to use security policy

mmerickel:
- [X] rename `identify` to `authenticated_identity` on `ISecurityPolicy`
- [X] quick_tutorial
- [x] sync cookiecutter to wiki tutorial
- [x] use security policy in wiki tutorial
- [X] sync cookiecutter to wiki2
- [X] use security policy in wiki2 tutorial
- [x] revise security chapter - this is at least partially complete
- [x] sync cookiecutter to starter project

rendered: https://mmerickel-pyramid-fork.readthedocs.io/en/security-docs/

fixes #3548

PR #3563, opened 2020-01-13
Move acl security to authorization

mmerickel:
This deprecates the ACL-specific parts of `pyramid.security` and moves them into `pyramid.authorization` since ACL features are not specifically top-level in the security API any longer.

Here are the options:

1. merge this, and isolate anything ACL-related in `pyramid.authorization`.
2. come up with a way to un-deprecate effective_principals features so that it can be used by the framework as makes sense.

My feeling is that it's good to isolate this stuff into one spot, and if there is pushback by the community after release because `request.effective_principals` and the predicate are deprecated, then we can re-evaluate them by potentially adding an optional method to the security policy or something.

cc @luhn


luhn:
👍 

I agree it's the right way to organize things, even if its temporarily a bit more messy until we can remove the deprecated items.

Re `effective_principals`, I think if users still want to access it in a new security policy we should recommend they put it in `authenticated_identity`.   It is part of the user's identity, imo.  Also I'm loathe to add an extra method to the security policy that's only applicable for specific use cases.  But I guess we'll see what the response is.

mmerickel:
Ok I will fix the changelog and docs. There are more changes once #3557 is merged as well, so I will leave this open until then.

src/pyramid/security.py
```diff
@@ -114,7 +92,7 @@ def principals_allowed_by_permission(context, permission):
     in effect, return a sequence of :term:`principal` ids that possess
     the permission in the ``context``.  If no authorization policy is
     in effect, this will return a sequence with the single value
-    :mod:`pyramid.security.Everyone` (the special principal
+    :mod:`pyramid.authorization.Everyone` (the special principal
```
merwok: Should the `principals_allowed_by_permission` stay in `pyramid.security`?

mmerickel: no it's fine as-is. The function is deprecated and there is not a way to make it work without changes going forward as it currently depends on an authorization policy. I can't put a version in pyramid.authorization that actually works going forward.

CHANGES.rst
```diff
@@ -7,6 +7,19 @@ Features
 - Add support for Python 3.8.
   See https://github.com/Pylons/pyramid/pull/3547
 
+- New security APIs have been added to support a massive overhaul of the
+  authentication and authorization system. Read
+  "Upgrading Authentication/Authorization" in the "What's New in Pyramid 2.0"
```
merwok: I understand instructions like this in Python warnings, but why not use a `ref` role here in a reST doc?

docs/api/authorization.rst
```diff
@@ -10,3 +10,46 @@
...
   .. autoclass:: ACLAuthorizationPolicy
 
+Constants
+---------
+
+.. attribute:: Everyone
+
+    The special principal id named 'Everyone'.  This principal id is
+    granted to all requests.  Its actual value is the string
+    'system.Everyone'.
```
merwok:
```suggestion
    ``'system.Everyone'``.
```

docs/api/authorization.rst
```diff
@@ -10,3 +10,46 @@
...
+
+.. attribute:: Authenticated
+
+    The special principal id named 'Authenticated'.  This principal id
+    is granted to all requests which contain any other non-Everyone
+    principal id (according to the :term:`authentication policy`).
+    Its actual value is the string 'system.Authenticated'.
```
merwok:
```suggestion
    Its actual value is the string ``'system.Authenticated'``.
```

docs/api/security.rst
```diff
@@ -22,40 +22,56 @@ Authorization API Functions
 Constants
 ---------
 
+.. attribute:: NO_PERMISSION_REQUIRED
+
+	A special permission which indicates that the view should always
+	be executable by entirely anonymous users, regardless of the
+	default permission, bypassing any :term:`authorization policy`
+	that may be in effect.  Its actual value is the string
+	'__no_permission_required__'.
```
merwok:
```suggestion
	``'__no_permission_required__'``.
```

CHANGES.rst
mmerickel: the changes.rst is published to pypi and it does not understand most sphinx directives

docs/api/security.rst
merwok:
```suggestion
	``'__no_permission_required__'``.
```

docs/api/security.rst
```diff
@@ -22,40 +22,56 @@ Authorization API Functions
...
+
 .. attribute:: Everyone
 
     The special principal id named 'Everyone'.  This principal id is
     granted to all requests.  Its actual value is the string
     'system.Everyone'.
```
merwok:
```suggestion
    ``'system.Everyone'``.
```

docs/api/security.rst
```diff
@@ -22,40 +22,56 @@ Authorization API Functions
...
+    .. deprecated:: 2.0
+
+        Moved to :data:`pyramid.authorization.Everyone`.
+
 .. attribute:: Authenticated
 
     The special principal id named 'Authenticated'.  This principal id
     is granted to all requests which contain any other non-Everyone
     principal id (according to the :term:`authentication policy`).
     Its actual value is the string 'system.Authenticated'.
```
merwok:
```suggestion
    Its actual value is the string ``'system.Authenticated'``.
```

docs/api/security.rst
```diff
@@ -66,12 +82,20 @@ Return Values
     'read')`` that means allow access.  A sequence of ACEs makes up an
     ACL.  It is a string, and its actual value is "Allow".
```
merwok:
```suggestion
    ACL.  It is a string, and its actual value is ``'Allow'``.
```

docs/api/security.rst
```diff
@@ -66,12 +82,20 @@ Return Values
...
 
+    .. deprecated:: 2.0
+
+        Moved to :data:`pyramid.authorization.Allow`.
+
 .. attribute:: Deny
 
     The ACE "action" (the first element in an ACE e.g. ``(Deny,
     'george', 'read')`` that means deny access.  A sequence of ACEs
     makes up an ACL.  It is a string, and its actual value is "Deny".
```
merwok:
```suggestion
    makes up an ACL.  It is a string, and its actual value is ``'Deny'``.
```

docs/api/security.rst
```diff
@@ -22,40 +22,56 @@ Authorization API Functions
...
     granted to all requests.  Its actual value is the string
     'system.Everyone'.
 
+    .. deprecated:: 2.0
+
+        Moved to :data:`pyramid.authorization.Everyone`.
+
 .. attribute:: Authenticated
 
     The special principal id named 'Authenticated'.  This principal id
```
merwok:
```suggestion
    The special principal id named ``Authenticated``.  This principal id
```

docs/api/security.rst
```diff
@@ -22,40 +22,56 @@ Authorization API Functions
...
+
+	A special permission which indicates that the view should always
+	be executable by entirely anonymous users, regardless of the
+	default permission, bypassing any :term:`authorization policy`
+	that may be in effect.  Its actual value is the string
+	'__no_permission_required__'.
+
 .. attribute:: Everyone
 
     The special principal id named 'Everyone'.  This principal id is
```
merwok:
```suggestion
    The special principal id named ``Everyone``.  This principal id is
```

CHANGES.rst
merwok: Right!  And you prefer giving instructions rather than risking link rot.

docs/api/security.rst
mmerickel: duplicate

merwok: It’s because I have done the same changes to docs in pyramid.security and pyramid.authorization :slightly_smiling_face:

docs/api/authorization.rst
```diff
@@ -10,3 +10,45 @@
 
   .. autoclass:: ACLAuthorizationPolicy
 
+Constants
+---------
+
+.. attribute:: Everyone
+
+    The special principal id named 'Everyone'.  This principal id is
```
merwok:
```suggestion
    The special principal id named ``Everyone``.  This principal id is
```

docs/api/authorization.rst
```diff
@@ -10,3 +10,45 @@
...
+    granted to all requests.  Its actual value is the string
+    ``'system.Everyone'``.
+
+.. attribute:: Authenticated
+
+    The special principal id named 'Authenticated'.  This principal id
```
merwok:
```suggestion
    The special principal id named ``Authenticated``.  This principal id
```

src/pyramid/authorization.py
```diff
@@ -1,10 +1,51 @@
...
+Authenticated = Authenticated  # api
+Allow = Allow  # api
+Deny = Deny  # api
+
+
+class AllPermissionsList(_AllPermissionsList):
+    pass
+
+
+# subclass to fix __qualname__
```
merwok: Do you mean `__module__`?  `__qualname__` is about nested functions/classes, but does not include module path.

merwok:
> The mix of double-backticking `thing` and `'other'` regarding `'` is inconsistent

If you mean that the doc for pyramid.security docs says
```
the value is the string ``'system.Everyone'``
```
but the one for pyramid.authorization has
```
the value is the string 'system.Everyone'
```
I have sent change suggestions to make them consistent.


If you mean the difference between
```
the name ``Everyone``
```
and
```
the value is the string ``'system.Everyone'``
```
then I don’t see the inconsistency: the name is not a string.

src/pyramid/authorization.py
mmerickel: yep, thanks

mmerickel:
I re-synced some of the docs after merging #3557. This should be good to go.

Issue #3566, opened 2020-03-05
correct the code exmaple for imperative configuration


slobinger:
## Documentation Suggestion
It seems that the code example for [imperative configuration](https://docs.pylonsproject.org/projects/pyramid/en/master/narr/configuration.html#imperative-configuration) is not working.

For me, new to Pyramid it was very confusing because in the code sample provided it is not clear if pyramid somehow guesses how to map routes to views. But according to everything I knew at that time it could not be possible because that would be very opinionated :-). So I tested if the sample code works and get a 404 Error.
I compared the code sample with that provided in the project readme and in this sample it is Totaly clear for me that there is a deterministic mapping.

So it would be nice for all beginners if this sample I linked to would be consistent with the getting started sample.

mmerickel:
The example does work but I agree that it is misleading. The `add_view` call should have a `route_name='hello'` argument.

mmerickel:
Oh wait I read it again and you're right it is broken. Anyway the fix is in my previous comment.

stevepiercy:
@slobinger would you please make a PR against master with the fix? I will backport it to other releases. Thank you!

FWIW, we got it right on https://trypyramid.com/

stevepiercy:
Closed by #3567

PR #3567, opened 2020-03-06
correct configuration.rst add missing route_name

slobinger:
Add missing route_name argument to the call of config.add_view in the imperative configuration example.

according to #3566


stevepiercy:
Thank you!

PR #3571, opened 2020-04-18
py3 features

slav0nic:
some addition to https://github.com/Pylons/pyramid/pull/3421

- `class Foo(object):` replaced with `class Foo:`, `(object)` make no sense in py3
- `__nonzero__()`  -> `__bool__()`, first method was removed from py3
- use new py3 `super()` syntax without `(class, self)`
- `/` replaced with `//`
https://github.com/Pylons/pyramid/blob/48cf45b0a1ac04e701101fa18778164825edd429/src/pyramid/config/predicates.py#L208
in py2 it returned `int` type, now in py3 - `float`, that looks not correct (but not critical :])
- `__unicode__` method was removed in py3, so we don't need any extra checks
- unittest.AssertEqual**s** [was deprecated](https://docs.python.org/3/library/unittest.html#deprecated-aliases)

### DIFF
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -11,9 +11,9 @@ class TestRegistry(unittest.TestCase):
     def _makeOne(self, *args, **kw):
         return self._getTargetClass()(*args, **kw)
 
-    def test___nonzero__(self):
+    def test___bool__(self):
         registry = self._makeOne()
-        self.assertEqual(registry.__nonzero__(), True)
+        self.assertEqual(registry.__bool__(), True)
 
     def test__lock(self):
         registry = self._makeOne()
@@ -379,10 +379,6 @@ class TestIntrospectable(unittest.TestCase):
             repr(inst),
             "<Introspectable category 'category', discriminator 'discrim'>",
         )
-
-    def test___nonzero__(self):
-        inst = self._makeOnePopulated()
-        self.assertEqual(inst.__nonzero__(), True)
 
     def test___bool__(self):
         inst = self._makeOnePopulated()
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -407,7 +407,7 @@ class TestIntrospectable(unittest.TestCase):
         )
 
 
-class DummyIntrospector(object):
+class DummyIntrospector:
     def __init__(self):
         self.intrs = []
         self.relations = []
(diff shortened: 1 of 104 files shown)


tests/test_registry.py
```diff
@@ -380,13 +380,9 @@ def test___repr__(self):
...
         )
 
-    def test___nonzero__(self):
-        inst = self._makeOnePopulated()
-        self.assertEqual(inst.__nonzero__(), True)
-
     def test___bool__(self):
         inst = self._makeOnePopulated()
-        self.assertEqual(inst.__bool__(), True)
+        self.assertEqual(bool(inst), True)
```
mmerickel, line 385: this is a dangerous change to the test because any object instance will test true if it doesn't override `__bool__`... I think it'd be better to keep it the way it was and explicitly check that it is defining the protocol

tests/test_registry.py
```diff
@@ -11,9 +11,9 @@ def _getTargetClass(self):
     def _makeOne(self, *args, **kw):
         return self._getTargetClass()(*args, **kw)
 
-    def test___nonzero__(self):
+    def test___bool__(self):
         registry = self._makeOne()
-        self.assertEqual(registry.__nonzero__(), True)
+        self.assertEqual(bool(registry), True)
```
mmerickel, line 16: same issue here, it should invoke `__bool__` directly to ensure it's testing what it wants to test - implementing the protocol

tests/test_testing.py
```diff
@@ -103,13 +103,9 @@ class Dummy:
...
         self.assertEqual(len(resource), 2)
 
-    def test_nonzero(self):
-        resource = self._makeOne()
-        self.assertEqual(resource.__nonzero__(), True)
-
     def test_bool(self):
         resource = self._makeOne()
-        self.assertEqual(resource.__bool__(), True)
+        self.assertEqual(bool(resource), True)
```
mmerickel, line 108: same thing... should use the old version of the code here

slav0nic:
Rebased, changed this commit  https://github.com/Pylons/pyramid/pull/3571/commits/aadf19a5b81645b18e50699fe2f180aeff3409b3

mmerickel:
Good stuff, thank you @slav0nic !

PR #3572, opened 2020-04-18
Fix docs build on Sphinx 3.0+

slav0nic:
`add_directive` arguments was changed in latest sphinx
https://github.com/sphinx-doc/sphinx/blob/v3.0.1/sphinx/application.py#L593


slav0nic:
required much more work

Issue #3573, opened 2020-04-18
Sphinx 3+ support


slav0nic:
Sphinx 3 was released with some changes

Current build (https://travis-ci.org/github/Pylons/pyramid/jobs/676530356#L248) can be fixed with
```diff
diff --git a/docs/conf.py b/docs/conf.py
index 472485380..613f38dca 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -379,9 +379,9 @@ def app_role(role, rawtext, text, lineno, inliner, options={}, content=[]):
 
 def setup(app):
     app.add_role('app', app_role)
-    app.add_directive('frontmatter', FrontMatter, 1, (0, 0, 0))
-    app.add_directive('mainmatter', MainMatter, 1, (0, 0, 0))
-    app.add_directive('backmatter', BackMatter, 1, (0, 0, 0))
+    app.add_directive('frontmatter', FrontMatter)
+    app.add_directive('mainmatter', MainMatter)
+    app.add_directive('backmatter', BackMatter)
     app.connect('autodoc-process-signature', resig)
```
but building still failed https://travis-ci.org/github/Pylons/pyramid/jobs/676546160#L412
```
/home/travis/build/Pylons/pyramid/.tox/docs/lib/python3.6/site-packages/pyramid/request.py:docstring of pyramid.request.Request.response:1:duplicate object description of pyramid.request.Request.response, other instance in api/request, use :noindex: for one of them

Makefile:105: recipe for target 'doctest' failed
```

without -W flag here is so many warning:
<details><summary>Warnings list</summary>

```
/mnt/data/coding/pyramid/src/pyramid/request.py:docstring of pyramid.request.Request.response:1: WARNING: duplicate object description of pyramid.request.Request.response, other instance in api/request, use :noindex: for one of them
/mnt/data/coding/pyramid/src/pyramid/request.py:docstring of pyramid.request.Request.session:1: WARNING: duplicate object description of pyramid.request.Request.session, other instance in api/request, use :noindex: for one of them

/mnt/data/coding/pyramid/src/pyramid/config/__init__.py:docstring of pyramid.config.Configurator:43: WARNING: term not in glossary: deployment settings
/mnt/data/coding/pyramid/src/pyramid/config/__init__.py:docstring of pyramid.config.Configurator.include:8: WARNING: term not in glossary: Configurator
[...]

PR #3576, opened 2020-04-23
Allow multiple values for header predicate

### LINKED ISSUE #3575: Allow multiple values for `header` predicate
## Feature Request

**Is your feature request related to an issue? Please describe.**
We can use predicates like `header="X-GitHub-Whatever"` or `header="X-Facebook-Thing:cool-value"` within a route or view config.  However, it is not possible to specify more than one header, so one would need to write a custom predicate here, or perform checks inside the view.

**Describe the solution you'd like**
`header` should accept a string or tuple of strings, like `request_param` does.

**Describe alternatives you've considered**
New parameter `headers` always taking a sequence—would not fit with existing behaviour for `request_param` and others.

**Additional context**
Implementation should be easy in `pyramid.predicates.HeaderPredicate`.
Tuples were not accepted before, so no compatibility issue.

merwok:
fixes #3575 

- [ ] validate wording in docstrings
- [ ] also edit narrative doc


merwok:
Test failure is a sphinx issue that's certainly on master too

stevepiercy:
@merwok see #3573 for docs failure. You can fix the lint failure in the meanwhile.

src/pyramid/predicates.py
```diff
@@ -98,33 +98,43 @@ def __call__(self, context, request):
...
+            if ':' in name:
+                name, val_str = name.split(':', 1)
+                try:
+                    v = re.compile(val_str)
+                except re.error as why:
+                    raise ConfigurationError(why.args[0])
+
+            values.append((name, v, val_str))
+
+        self.val = values
```
stevepiercy, line 115: I'm surprised that this was not covered by [Black](https://github.com/Pylons/pyramid/blob/master/HACKING.txt#L51). @mmerickel or @bertjwregeer should this be covered by Black?

mmerickel: I don't understand the question. Why do you think it's not covered.

stevepiercy: `'` instead of `"`, or do we not care about that?

bertjwregeer: This repository is specifically configured to not change from `'` to `"`. https://github.com/Pylons/pyramid/blob/master/pyproject.toml#L6

bertjwregeer:
I've kicked the Travis docs job, now that our docs are back online it should pass.

src/pyramid/config/routes.py
```diff
@@ -220,7 +220,8 @@ def add_route(
         header
 
           This argument represents an HTTP header name or a header
-          name/value pair.  If the argument contains a ``:`` (colon),
+          name/value pair, or a sequence of them.
+          If the argument contains a ``:`` (colon),
```
ewjoachim, line 224: Suggestion: 
```suggestion
          If one of the arguments contains a ``:`` (colon),
```
(just to be explicit that it is evaluated for each element, not for every elements at once)

src/pyramid/config/views.py
```diff
@@ -671,7 +671,8 @@ def wrapper(context, request):
         header
 
           This value represents an HTTP header name or a header
-          name/value pair.  If the value contains a ``:`` (colon), it
+          name/value pair, or a sequence of them.
+          If the value contains a ``:`` (colon), it
```
ewjoachim, line 675: (same)

src/pyramid/config/routes.py
merwok: What do you think of this:
```suggestion
          If a value contains a ``:`` (colon),
```
to avoid clunky «if the argument or one of the arguments»

ewjoachim: Perfect 👌

merwok: Ah, that doesn’t work for the second instance below that already uses «value» instead of «parameters».
I think I have to sit back and rewrite the whole four lines here to make it read better.

stevepiercy: "If any ____"?

stevepiercy:
@merwok docs build is succeeding now. When you get the time, would you please polish off the wording?

merwok:
Yes I was subscribed to the docs PR, I can merge master and do the reword soon.

src/pyramid/config/routes.py
merwok: I found that my issue was the repeated switching between «if this is a header name only» and «if the value contains a colon», so I grouped all info about name-only values then all info about name-with-value-regex values to make it read better.

src/pyramid/config/routes.py
```diff
@@ -219,23 +219,26 @@ def add_route(
...
-          value must match the header value.  Whether or not the value
-          represents a header name or a header name/value pair, the
+          value must match the header value.
+
+          Whether or not the strings
+          represent a header name or a header name/value pair, the
           case of the header name is not significant.  If this
           predicate returns ``False``, route matching continues.
+          When the value is an iterable of strings, they must all
+          return ``True`` for this predicate to return ``True``.
```
merwok, line 241: Wasn’t sure to add this here at the end or rather before line 236.

stevepiercy: p.s. I won't revise the similar instance in `views.py`, pending further review.

merwok: views has the exact same text (including the erroneous `route matching continues`!) so one review is fine.

I’m ok to add hyphens to the two main paragraphs, but it seems like overformatting to me for the second instance.

src/pyramid/config/routes.py
```diff
@@ -220,25 +220,23 @@ def add_route(
...
-          the request *and* the regular expression specified as the
-          value must match the header value.
-
-          Whether or not the strings
-          represent a header name or a header name/value pair, the
-          case of the header name is not significant.  If this
-          predicate returns ``False``, route matching continues.
-          When the value is an iterable of strings, they must all
-          return ``True`` for this predicate to return ``True``.
+          header names.  The header names are determined as follow:
```
stevepiercy, line 223:
```suggestion
          header names.  The header names are determined as follows.
```

src/pyramid/config/routes.py
```diff
@@ -220,25 +220,23 @@ def add_route(
         header
 
           This value can be a string or an iterable of strings for HTTP
```
stevepiercy, line 222: Because "value" is used in the docs below, although in a different context, I would like to reserve its use for the explanation to avoid potential confusion.

```suggestion
          This argument can be a string or an iterable of strings for HTTP
```

src/pyramid/config/routes.py
```diff
@@ -220,25 +220,23 @@ def add_route(
...
-          Whether or not the strings
-          represent a header name or a header name/value pair, the
-          case of the header name is not significant.  If this
-          predicate returns ``False``, route matching continues.
-          When the value is an iterable of strings, they must all
-          return ``True`` for this predicate to return ``True``.
+          header names.  The header names are determined as follow:
+
+          - If a string does not contain a ``:`` (colon), it will be
+            considered to be the header name (e.g.  ``If-Modified-Since``).
```
stevepiercy, line 226: I much prefer "For example" because I never pronounce "e.g." when reading or speaking.

Nonetheless, if it is to be used, then an extra space is not needed.

```suggestion
            considered to be the header name (e.g. ``If-Modified-Since``).
```

src/pyramid/config/routes.py
```diff
@@ -220,25 +220,23 @@ def add_route(
...
+            In this case, the header specified by the name must be present
+            in the request for this string to match.  Case is not significant.
+
+          - If a string contains a colon, it will be considered a
+            name/value pair (e.g. ``User-Agent:Mozilla/.*`` or
+            ``Host:localhost``), where the value part is a regular
+            expression.  The header specified by the name must be present
+            in the request *and* the regular expression specified as the
+            value must match the header value.  Case is not significant for
```
stevepiercy, line 235: > the regular expression specified as the
> value must match the header value

The context and meaning of "header value" is ambiguous and confusing relative to the first "value" in this sentence and to the previous sentence. Is it the "value" in the context of a name/value pair or the value of the request's header?

stevepiercy: I see a "thumbs up" but no change. Did you miss editing this one or still thinking how to better word it?

merwok: I missed time!  maybe later today

docs/narr/viewconfig.rst
```diff
@@ -448,24 +448,23 @@ configured view.
...
-  exists *and* the HTTP header matches the value requested.  When the
-  ``headervalue`` contains a ``:`` (colon), it will be considered a name/value
-  pair (e.g., ``User-Agent:Mozilla/.*`` or ``Host:localhost``).  The value
-  portion should be a regular expression.
-
-  Whether or not the value represents a header name or a header name/value
-  pair, the case of the header name is not significant.
+  This param matches one or more HTTP header names or header name/value pairs.
+  If specified, this param must be a string or a sequence of strings,
+  each string being a header name or a ``headername:headervalue`` pair.
```
stevepiercy, line 453: ``headername:headervalue`` pair.

YES!

docs/narr/viewconfig.rst
```diff
@@ -448,24 +448,23 @@ configured view.
...
-  If ``header`` is specified, and possesses a name/value pair (e.g.,
-  ``User-Agent:Mozilla/.*``), the view will only be invoked if the HTTP header
-  exists *and* the HTTP header matches the value requested.  When the
-  ``headervalue`` contains a ``:`` (colon), it will be considered a name/value
-  pair (e.g., ``User-Agent:Mozilla/.*`` or ``Host:localhost``).  The value
-  portion should be a regular expression.
-
-  Whether or not the value represents a header name or a header name/value
-  pair, the case of the header name is not significant.
+  This param matches one or more HTTP header names or header name/value pairs.
```
stevepiercy, line 451: I just realized, I think we should use [RFC 2616 terminology](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2). I wish I had thought of this earlier.

mmerickel: @stevepiercy Thank you for being so thorough - I'm happy if you want to continue improving things but at this point extra suggestions can be saved for a separate PR as this one has already gone through a few iterations of review and is up to the standard set by the previous code that existed. I'm going to merge this now.

mmerickel:
Thank you @merwok !

src/pyramid/config/routes.py
merwok: Oops I missed that one.  @mmerickel if you do a follow-up PR!

merwok:
I wanted approval from @stevepiercy given that I didn’t take his wording suggestions 100% exactly, but all right!  Glad you squash merged.

mmerickel:
@merwok sounds good, in that case sorry about merging it too quickly - that being said I am happy with the feature and the state of the docs even if there's always room for improvement.

merwok:
No worry!

merwok:
Ah I misread the commit log, there was no squash merge but all my messy commits are in history now!  Will rebase more swiftly next times.

PR #3578, opened 2020-05-04
Switch to pytest

bertjwregeer:
This branch switches Pyramid to using pytest instead of nosetest.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,6 +6,7 @@ source =
 [paths]
 source =
     src/pyramid
+    */src/pyramid
     */site-packages/pyramid
 
 [report]
--- a/HACKING.txt
+++ b/HACKING.txt
@@ -8,7 +8,7 @@ Using a Development Checkout
 ----------------------------
 
 You will have to create a development environment to hack on Pyramid, using a
-Pyramid checkout.  We use `tox` to run tests, run test coverage, and build
+Pyramid checkout. We use `tox` to run tests, run test coverage, and build
 documentation.
 
 tox docs: https://tox.readthedocs.io/en/latest/
@@ -25,6 +25,22 @@ tox on PyPI: https://pypi.org/project/tox/
 
   Alternatively, create a writeable fork on GitHub and clone it.
 
+- Make sure that `tox` is installed, either in your path, or locally. Examples
+  below assume that `tox` was installed with:
+
+    $ pip3 install --user tox
+    $ export TOX=$(python3 -c 'import site; print(site.USER_BASE + "/bin")')/tox
+
+  Feel free to use whatever is your favorite method for installation. One
+  popular method uses `pipx` (https://github.com/pipxproject/pipx).
+
+  Before you file a pull request, we recommend that you run your proposed
+  change through `tox`. `tox` will fully validate that all tests work, all
+  supported formats of documentation will build and their doctests pass, and
+  test coverage is 100%, across all supported versions of Python. `tox` will
+  only run builds for Python versions that you have installed and made
+  available to `tox`. Setting up that environment is outside the scope of this
+  document.
 
 Adding Features
 ---------------
@@ -51,58 +67,43 @@ Coding Style
 - Pyramid uses Black (https://pypi.org/project/black/) and isort (https://pypi.org/project/isort/) for code formatting style.
   To run formatters:
 
-    $ tox -e format
+    $ $TOX -e format
 
 
 Running Tests
 -------------
 
-- The `tox.ini` uses `nose` and `coverage`. As such `tox` may be used
+- The `tox.ini` uses `pytest` and `coverage`. As such `tox` may be used
   to run groups of tests or only a specific version of Python. For example, the
-  following command will run tests on Python 3.7 only without coverage:
+  following command will run tests on the same version of Python that `tox` is
+  installed with:
 
-    $ tox -e py37
+    $ $TOX -e py
 
-  This command will run tests on the latest version of Python 3 with coverage.
+  To run `tox` for Python 3.8 explicitly, you may use:
 
-    $ tox -e py36-cover,coverage
+    $ $TOX -e py38
 
-- To run individual tests (i.e., during development), you can use `nosetests`
+- To run individual tests (i.e., during development), you can use `pytest`
   syntax as follows, where `$VENV` is an environment variable set to the path
   to your virtual environment:
 
     # run a single test
-    $ $VENV/bin/nosetests tests.test_module:ClassName.test_mytestname
+    $ $TOX -e py -- --no-cov tests/test_httpexceptions.py::TestHTTPMethodNotAllowed::test_it_with_default_body_tmpl
 
     # run all tests in a class
-    $ $VENV/bin/nosetests tests.test_module:ClassName
+    $ $TOX -e py -- --no-cov tests/test_httpexceptions.py::TestHTTPMethodNotAllowed
 
-  Optionally you can install a nose plugin `nose-selecttests` to run specific
-  tests.
-  https://pypi.org/project/nose-selecttests/
-  For example, use a regular expression with the `-t` parameter to run tests.
-
-    # run a single test
-    $ $VENV/bin/nosetests -t test_mytestname
-
-- The tests can also be run using `pytest`.
-  https://docs.pytest.org/en/latest/
-  This is intended as a convenience for people who prefer `pytest`. Run the
-  tests like so:
-
-    $ $VENV/bin/pip install pytest
-    $ $VENV/bin/pytest --strict pyramid/ tests/
-
-  To run individual tests (i.e., during development), see "pytest usage -
-  Specifying tests / selecting tests":
+- For more information on how to use pytest, please refer to the pytest
+  documentation for selecting tests:
   https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests
 
 
 Test Coverage
 -------------
 
-- The codebase *must* have 100% test statement coverage after each commit.  You
-  can test coverage via `tox -epy3-cover,coverage`.
+- The codebase *must* have 100% test statement coverage after each commit. You
+  can test coverage via `tox -e py38`.
 
 
 Documentation Coverage and Building HTML Documentation
@@ -111,11 +112,11 @@ Documentation Coverage and Building HTML Documentation
 If you fix a bug, and the bug requires an API or behavior modification, all
 documentation in this package which references that API or behavior must be
 changed to reflect the bug fix, ideally in the same commit that fixes the bug
-or adds the feature.  To build and review docs, use the following steps.
+or adds the feature. To build and review docs, use the following steps.
 
 1. In the main Pyramid checkout directory, run `tox -e docs`:
 
-     $ tox -e docs
+     $ $TOX -e docs
 
 2. Open the `.tox/docs/html/index.html` file to see the resulting HTML
    rendering.
@@ -125,6 +126,6 @@ Change Log
 ----------
 
 - Feature additions and bugfixes must be added to the `CHANGES.rst`
-  file in the prevailing style.  Changelog entries should be long and
-  descriptive, not cryptic.  Other developers should be able to know
+  file in the prevailing style. Changelog entries should be long and
+  descriptive, not cryptic. Other developers should be able to know
   what your changelog entry means.
--- a/docs/typographical-conventions.rst
+++ b/docs/typographical-conventions.rst
@@ -98,10 +98,11 @@ ini:
 
 .. code-block:: ini
 
-    [nosetests]
-    match=^test
-    where=pyramid
-    nocapture=1
+    [tool:pytest]
+    python_files = test_*.py
+    testpaths =
+        tests
+    addopts = -W always --cov
 
 Interactive Python:
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,4 +22,4 @@ ignore-bad-ideas =
 python_files = test_*.py
 testpaths =
     tests
-addopts = -W always --cov
+addopts = -W always
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,10 +1,6 @@
 [easy_install]
 zip_ok = false
 
-[nosetests]
-match = ^test
-nocapture = 1
-
 [aliases]
 dev = develop easy_install pyramid[testing]
 docs = develop easy_install pyramid[docs]
@@ -21,3 +17,9 @@ ignore =
 ignore-default-rules = true
 ignore-bad-ideas =
     tests/pkgs/localeapp/**/*.mo
+
+[tool:pytest]
+python_files = test_*.py
+testpaths =
+    tests
+addopts = -W always --cov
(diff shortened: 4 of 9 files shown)


HACKING.txt
```diff
@@ -57,52 +57,28 @@ Coding Style
...
-  This is intended as a convenience for people who prefer `pytest`. Run the
-  tests like so:
-
-    $ $VENV/bin/pip install pytest
-    $ $VENV/bin/pytest --strict pyramid/ tests/
-
-  To run individual tests (i.e., during development), see "pytest usage -
-  Specifying tests / selecting tests":
-  https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests
+    $ tox -e py38 -- --no-cov tests/test_httpexceptions.py::TestHTTPMethodNotAllowed
```
stevepiercy, line 74: I like the link to [pytest usage - Specifying tests / selecting tests](https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests). I think it might help with on-boarding. Maybe have one abstract example and the link for additional invocations?

.coveragerc
```diff
@@ -6,6 +6,7 @@ source =
 [paths]
 source =
     src/pyramid
+    */src/pyramid
```
stevepiercy, line 9: Is there an environment that does not find src files without this line? I get the same result when toggling between un/commenting this line on macOS.

HACKING.txt
```diff
@@ -57,52 +57,28 @@ Coding Style
...
-    $ tox -e py36-cover,coverage
-
-- To run individual tests (i.e., during development), you can use `nosetests`
+- To run individual tests (i.e., during development), you can use `pytest`
   syntax as follows, where `$VENV` is an environment variable set to the path
   to your virtual environment:
 
     # run a single test
-    $ $VENV/bin/nosetests tests.test_module:ClassName.test_mytestname
+    $ tox -e py38 -- --no-cov tests/test_httpexceptions.py::TestHTTPMethodNotAllowed::test_it_with_default_body_tmpl
```
stevepiercy, line 71: I prefer the abstract to the concrete.

If we drop `$VENV`, and to ease on-boarding, we should:
- remove `$VENV` from the previous sentence, and
- suggest a method for non-experts to install tox

```suggestion
    $ tox -e py38 -- --no-cov tests/test_module.py::ClassName::test_my_test_name
```

stevepiercy: @mmerickel what do you suggest for installing tox? I install it into the project's virtual environment, not that it's the best way, but it is easy to manage for on-boarding.

.coveragerc
bertjwregeer: I don't remember why I have this in all of the `.coveragerc` files, it's something that in the past I've found makes coverage report work correctly.

HACKING.txt
bertjwregeer: @mmerickel suggested linking to https://tox.readthedocs.io/en/latest/install.html

I honestly don't think we should be in the business of telling people how to install their tools. In the case that we do have a $VENV it quickly becomes dangerous because if they install pyramid into that venv and run tests, then that venv may leak through tox and cause false positives or errors as things are being imported from the wrong location. This is even worse if the venv is activated by the user.

I hate to say it, but when was the last time we on-boarded anyone?

setup.cfg
```diff
@@ -20,4 +16,10 @@ ignore =
...
 ignore-default-rules = true
 ignore-bad-ideas =
-    tests/pkgs/localeapp/*
+    tests/pkgs/localeapp/**/*.mo
+
+[tool:pytest]
+python_files = test_*.py
+testpaths =
+    tests
+addopts = -W always --cov --cov-report=term-missing
```
mmerickel, line 25: I tested it and I'm pretty sure you can drop the cov-report here because it uses the values from .coveragerc by default, which already have it setup how we want it.

tox.ini
```diff
@@ -2,12 +2,12 @@
...
     py35,py36,py37,py38,pypy3,
-    docs,py36-cover,coverage,
+    docs,coverage,
 
 [testenv]
 commands =
-    cover: coverage run \
-    {envbindir}/nosetests --with-xunit --xunit-file=nosetests-{envname}.xml {posargs:}
+    python --version
+    pytest {posargs:}
```
mmerickel, line 10: I don't know how much we care but those nosetests-...xml files are consumed by the pylonsproject jenkins afaik which I assume is still running. Anyway if we're getting rid of them we do need to followup and see what's going on there.

HACKING.txt
stevepiercy: A link is fine.

Search pull requests and issues for `HACKING.txt` as an indicator of on-boarding people who were referred to it when they attempted to make a PR and hit issues. I know I do it several times a year.

tox.ini
stevepiercy: Didn't @tseaver set that up? I couldn't find Jenkins in the repo or org config, and I can't remember where it is (was?) hosted.

stevepiercy: I found the URL: http://jenkins.pylonsproject.org/job/pyramid/ Linode DNS points to 45.55.247.238, which appears to be dead.

stevepiercy: For context: https://github.com/Pylons/pyramid/pull/2467#discussion-diff-59150297

I think it has been abandoned. I can also remove the DNS A record from Linode.

bertjwregeer: @mmerickel I'll re-add them if there is someone that yells, I've removed them from three or four other projects now and so far I haven't heard anything.

That being said, before your comment today, the last mention of Jenkins was in 2016, and all those links are dead as Steve just mentioned.

setup.cfg
bertjwregeer: Done.

HACKING.txt
bertjwregeer: I've added that link back in.

HACKING.txt
bertjwregeer: I've updated to add installation of `tox` using `pip3 install --user`.

docs/typographical-conventions.rst
```diff
@@ -98,10 +98,11 @@ ini:
...
 
-    [nosetests]
-    match=^test
-    where=pyramid
-    nocapture=1
+    [tool:pytest]
+    python_files = test_*.py
+    testpaths =
+        tests
+    addopts = -W always --cov --cov-report=term-missing
```
mmerickel, line 105: drop --cov-report arg

HACKING.txt
```diff
@@ -25,6 +25,21 @@ tox on PyPI: https://pypi.org/project/tox/
...
+- Make sure that `tox` is installed, either in your path, or locally. Examples
+  below assume that `tox` was installed with:
+
+    $ pip3 install --user tox
+    $ export TOX=$(python3 -c 'import site; print(site.USER_BASE + "/bin")')/tox
+
+  Feel free to use whatever your favourite method for installation is, one
+  popular one is `pipx`: https://github.com/pipxproject/pipx which then may be
+  easily added to your shell's PATH variable so you may execute tox without
+  providing a full path
```
stevepiercy, line 37: Do we really want to get into recommending tools for installing packages? _/me backs away slowly._

I prefer to link to the [PyPA's Guides](https://packaging.python.org/guides/) and call it soup.

tox.ini
```diff
@@ -2,12 +2,12 @@
...
 extras =
     testing
 deps =
```
mmerickel, line 13: the cover qualifier under here is not necessary since you added pytest-cov to the testing extra

HACKING.txt
```diff
@@ -25,6 +25,21 @@ tox on PyPI: https://pypi.org/project/tox/
...
+
+  To fully validate that all tests work under all versions of Python, you will
+  need to have those versions of Python installed and available to `tox`.
```
stevepiercy, line 40:
```suggestion
  Before you file a pull request, we recommend that you run your proposed change through `tox`.
  `tox` will fully validate that all tests work, all supported formats of documentation will build and their doctests pass, and test coverage is 100%, across all supported versions of Python.
  `tox` will only run builds for Python versions that you have installed and made available to `tox`.
```

HACKING.txt
```diff
@@ -51,58 +66,43 @@ Coding Style
 - Pyramid uses Black (https://pypi.org/project/black/) and isort (https://pypi.org/project/isort/) for code formatting style.
   To run formatters:
 
-    $ tox -e format
+    $ $TOX -e format
```
stevepiercy, line 69: I prefer not to use environment variables.

docs/typographical-conventions.rst
bertjwregeer: This is in the docs/typographical conventions. I'll drop it, but it's just an example.

HACKING.txt
```diff
@@ -25,6 +25,21 @@ tox on PyPI: https://pypi.org/project/tox/
 
   Alternatively, create a writeable fork on GitHub and clone it.
 
+- Make sure that `tox` is installed, either in your path, or locally. Examples
```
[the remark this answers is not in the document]
bertjwregeer: That documentation doesn't really tell someone how to install it so that it's in their path... so I'd prefer to not just link there.

HACKING.txt
bertjwregeer: You were the one that argued for having instructions for users.

HACKING.txt
bertjwregeer: `$VENV` used to exist before. It's the easiest way I can make sure that users that are copy/pasting commands will get the right `tox`.

HACKING.txt
stevepiercy: Sure, but not going into details of using pipx or how to invoke tox according to where it is installed (i.e., global, user, virtual environment).

bertjwregeer:
> we can leave coverage on by default but there's several reasons it was off and reserved to specific targets:
> 
> 1. it slows everything down immensely

It's an additional 10 seconds on my laptop... but okay.

> 2. we tend to only care about coverage on specific versions and it's a waste to test them all

¯\_(ツ)_/¯  I strongly believe that if someone is adding changes and runs them on Py 3.7 locally they should see coverage reporting without necessarily having to resort to installing Py 3.8 just to get a coverage report and know if their new changes are adequately covered.

> 3. we have never tested coverage on windows before, and this enables it

This IMHO is a good idea, and code paths that differ between Linux/Windows should probably be marked as such, there is syntax to allow easily switching coverage reporting to on/off depending on the code path/which OS it's running on.

> 4. we've had cases where coverage didn't run properly on pypy in the past and wanted to keep it out of there

¯\_(ツ)_/¯  this works without issues.

> 5. we're not doing anything right now to properly combine the coverage from every target into a total when running `tox -e coverage`, especially not across windows/mac/linux so for that final report we are still picking a specific target we want to check

For right now, yes. But let's say we have a hack for Python 3.6 and 3.7 that is not in >3.8, with coverage enabled across all we make a simple change to the CI builder to build another environment and `coverage combine` picks them up. Without needing to make additional changes to `tox.ini`.

> 
> if you still feel strongly despite all these past, and some current, issues then I guess it's fine but if it were me I'd leave it off by default

I feel strongly about getting this merged. I've disabled coverage reporting expect on a special case for right now. I don't like it, I don't think it's a smart idea, and I think the extra 10 seconds it takes to run the test suite are nothing in the grand scheme of things, but it is what it is.

HACKING.txt
stevepiercy: For total newbs in narrative and tutorial docs, we definitely need `$VENV` with all of its context about a virtual environment. For the audience of HACKING.txt, is `$VENV` the best way? I dunno.

Perhaps a disclaimer like, "For brevity with commands, we assume you know where your `tox` is installed and how to invoke it" placed before the first instance of invoking tox?

tox.ini
stevepiercy: I deleted the A record from Linode, too. The wayback machine has only one snapshot of it from 2017.

HACKING.txt
```diff
@@ -25,6 +25,26 @@ tox on PyPI: https://pypi.org/project/tox/
...
+  providing a full path
+
+  Before you file a pull request, we recommend that you run your proposed
+  change through `tox`. `tox` will fully validate that all tests work, all
+  supported formats of documentation will build and their doctests pass, and
+  test coverage is 100%, across all supported versions of Python.  `tox` will
+  only run builds for Python versions that you have installed and made
+  available to `tox`.
+
+  Setting up that environment is outside the scope of this document.
```
stevepiercy, line 46: Merge two above paragraphs into one.

HACKING.txt
```diff
@@ -25,6 +25,26 @@ tox on PyPI: https://pypi.org/project/tox/
 
   Alternatively, create a writeable fork on GitHub and clone it.
 
+- Make sure that `tox` is installed, either in your path, or locally. Examples
+  below assume that `tox` was installed with:
+
+    $ pip3 install --user tox
+    $ export TOX=$(python3 -c 'import site; print(site.USER_BASE + "/bin")')/tox
+
+  Feel free to use whatever your favourite method for installation is, one
```
stevepiercy, line 34:
```suggestion
  Feel free to use whatever is your favorite method for installation. One
```

HACKING.txt
```diff
@@ -25,6 +25,26 @@ tox on PyPI: https://pypi.org/project/tox/
...
+  popular one is `pipx`: https://github.com/pipxproject/pipx which then may be
```
stevepiercy, line 35:
```suggestion
  popular method uses `pipx` (https://github.com/pipxproject/pipx).
```

HACKING.txt
```diff
@@ -25,6 +25,26 @@ tox on PyPI: https://pypi.org/project/tox/
...
+  easily added to your shell's PATH variable so you may execute tox without
+  providing a full path
```
stevepiercy, line 37: Break run-on sentence into a second sentence, line-wrapping as needed.

```
If you add `pipx` to your PATH, then you may execute `tox` without providing a full path when invoking it.
```

stevepiercy: Ugh, that's not clear. Should be delete "which then may be..." and replace with a complete sentence.

.coveragerc
merwok: I think this is needed to report proper source paths even when tests are run against the package in tox’ testenv’s site-packages.

bertjwregeer:
> I think windows uses `Scripts` for `bin`

That has been an issue throughout the history of this document: https://github.com/Pylons/pyramid/blob/master/HACKING.txt#L75

merwok:
Ah right :confused:  I wanted to be optimistic and ask whether links to external documentation such as https://docs.python.org/3/using/windows.html#configuring-python and https://docs.python.org/3/library/venv.html#creating-virtual-environments wouldn’t be enough, but even so the commands are not identical (`py` launcher recommended for windows, and even on unix there is a recommendation to use `python -m module` to be sure that it’s the same environment as `python`).  So maybe the inelegant environment variable is the most lightweight thing to use for these docs :slightly_frowning_face:

bertjwregeer:
> There was discussion about running coverage across all versions of Python versus just one. Does this sentence need to be revised?

That sentence was from the original HACKING.txt just relocated, and is technically still true, previously that was Python 2.7 + 3.6, now it's just Python 3.8 that's used for coverage reporting, with the assumption that what is covered in 3.8 is also covered in 3.5, 3.6, 3.7

PR #3579, opened 2020-05-04
Backport #3577

stevepiercy:


stevepiercy:
This PR is not the cause of the build failure. I was going to merge, but reckon that y'all might want to take a look at what's going on with lint.

I'll try to fix the broken coverage in `pyramid.compat`.

stevepiercy:
Actually I won't make an attempt to fix compat. When I ran coverage with `xml -i`, there were too many missed statements, and I should talk with someone before I fix something that ain't broke.

stevepiercy:
Lint is now passing. However the coverage report now shows [two lines as misses](https://travis-ci.org/github/Pylons/pyramid/jobs/684153067#L264).

I checked both `src/pyramid/compat.py` and `tests/test_compat.py`, but I have no idea what is the proper way to get the missed statements to get coverage.

bertjwregeer:
The issue is that `virtualenv` has pulled in `configparser` from PyPI, which back ports the new `configparser` from Python 3.x to Python 2.7.

We need to slap a `pragma: no cover` on that. Previously on Python 2.7 `configparser` would fail to import, and we would fall back to importing `ConfigParser`.

stevepiercy:
Thanks for the assist @bertjwregeer !

PR #3580, opened 2020-05-04
Backport #3577

stevepiercy:
Sort of. In Sphinx 3.0.x, lots of things blow up on this branch, so I pinned it to `Sphinx < 3.0.0`. I think it is not worth the effort to fix everything in this branch for the sake of Sphinx 3.0.x, given that Pyramid 2.0 will soon be released and we won't be doing further updates.

Use links to cookiecutter instead of intersphinx.


stevepiercy:
This PR is not the cause of the build failure. See also https://github.com/Pylons/pyramid/pull/3579#issuecomment-623384750

PR #3581, opened 2020-05-06
Fix lint check

merwok:


stevepiercy:
We might need to backport this one to 1.10-branch and 1.9-branch, too.

stevepiercy:
Actually just 1.10-branch. 1.9-branch uses only PEP8 and not the full linting.

bertjwregeer:
That was fixed in my pytest branch too... guess I'll rebase again.

PR #3583, opened 2020-05-11
Update json.rst

DrPeryCox:
view_defaults decorator stays the same, we only add new view_config decorator


stevepiercy:
This revision does not make sense, but the original does. The `view_defaults` in the class are overridden.

```
@view_defaults(renderer='home.pt')
class TutorialViews:
```

stevepiercy:
Closing this PR as nonsensical and no response from @DrPeryCox, who is welcome to create a new PR that does make sense.

PR #3586, opened 2020-05-28
cur_domain is in practice equivalent to '.' + cur_domain and therefor…

ericatkin:
…e negates the effect of wild_domain

### DIFF
--- a/src/pyramid/authentication.py
+++ b/src/pyramid/authentication.py
@@ -932,8 +932,8 @@ class AuthTktCookieHelper(object):
                 domains.append('.' + cur_domain.split('.', 1)[1])
             else:
                 domains.append(None)
-                domains.append(cur_domain)
                 if self.wild_domain:
+                    domains.append(cur_domain)
                     domains.append('.' + cur_domain)
 
         profile = self.cookie_profile(request)
--- a/tests/test_authentication.py
+++ b/tests/test_authentication.py
@@ -1173,28 +1173,6 @@ class TestAuthTktCookieHelper(unittest.TestCase):
         self.assertTrue('; secure' in result[2][1])
         self.assertTrue(result[2][1].startswith('auth_tkt='))
 
-    def test_remember_wild_domain(self):
-        helper = self._makeOne('secret', wild_domain=True)
-        request = self._makeRequest()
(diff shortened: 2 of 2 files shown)


ericatkin:
So, the lint failure is not related to this change. Do I need to do anything about that?

bertjwregeer:
Is this an issue that also exists on `master`? We'd prefer that pull requests are opened against `master` in that case and we as maintainers will do the work to back port the changes as necessary.

ericatkin:
Yes, I believe so. I can work up a PR against master, but it is impossible
for me to test in context because I haven't ported my codebase to the
pyramid 2 semantics. This is so simple though, I think it will be ok.
Eric

mmerickel:
This change is bw-incompat in the cookie format so I think I'm unlikely to want to release it in the 1.10 series. We'll have to decide whether to backport or not after it makes it into master.

ericatkin:
What do you mean by backward incompatible? Do you mean it is important to
preserve the current (broken) behavior? Perhaps I don't understand all the
implications of the wild_domain arg, but it appears to be useless (setting
it to False has no practical effect) without this change and I don't see
value in preserving its uselessness, though that would be bw compatible :)
Eric

bertjwregeer:
You are correct that currently Pyramid's handling of `wild_domain` is broken. However we take backwards compatibility seriously, and if we make a change to this now, we may may unwittingly break peoples deployments using Pyramid that are depending on the current cookie handling behaviour (including its brokenness).

We tend to weigh the potential impacts on making changes to our long support branches that may impact current users.

Once we have the change on master, and we think through all of the ramifications, we will make a decision on the best way to make sure we don't catch people unawares.

ericatkin:
Makes sense, and this is something I really appreciate about the pyramid
folks. I just wanted to make sure I understand. I'll send a PR on master
shortly. Should I close this one?
Thanks,
Eric

bertjwregeer:
Feel free to close/refer to this one when you make your PR against master. You might find that most of this patch applies against `master`, but don't quote me on that!

ericatkin:
See PR #3587

mmerickel:
Thanks for clarifying @bertjwregeer.

PR #3589, opened 2020-05-30
Add warning to docs about venusian's limitation of only scanning for *.py files

mirko:
It appears to be common sense that pre-compiling python files into byte code does not have any non-obvious disadvantages[1].
During my effort of porting pyramid and its dependencies to OpenWrt I also pre-compiled all python packages into byte-code, as the final to-be-flashed OpenWrt images are focused on size and memory footprint.
I however had to realise, that pyramid "just doesn't work" - without throwing any error, exception or alike - only returning 404 for every HTTP request.
Turns out, pyramid uses venusian to scan for its views (and potentially other components), where venusian skips all non-*.py files - including *.pyc/*.pyo files. Hence, no views in my case.
This is documented here[2].

WARNING: This PR is *untested*, as in, I only added text without (re-)rendering the docs.

[1]https://stackoverflow.com/questions/35618159/should-i-generate-pyc-files-when-deploying/35619259
[2]https://docs.pylonsproject.org/projects/venusian/en/latest/#using-venusian

### DIFF
--- a/src/pyramid/view.py
+++ b/src/pyramid/view.py
@@ -196,21 +196,26 @@ class view_config:
         See also :ref:`mapping_views_using_a_decorator_section` for
         details about using :class:`pyramid.view.view_config`.
 
-    .. warning::
+    .. note::
 
-        ``view_config`` will work ONLY on module top level members
-        because of the limitation of ``venusian.Scanner.scan``.
+        Because of the limitation of ``venusian.Scanner.scan``, note that
+        ``view_config`` will work only for the following conditions.
 
-        Also it will ONLY work on views present as plaintext python
-        files (*.py), ignoring all non-*.py ones which also includes
-        *.pyc and *.pyo (python precompiled byte code / optimized) files.
+        -   In Python packages that have an ``__init__.py`` file in their
+            directory.
 
-        Another venusian specific precondition to take into account is that
-        python packages to be found by venusian need to have a __init__.py
-        file in their root directory.
+            .. seealso::
 
-        This is a limitation in venusian documented in:
-        https://docs.pylonsproject.org/projects/venusian/en/latest/#using-venusian
+                See also https://github.com/Pylons/venusian/issues/68
+
+        -   On module top level members.
+        -   On Python source (``.py``) files.
+            Compiled Python files (``.pyc``, ``.pyo``) without a corresponding
+            source file are ignored.
+        .. seealso::
+
+            See also the `Venusian documentation
+            <https://docs.pylonsproject.org/projects/venusian/en/latest/#using-venusian>`_.
     """
 
     venusian = venusian  # for testing injection
--- a/src/pyramid/view.py
+++ b/src/pyramid/view.py
@@ -212,6 +212,7 @@ class view_config:
         -   On Python source (``.py``) files.
             Compiled Python files (``.pyc``, ``.pyo``) without a corresponding
             source file are ignored.
+
         .. seealso::
 
             See also the `Venusian documentation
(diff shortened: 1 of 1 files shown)


src/pyramid/view.py
```diff
@@ -201,6 +201,16 @@ def my_view(context, request):
         ``view_config`` will work ONLY on module top level members
         because of the limitation of ``venusian.Scanner.scan``.
 
+        Also it will ONLY work on views present as plaintext python
+        files (*.py), ignoring all non-*.py ones which also includes
+        *.pyc and *.pyo (python precompiled byte code / optimized) files.
+
+        Another venusian specific precondition to take into account is that
+        python packages to be found by venusian need to have a __init__.py
+        file in their root directory.
```
bertjwregeer, line 210: This is unfortunately an issue with Python packages. Packages that don't contain a `__init__.py` are considered namespace packages, and the default tools to walk a package/module don't actually return namespace packages.

See this issue for further clarification:
https://github.com/Pylons/venusian/issues/68

src/pyramid/view.py
```diff
@@ -201,6 +201,16 @@ def my_view(context, request):
...
+
+        This is a limitation in venusian documented in:
+        https://docs.pylonsproject.org/projects/venusian/en/latest/#using-venusian
```
bertjwregeer, line 213: This paragraph should get moved up underneath the comment about `*.py{c,o}` because the Venusian docs don't currently warn about `__init__.py` being required.

src/pyramid/view.py
```diff
@@ -201,6 +201,16 @@ def my_view(context, request):
         ``view_config`` will work ONLY on module top level members
         because of the limitation of ``venusian.Scanner.scan``.
 
+        Also it will ONLY work on views present as plaintext python
+        files (*.py), ignoring all non-*.py ones which also includes
+        *.pyc and *.pyo (python precompiled byte code / optimized) files.
```
bertjwregeer, line 206: With the change to move files into `__pycache__` and both Venusian and Pyramid going Python 3.x only, this is something we can solve in venusian. There is no reason for venusian to no longer attempt to import .pyc/.pyo files. This was mainly done to protect against the case in Python 2 where a user removed a .py file, but accidentally left a .pyc file behind.

stevepiercy: I'm not clear what you are requesting here for docs. Do you want to include mention of `__pycache__`? Remove .pyc/.pyo mention before it is solved in Venusian? Something else?

stevepiercy:
I was in the middle of rewriting this warning when Bert's review came in. Here's a screenshot of the current iteration.

Google Chrome007

src/pyramid/view.py
bertjwregeer: I'm not requesting anything, I was merely adding commentary to help future people that stumble upon this, and leaving a note to myself to fix this in venusian.

stevepiercy:
Ooops, `tox -e docs` is more strict than my Sphinx config. Fixing.

src/pyramid/view.py
```diff
@@ -196,11 +196,27 @@ def my_view(context, request):
         See also :ref:`mapping_views_using_a_decorator_section` for
         details about using :class:`pyramid.view.view_config`.
 
-    .. warning::
+    .. note::
 
-        ``view_config`` will work ONLY on module top level members
-        because of the limitation of ``venusian.Scanner.scan``.
+        Because of the limitation of ``venusian.Scanner.scan``, note that
```
bertjwregeer, line 201:
```suggestion
        Because of a limitation of ``venusian.Scanner.scan``, note that
```

PR #3594, opened 2020-06-03
Handle non string object passed to aslist method

raverat:


src/pyramid/settings.py
```diff
@@ -21,14 +21,17 @@ def aslist_cronly(value):
 
 
 def aslist(value, flatten=True):
-    """ Return a list of strings, separating the input based on newlines
+    """ Return a list. Separating the input based on newlines
     and, if flatten=True (the default), also split on spaces within
-    each line."""
+    each line when value is a string."""
```
stevepiercy, line 26:
```suggestion
    Return a list, separating the input based on newlines.
    Also if ``flatten`` is ``True`` (the default), and if the line
    is a string, then the line will be split on spaces.
```

Line wrap as needed for docstrings.

raverat:
The new description seems accurate! I had to manually merge as Github did a bad job..

stevepiercy:
Yeah, GitHub's UI for approving suggestions works only on a single line, not multiple lines. They've been flip-flopping on their implementation.

mmerickel:
Thank you @raverat !

PR #3608, opened 2020-08-13
Clarify that the string 'None' should be the value for `samesite`

stevepiercy:
- See https://github.com/Pylons/pyramid/issues/2733#issuecomment-672936565


stevepiercy:
If this is good, then it should be backported to 1.10 and 1.09.

bertjwregeer:
This is good.

bertjwregeer:
Also, validate that it needs to be ported to 1.09, I am not sure when we introduced support for `SameSite='none'`

stevepiercy:
Yup, 1.10 only, not 1.9. Thanks for the review.

PR #3612, opened 2020-09-29
Backport #3608

stevepiercy:


stevepiercy:
Lint fails here due to a recent change in Black defaults. Merging anyway.

Issue #3614, opened 2020-09-29
Black v20.8b1 causes failing builds


stevepiercy:
Black v20.8b1 causes failing builds. It now changes docstring formatting, where it removes a leading space: `""" docstring` => `"""docstring`

Should I create a PR with just Blackification of the code base on both `master` and `1.10-branch`? I would not want to interfere with pending PRs that would require a rebase by the contributor.

mmerickel:
I think it's ok if you want to re-blackify in a separate PR.

stevepiercy:
OK, I'll wait till the next release of isort in a couple of days.

stevepiercy:
The maintainer released a hotfix due to another issue, so I done did a blackify, but there was a minor issue. [I manually fixed it, but that was one of three options](https://github.com/Pylons/pyramid/pull/3616#issuecomment-701167223).

PR #3616, opened 2020-09-30
Blackify for v20.8b1

stevepiercy:
See #3614

### DIFF
--- a/src/pyramid/asset.py
+++ b/src/pyramid/asset.py
@@ -18,9 +18,9 @@ def resolve_asset_spec(spec, pname='__main__'):
 
 
 def asset_spec_from_abspath(abspath, package):
-    """ Try to convert an absolute path to a resource in a package to
+    """Try to convert an absolute path to a resource in a package to
     a resource specification if possible; otherwise return the
-    absolute path.  """
+    absolute path."""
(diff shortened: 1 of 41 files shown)


stevepiercy:
https://github.com/Pylons/pyramid/pull/3616/checks?check_run_id=1185952976#step:5:11

```
src/pyramid/config/actions.py:190:80: E501 line too long (80 > 79 characters)
        """Add an action with the given discriminator, callable and arguments"""
                                                                               ^
```

I'm not sure which item to change in the configuration so that they align with what y'all want.

https://github.com/Pylons/pyramid/blob/6116c6c8be01909724127b055be633c53c4782fc/.flake8#L1-L10

or

https://github.com/Pylons/pyramid/blob/6116c6c8be01909724127b055be633c53c4782fc/pyproject.toml#L4-L5

or manually edit it as I did in my final commit to get it to pass. Fortunately there was only one edge case.

mmerickel:
You should be able to hand-edit the line to wrap it. Black is supposed to leave lines alone that are shorter than the max line length.

stevepiercy:
OK, I ended up manually editing it, so I chose correctly. Thank you!

PR #3648, opened 2021-01-09
More 2.0 docs prep

### LINKED ISSUE #3647: Extra `**kw` in wiki tutorial?
I'm not sure if this is an actual issue, but I'm opening it just in case. PyCharm issued a code inspection warning while I was working through the authentication and authorization steps in the wiki tutorial.

Under [Add ACL support](https://docs.pylonsproject.org/projects/pyramid/en/master/tutorials/wiki2/authorization.html#add-acl-support), there is a `**kw` parameter for both `remember` and `forget`.

```
    def remember(self, request, userid, **kw):
        return self.authtkt.remember(request, userid, **kw)

    def forget(self, request, **kw):
        return self.authtkt.forget(request, **kw)
```

However neither of these methods in the class `AuthTktCookieHelper` support arbitrary kwargs. https://github.com/Pylons/pyramid/blob/master/src/pyramid/authentication.py#L1116

I'm not absolutely sure, but I think we want to remove `**kw` from these methods. Please let me know.

stevepiercy:
Closes #3647.

See https://github.com/Pylons/pyramid-cookiecutter-starter/pull/110 and https://github.com/Pylons/pyramid-cookiecutter-starter/pull/108

Also I forgot I need to work through the rest of the docs, not just the one wiki.


stevepiercy:
I think that covered everything.

PR #3649, opened 2021-01-15
Break reference cycle between request and context.

luhn:
When implementing traversal, it's common (at least for me) to store the request in the resource objects so you can do things like database lookups.

```python
class MyResource:
    def __init__(self, request):
        self.request = request

    def __getitem__(self, key):
        if self.request.db.lookup(key):
            return ChildResource(self.request, key)
        else:
            raise KeyError()
```

I realized this morning that this will cause a reference cycle between `request.context` and `context.request`.  Not the end of the world, obviously, but unnecessary pressure on the GC.

Maybe this is the user's responsibility?  We could document not to store the request in the context, or to use a WeakRef.

The framework can also prevent this by breaking the cycle once the request has been processed, which is what this PR does.  Really it's just a single line (`request.__dict__.pop('context', None)`), but it got a bit more complicated than that because several router tests rely on the context object being in the request.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Unreleased
 ==========
 
-- Break potential reference cycle between `request` and `context`.
+- Break potential reference cycle between ``request`` and ``context``.
   See https://github.com/Pylons/pyramid/pull/3649
 
 2.0b0 (2020-12-15)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+Unreleased
+==========
+
+- Break potential reference cycle between `request` and `context`.
+  See https://github.com/Pylons/pyramid/pull/3649
+
 2.0b0 (2020-12-15)
 ==================
 
--- a/src/pyramid/router.py
+++ b/src/pyramid/router.py
@@ -252,8 +252,12 @@ class Router:
             return response
 
         finally:
-            if request.finished_callbacks:
-                request._process_finished_callbacks()
+            self.finish_request(request)
+
+    def finish_request(self, request):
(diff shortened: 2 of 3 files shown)


src/pyramid/router.py
```diff
@@ -252,8 +252,12 @@ def invoke_request(self, request, _use_tweens=True):
...
 
         finally:
-            if request.finished_callbacks:
-                request._process_finished_callbacks()
+            self.finish_request(request)
+
+    def finish_request(self, request):
+        if request.finished_callbacks:
+            request._process_finished_callbacks()
+        request.__dict__.pop('context', None)
```
mmerickel, line 260: please add a comment about breaking the cycle

mmerickel:
LGTM, do you mind updating the changelog?

luhn:
Done.  I'll also backport to 1.10.

mmerickel (approved):

CHANGES.rst
```diff
@@ -1,3 +1,9 @@
+Unreleased
+==========
+
+- Break potential reference cycle between `request` and `context`.
```
stevepiercy, line 4: reST literals must be double-ticked.

```suggestion
- Break potential reference cycle between ``request`` and ``context``.
```

stevepiercy (changes requested):
Minor reST syntax fix.

luhn:
Anything more that needs to happen on this before merge?

@stevepiercy, I made the requested rst changes.

mmerickel:
It's ready once @stevepiercy releases it.

stevepiercy (approved):
Sorry, I must have missed the GitHub notification of a pushed commit.

PR #3650, opened 2021-01-15
Backport #3649 to 1.10 branch.

luhn:
#3649

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+Unreleased
+==========
+
+- Break potential reference cycle between `request` and `context`.
+  See https://github.com/Pylons/pyramid/pull/3650
+
 .. _changes_1.10.5:
 
 1.10.5 (2020-11-08)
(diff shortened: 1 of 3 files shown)


mmerickel:
@luhn if you get a chance, it'd be nice to deprecate the pyramid.compat module in 1.10-branch to provide users a warning prior to upgrading. Is that a thing you could look into? We want to wholesale deprecate the entire module and tell users to use six or something instead if they need the features.

luhn:
@mmerickel, I should have some time this weekend to do that.

To prevent triggering warnings from Pyramid own use of the compat module, maybe have `_compat` for internal use and `compat` to throw deprecation warnings.  Am I thinking about this right?

At the moment I should go back to work, I fell down a rabbit hole and ended up submitting this PR. 🙃

mmerickel:
Yeah that's a great question, I hadn't thought about that issue. That sounds like a good approach. Alternatively we don't have to do it - it just bit one user and some of the other removals I have tried to deprecate in 1.10.

PR #3652, opened 2021-01-17
Deprecate pyramid.compat

luhn:
I moved `pyramid.compat` to `pyramid._compat` and updated all internal references.  The new `pyramid.compat` imports all of `pyramid._compat` and raises a deprecation warning.  This way Pyramid won't trip the warning internally, but any external use will.

Generally `from x import *` is frowned upon, but I figure it's okay in this case because a) it won't exist in 2.0 and beyond and b) it guarantees that we don't miss anything.


stevepiercy (approved):
LGTM

mmerickel:
Thank you @luhn

PR #3656, opened 2021-02-17
Pass documentation around.

### LINKED ISSUE #3655: Sphinx doesn't like the reify decorator
> Failed to get a method signature for kinto.core.resource.Resource.timestamp: <pyramid.decorator.reify object at 0x7f175ae45640> is not a callable object

If I understand correctly this is because the reify decorator doesn't pass along the function signature.

Natim:
Fixes #3655 

Use [wraps](https://github.com/python/cpython/blob/master/Lib/functools.py#L65) which also link the documentation.

### DIFF
--- a/src/pyramid/decorator.py
+++ b/src/pyramid/decorator.py
@@ -36,6 +36,7 @@ class reify:
     def __init__(self, wrapped):
         self.wrapped = wrapped
         update_wrapper(self, wrapped)
+        self.__doc__ = wrapped.__doc__
 
     def __get__(self, inst, objtype=None):
         if inst is None:


src/pyramid/decorator.py
```diff
@@ -36,6 +36,9 @@ class reify:
     def __init__(self, wrapped):
         self.wrapped = wrapped
         update_wrapper(self, wrapped)
+        if wrapped is not None:
+            doc = wrapped.__doc__
+        self.__doc__ = doc
```
Natim, line 41:
```suggestion
        self.__doc__ = wrapped.__doc__
```

luhn:
I double checked the docs and I don't see anything wrong with our usage of `update_wrapper`.  `wraps` is meant to be a decorator, and since it's not being used as one here I think it's a no-op.

Natim:
@luhn update_wrapper is meant to be a decorator too: https://github.com/python/cpython/blob/master/Lib/functools.py#L25-L78

The difference between the two, is the list of information that are passed around.

In the @property code, they pass around `__doc__` which is the only missing part to please Sphinx.

I found it better to use wraps that does everything needed but I can keep update_wrapper and just define `__doc__` if you find it better.

src/pyramid/decorator.py
```diff
@@ -36,6 +36,7 @@ class reify:
     def __init__(self, wrapped):
         self.wrapped = wrapped
         update_wrapper(self, wrapped)
+        self.__doc__ = wrapped.__doc__
```
Natim, line 39: update_wrapper is supposed to do it :confused:

mmerickel: It's pretty clear, as you said, that update_wrapper is supposed to do it and is also more defensive about doing it than this code (i.e. if `wrapped` doesn't have a doc attr). I'd want to know why the expected version isn't working first if you could step through update_wrapper and determine why it's not setting this itself.

Natim: Maybe that's why, this would set it to `None` if not defined which would Sphinx to find it.

mmerickel: According to this there's no tests for `None`, it's just pulling the value and setting it.

https://github.com/python/cpython/blob/3.9/Lib/functools.py#L50-L56

luhn:
This `__doc__` stuff is a red herring.

The issue is that sphinx uses [isattributedescriptor](https://github.com/sphinx-doc/sphinx/blob/f8878bb0b67df6e56ca95c003a41edd9eb59d79e/sphinx/util/inspect.py#L287) to [determine if a descriptor represents an attribute](https://github.com/sphinx-doc/sphinx/blob/f8878bb0b67df6e56ca95c003a41edd9eb59d79e/sphinx/ext/autodoc/__init__.py#L2393) or not.

```python
import inspect
from pyramid.decorator import reify
from functools import cached_property
from sphinx.util.inspect import isattributedescriptor


class Widget:
    @property
    def prop(self):
        return 'Foo'

    @cached_property
    def cached(self):
        return 'Foo'

    @reify
    def reify(self):
        return 'Foo'


print(isattributedescriptor(Widget.prop))
print(isattributedescriptor(Widget.cached))
print(isattributedescriptor(Widget.reify))
```

Output is:

```
True
True
False
```

So Sphinx doesn't recognize `reify` as an attribute.  I haven't dug into the implementation to figure out what the qualifications for an attribute are.

luhn:
Funny, I think the problem is that we're using `update_wrapper`.  It's making the descriptor look like the method it's wrapping, but that's specifically *not* what we want to do because reify doesn't behave like a method, it behaves like an attribute.

So @Natim, your fix using `wraps` worked because it was a no-op, and getting rid of `update_wrapper` fixes the problem.

Natim:
Thank you for your great analysis. Feel free to update the PR if you want or I will be able to do it in about 10 hours. 🙏

luhn:
Made the change in #3657

bertjwregeer:
Closing this in favor of https://github.com/Pylons/pyramid/pull/3657. Thanks @Natim for the initial report!

PR #3657, opened 2021-02-17
Remove update_wrapper from reify

### LINKED ISSUE #3655: Sphinx doesn't like the reify decorator

luhn:
Reify uses `update_wrapper`, which makes the descriptor look like the wrapped method.  However, this isn't the desired behavior, because reify changes the method to act like the property.

This causes problems with Sphinx autodoc as reported in #3655, because Sphinx sees the reify'd attribute as a method and throws an error when attempting to get a signature.  You can see the problem here:

```python
import inspect
from pyramid.decorator import reify
from functools import cached_property
from sphinx.util.inspect import isattributedescriptor


class Widget:
    @property
    def prop(self):
        return 'Foo'

    @cached_property
    def cached(self):
        return 'Foo'

    @reify
    def reify(self):
        return 'Foo'


print(isattributedescriptor(Widget.prop))
print(isattributedescriptor(Widget.cached))
print(isattributedescriptor(Widget.reify))
```

Output is:

```
True
True
False
```

This fix removes `update_wrapper` from reify and just sets `__doc__` manually.  This is in line with stdlib's implementation of [functools.cached_property](https://github.com/python/cpython/blob/ebe20d9e7eb138c053958bc0a3058d34c6e1a679/Lib/functools.py#L933).

Fixes #3655.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Unreleased
 - Break potential reference cycle between ``request`` and ``context``.
   See https://github.com/Pylons/pyramid/pull/3649
 
+- Remove ``update_wrapper`` from ``pyramid.decorator.reify``.
+  See https://github.com/Pylons/pyramid/pull/3657
+
 2.0b0 (2020-12-15)
 ==================
 
--- a/src/pyramid/decorator.py
+++ b/src/pyramid/decorator.py
@@ -1,6 +1,3 @@
-from functools import update_wrapper
-
-
 class reify:
     """Use as a class method decorator.  It operates almost exactly like the
     Python ``@property`` decorator, but it puts the result of the method it
@@ -35,7 +32,7 @@ class reify:
 
     def __init__(self, wrapped):
         self.wrapped = wrapped
-        update_wrapper(self, wrapped)
+        self.__doc__ = wrapped.__doc__
 
     def __get__(self, inst, objtype=None):
         if inst is None:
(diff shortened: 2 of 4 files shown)


luhn:
So this causes [this test](https://github.com/Pylons/pyramid/blob/d76903c58c5d0dfd8edc9de40b8ad2382d4f5634/tests/test_config/test_views.py#L4298) to fail.

I'm confused what the test is doing.  Why are we reify'ing the `viewdefaults` wrapper?  That seems like a mistake.

mmerickel:
Huh, yeah so I traced that test back to https://github.com/Pylons/pyramid/commit/52fde9aa164979a5c9e547826c811e9b7116ae93. It looks to me like I totally screwed up the test case when porting it to the new location in the codebase.

luhn:
Since it looks like the original intent was to test reify, I've removed the failing test and added some new tests for reify.

However, I don't see any tests dedicated to viewdefaults...  Am I overlooking them or do we need to write some?

luhn:
Should this be backported?  Technically this isn't a bw compat change, but the differences are subtle.

mmerickel:
Not sure why they aren't showing up nicely in github search but they are here: https://github.com/Pylons/pyramid/blob/ee7ca28cc51cf40d1190144834704e287c9fc72d/tests/test_view.py#L917

mmerickel:
I think it's ok to backport if you want to.

PR #3661, opened 2021-03-01
Make sentence general by removing explicit commands.

stevepiercy:
This should be backported to 2.0-branch and 1.10-branch, if approved.


bertjwregeer:
:+1: IMHO.

PR #3671, opened 2021-06-12
add optional scope to `ISession.changed()`

gjo:
I'd like to create an `ISession` implementation that persists session data, split by key.
In the current interface, when `ISession.changed()` is called, all the data will be persisted. If we can get the key that was changed during `ISession.changed()`, we can narrow down the scope of persistence.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -354,3 +354,5 @@ Contributors
 - Sergey Maranchuk, 2020/04/18
 
 - Thibault Ravera, 2020/06/03
+
+- Gouji Ochiai, 2021/06/12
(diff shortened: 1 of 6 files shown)


mmerickel:
I don't see a need to change the ISession contract to handle this. The Session impl should be able to detect which keys changed internally most of the time. Remember a Session impl can invoke any other methods it wants internally when it detects keys change.

This API only needs to be used by app developers consuming the session - and I'm not convinced they need the current `.changed()` api very often, let alone a more granular version of it.

mmerickel:
Considering my comments above and also the fact that you can implement this feature on your session implementation without adding it to the interface at this time I'm going to close this. Apologies for not accepting this but thank you for trying to contribute your work!

PR #3683, opened 2021-12-14
Do not include segment_name in allowed values in doc example.

ianjosephwilson:
This is just a tiny doc change that accidentally placed the name of the segment to match against into the list of allowed matches for a custom route predicate.

I should be in [Contributors.txt](https://github.com/Pylons/pyramid/blob/2.0/CONTRIBUTORS.txt#L184) as 
`Ian Wilson, 2012/06/17`.

### DIFF
--- a/docs/narr/urldispatch.rst
+++ b/docs/narr/urldispatch.rst
@@ -1109,7 +1109,7 @@ See this code:
     class AnyOfPredicate:
         def __init__(self, val, info):
             self.segment_name = val[0]
-            self.allowed = tuple(val[0:])
+            self.allowed = tuple(val[1:])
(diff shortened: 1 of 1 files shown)


stevepiercy:
We have build failures due to recent changes in supported versions of OSes and their Python and PyPy versions in GitHub Actions. I created a new issue for that in https://github.com/Pylons/pyramid/issues/3684.

@merwok as the contributor, would you please review this change? Thank you!

If approved, this should be backported to `2.0-branch`.

merwok (approved):
Good catch!

stevepiercy:
Thank you!

PR #3689, opened 2022-01-07
Add support for Python 3.10

ergo:

PR #3690, opened 2022-01-07
Add version constraints to black, flake8, and isort

luhn:
#3689 got me thinking that formatting/linting should be stable, PRs shouldn't be required to reformat the entire project to get the tests to pass.  This PR adds version constraints to the black, flake8, and isort to produce stable results, according their release policies:

- flake8:  Stable within a minor release.  https://flake8.pycqa.org/en/latest/faq.html#why-does-flake8-use-ranges-for-its-dependencies
- black:  Beta, no stability guarantees.  https://black.readthedocs.io/en/stable/
- isort:  Stable within a major release.  https://pycqa.github.io/isort/docs/major_releases/release_policy.html

### DIFF
--- a/HACKING.txt
+++ b/HACKING.txt
@@ -69,6 +69,10 @@ Coding Style
 
     $ $TOX -e format
 
+- Pyramid uses flake8 (https://pypi.org/project/flake8/) to enforce PEP8 style guidelines.
+  To run flake8, as well as running checks for Black and isort:
+
+    $ $TOX -e lint
 
 Running Tests
 -------------
--- a/HACKING.txt
+++ b/HACKING.txt
@@ -74,6 +74,8 @@ Coding Style
 
     $ $TOX -e lint
 
+Black, isort, and flake8 versions are pinned for stability and reproducibility.
+
 Running Tests
 -------------
 
--- a/tox.ini
+++ b/tox.ini
@@ -29,7 +29,7 @@ commands =
     check-manifest
 deps =
     flake8~=4.0.1
-    black==21.12b0
+    black~=22.1.0
     isort~=5.10
     readme_renderer
     check-manifest
@@ -66,7 +66,7 @@ commands =
     isort src/pyramid tests setup.py
     black src/pyramid tests setup.py
 deps =
-    black==21.12b0
+    black~=22.1.0
     isort~=5.10
 
 [testenv:build]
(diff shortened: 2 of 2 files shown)


stevepiercy (approved):
This is very sensible. I would suggest adding a comment in [HACKING.txt](https://github.com/Pylons/pyramid/blob/0b24ac16cc04746b25cf460f1497c157f6d3d6f4/HACKING.txt#L67).

```
- Pyramid uses Black (https://pypi.org/project/black/), flake8 (https://pypi.org/project/flake8/), and isort (https://pypi.org/project/isort/) for code formatting style.
  We pin those versions for stability and reproducibility.
```

I would also suggest an addition to [RELEASING.txt](https://github.com/Pylons/pyramid/blob/0b24ac16cc04746b25cf460f1497c157f6d3d6f4/RELEASING.txt#L69-L71)

```
- Bump the version pins in ``tox.ini`` for Black, flake8, and isort to the latest.

- Make sure PyPI long description renders::

  $ tox -e lint
```

Thank you!

luhn:
I'm not sure I agree with bumping versions on release.  Bumping versions might cause any number of source changes, I don't think we should be messing with the codebase immediately prior to releasing.

stevepiercy:
Yeah, I'm not sure how we can set a reminder to check whether to check to do a version bump of tools. Usually that happens only when builds fail, but if we pin these tools, then the builds will not fail due to changes in these tools. How and when should we remember to check for the latest versions?

luhn:
I made the version constraints as loose as possible, so to a limited degree updates will be automatic.

I also think it's okay if we're a bit out of date.  It doesn't matter too much if we're using the latest and greatest formatting tools, as long as everybody is on the same version and getting the same results.

If you do really think we need a policy for updating, maybe we update the version for alpha and beta builds only?

mmerickel:
I don't think we need a policy on this. It's just formatting/linting and can be run whenever. This LGTM.

mmerickel (approved):

luhn:
Black is now stable!  Updated PR accordingly.

bertjwregeer:
I just merged a different PR that changes the versions we support, but also fixes the pypy issue, if you could merge master or rebase on top of master, so CI can run that would be great!

luhn:
Alright, rebased.

bertjwregeer:
Thank you!

PR #3696, opened 2022-03-13
get rid of deprecated getargspec and formatargspec

### LINKED ISSUE #3688: getargspec and formatspec were removed in Python 3.11
https://github.com/Pylons/pyramid/blob/0b24ac16cc04746b25cf460f1497c157f6d3d6f4/docs/conf.py#L395-L398

https://docs.python.org/3.11/whatsnew/3.11.html#removed

> the getargspec function, deprecated since Python 3.0; use inspect.signature() or inspect.getfullargspec() instead.
> the formatargspec function, deprecated since Python 3.5; use the inspect.signature() function and Signature object directly.

mmerickel:
fixes #3688

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,7 +14,6 @@
 import sys
 import os
 import datetime
-import inspect
 import warnings
 
 warnings.simplefilter('ignore', DeprecationWarning)
@@ -385,19 +384,7 @@ def setup(app):
     app.add_directive('frontmatter', FrontMatter)
     app.add_directive('mainmatter', MainMatter)
     app.add_directive('backmatter', BackMatter)
-    app.connect('autodoc-process-signature', resig)
 
-
-def resig(app, what, name, obj, options, signature, return_annotation):
-    """ Allow for preservation of ``@action_method`` decorated methods
-    in configurator """
-    docobj = getattr(obj, '__docobj__', None)
-    if docobj is not None:
-        params = list(inspect.signature(docobj).parameters.values())
-        if params and params[0].name in ('cls', 'self'):
-            del params[0]
-        signature = str(inspect.Signature(params))
-    return signature, return_annotation
 
 # turn off all line numbers in latex formatting
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -393,10 +393,10 @@ def resig(app, what, name, obj, options, signature, return_annotation):
     in configurator """
     docobj = getattr(obj, '__docobj__', None)
     if docobj is not None:
-        argspec = inspect.getargspec(docobj)
-        if argspec[0] and argspec[0][0] in ('cls', 'self'):
-            del argspec[0][0]
-        signature = inspect.formatargspec(*argspec)
+        params = list(inspect.signature(docobj).parameters.values())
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -393,10 +393,10 @@ def resig(app, what, name, obj, options, signature, return_annotation):
...
     docobj = getattr(obj, '__docobj__', None)
     if docobj is not None:
-        argspec = inspect.getargspec(docobj)
-        if argspec[0] and argspec[0][0] in ('cls', 'self'):
-            del argspec[0][0]
-        signature = inspect.formatargspec(*argspec)
+        params = list(inspect.signature(docobj).parameters.values())
+        if params and params[0].name in ('cls', 'self'):
+            del params[0]
+        signature = str(inspect.Signature(params))
```
stevepiercy, line 399: Is this actually used or displayed in the docs? I've always mentally "bleeped" over it when updating docs. So I tried to make sense of what this actually does, but I just don't grok it. Next I tried looking at any decorated method, such as https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.add_route but I don't see anything related to `action_method`.

mmerickel:
So all I did was confirm that this change operated the same as before. So it's acceptable from that perspective (I rendered old docs, and new docs and did a diff and they were identical).

That being said, I've now gone and compared the actual output with and without the `resig` hook.

**with resig**

Screen Shot 2022-03-13 at 11 59 50

**without resig**

Screen Shot 2022-03-13 at 12 00 47

So, it's easy to say right now that `resig` is making things worse and the the question is why it was added in the first place. Maybe it's just not necessary anymore. It was originally changed in https://github.com/Pylons/pyramid/commit/be5f12ca5624130e6a61297d4c944cc65f757f52 and I can't currently find a compelling reason to keep it.

It's hard to tell what version of sphinx was used to build each docs, but at least going back to 1.3-branch in RTD you can see the signature looks bad - but we don't pin sphinx very heavily so that is likely a newer version of sphinx than what the fix was created for. https://docs.pylonsproject.org/projects/pyramid/en/1.3-branch/api/config.html#pyramid.config.Configurator.add_static_view

mmerickel:
add_static_view may be a red herring as it has an overridden signature in the docs ... meaning all resig is doing is overriding that override for add_static_view. Maybe for other methods it's doing something helpful but it's not clear yet. I can say that for methods with `@action_method` the `obj` and `docobj` objects are the same which means sphinx is properly finding the original method and treating `@action_method` as a decorator that is ignored (unclear if that's good or bad).

**actual signature**

Screen Shot 2022-03-13 at 12 12 09

**docs override**

Screen Shot 2022-03-13 at 12 12 30

mmerickel:
After poking around for a bit the only thing I see `resig` actually changing from sphinx's default signature is that it's overriding our custom signatures we've put into the docs. I think we can get rid of it.

mmerickel:
I have updated the PR to get rid of `resig`. I put in code to output anytime it has an effect and it holds up to my analysis above that it never changes anything except for the worse (overriding custom signatures in the docs).

stevepiercy:
Thank you for doing the detective work. This LGTM.

PR #3702, opened 2022-04-10
Move notify for NewResponse for symmetry with other event exception h…

ericatkin:
…andling

### DIFF
--- a/src/pyramid/router.py
+++ b/src/pyramid/router.py
@@ -168,6 +168,8 @@ class Router:
                 msg = request.path_info
             raise HTTPNotFound(msg)
 
+        has_listeners and notify(NewResponse(request, response))
+
         return response
 
     def invoke_subrequest(self, request, use_tweens=False):
@@ -246,8 +248,6 @@ class Router:
 
             if request.response_callbacks:
                 request._process_response_callbacks(response)
-
-            has_listeners and notify(NewResponse(request, response))
 
             return response
 
--- a/src/pyramid/router.py
+++ b/src/pyramid/router.py
@@ -234,10 +234,6 @@ class Router:
         return the generated response.
 
         """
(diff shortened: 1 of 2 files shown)


bertjwregeer:
@ericatkin could you provide some more context/colour around this change? As it stands now I am not sure what exactly this is changing, or what bug this is fixing.

How does that affect this diagram: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/router.html?

ericatkin:
This moves the NewResponse event to the Router.handle_request() method so that exceptions in such events are handled in the same way as NewRequest, BeforeTraversal, and ContextFound (by tweens). As it is, I don't see any way to catch exceptions in NewResponse handlers.

With regard to the diagram, it would move NewResonse above the egress tween node. I don't know how to update that.

huntcsg:
The existing behavior is part of the documented public API. See the docs for [pyramid.request.Request.add_response_callback](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.add_response_callback):

> All response callbacks are called after the tweens and before the [pyramid.events.NewResponse](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/events.html#pyramid.events.NewResponse) event is sent.

Instead of making this change would you be able to register a tween at the correct position to accomplish whatever you are attempting to do in your `NewResponse` listener?

If that is impractical would it make sense to add an additional event/events that you could listen for instead of `NewResponse`?

ericatkin:
Yeah, now that I understand tweens, a tween under EXCVIEW works for me, but I found it surprising that NewResponse happens outside of the EXCVIEW tween as opposed to the other events. I imagine that would break a lot of code to move it as I proposed.
Thanks.

PR #3704, opened 2022-06-07
Change != to is not in pshell initialization

ztane:
I had a curious problem: I had thrown an object that was wrapped in lazy_object_proxy added into the `pshell` environment. It had worked before but for some reason stopped working here merely due to updates in either Python version or some of the dependent libraries. The error thrown was `AttributeError: __wrapped__` on this line. As `is not` is likely the more correct solution here, here's the pull request.


russellballestrini (approved):
looks good to me but i'm just a community member.

mmerickel:
thanks @ztane!

Issue #3710, opened 2022-09-29
pserve out function always prints to stderr


markonose:
In https://github.com/Pylons/pyramid/blob/master/src/pyramid/scripts/pserve.py#L146

The script outputs everything to the stderr file descriptor.  If you redirect the stdout and stderr stream to your logger, then for example the 'Starting server' message gets logged as an error.

I would propose to split the out function into two in PServeCommand:

```
    def out(self, msg):  # pragma: no cover
        if self.args.verbose > 0:
            print(msg)

    def out_err(self, msg):  # pragma: no cover
        if self.args.verbose > 0:
            print(msg, file=sys.stderr)
```

And call the out_err function to print 'You must give a config file'.

In the rest of the file all of the other prints could also be changed to stdout, so that hey would no longer output starting/serving messages to stderr.

In other scripts even errors get printed to stdout, so this could also be applied in other scripts where applicable: https://github.com/Pylons/pyramid/blob/master/src/pyramid/scripts/prequest.py#L135

I can create a PR if such a changes would be permitted.

mmerickel:
It's standard practice that logging systems log to stderr and leave stdout for core functionality. In the case of pserve everything is logging, leaving the core functionality to the wsgi server itself. I'm not sure this is really incorrect.

I would argue that prequest should be outputting several of those messages to stderr and could use some updates there.

markonose:
Sorry my bad did not realize that it was standard practice to output diagnostic data to stderr.

The idea behind my question was because we use pyramid with pserve, and in our app we redirect the stderr to our logger, in case something goes wrong and the config_uri argument does not get passed to the command so that it doesn't fail silently, but gets printed by our logger using our format. But this has an undesired side effect, where we also see the message 'Starting server in PID %i.'. But seeing as this is clearly diagnostic output, this is the common way a command acts and we have to work around it.

Regarding prequest the error output should then be written to stderr and not stdout to keep things consistent. I would be happy to go through all of the commands and change the outputs to the correct stream if needed.

mmerickel:
I think that'd be great if you wanted to do that!

PR #3711, opened 2022-10-02
Change error output in scripts to appropriate file descriptors

markonose:
https://github.com/Pylons/pyramid/issues/3710

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -358,3 +358,5 @@ Contributors
 - Jens Troeger, 2021/04/08
 
 - Karthikeyan Singaravelan, 2021/08/24
+
+- Marko Nose, 2022/09/30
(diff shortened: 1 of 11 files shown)


mmerickel (changes requested):
I like all of the locations you have changed to use stderr. I'd like to request a stylistic change which would have made it a bit easier to mock, which is instead of coupling the code to `sys.stderr`, rather either 1) add a `self.err()` used instead of `self.out()` for errors or 2) add an `error=True` arg to `self.out()`. This allows hiding the logic about where the messages are going from the call sites.

markonose:
Sorry been really busy.

Am much more in favor of the first approach.

Will update the PR over the weekend.

mmerickel:
I'm closing this due to inactivity. We can always revisit.

PR #3728, opened 2023-05-11
Update conf.py

skywalk163:
Change URL from '-' to '_' except for https://docs.pylonsproject.org/projects/pylons-webframework/en/latest/. This is because all other URLs have replaced “-” between words in their names with underscore symbols.


stevepiercy (approved):
Thank you!

PR #3729, opened 2023-08-09
2.0 branch 1

leocosta-io:
There is a small mistake on the section "Edit scripts/initialize_db.py":

We'll also change **_the the_** script to create two User objects (basic and editor) as well as a Page, rather than a MyModel, and add them to our dbsession.


mmerickel:
Please make the change on the main branch instead of 2.0-branch. I believe you made the change on the 2.0-branch and have submitted the PR targetting main here which is not what we want.

mmerickel:
replaced with #3730

Issue #3734, opened 2023-10-18
Adding script info to global_config during application setup


adroullier:
Pyramid startup scripts in python/bin directory (pserve, prequest, pshell) do not pass any special info to the wsgi app or "main()" function. As far as I know there is no way to find out how the wsgi application startup is triggered. 

In some cases it makes sense to call different code whether the application is created as server (bin/pserve) or for a single request (bin/prequest). For example setting up internal caching modules, maintenance or other background functions.

Right now you have use extra environment variables, multiple wsgi configurations or other workarounds which are not really elegant solutions.

My proposal is to extend the scripts with additional context information which simply refer to the scripts name. The additional info is passed to the main function as part of `global_config["__script__"]` dict during wsgi setup. 

For example, from a developers perspective it would look like:

```
from pyramid.config import Configurator

def main(global_config, **settings):
    config = Configurator()
    config.include('myapp.myconfig.includeme')
    
    # global_config.get("__script__") = either "prequest", "pserve", "pshell" or None 
    # signals how the application startup was triggered
    # if required handle application setup accordingly

```

I can provide a pull request if there a no objections. And the code change should have no impact on existings applications.

mmerickel:
I don't currently have a good vision on how this would be implemented - if it can be done in pyramid's shell scripts directly, and/or if plaster/plaster-pastedeploy would need to assist in some way but I'm happy to consider a PR to add this idea of marshaling a value through into the global_config.

mmerickel:
Fixed via https://github.com/Pylons/pyramid/pull/3735.

PR #3735, opened 2023-10-24
Adding script name to 'global_config' during application setup 

adroullier:
This code extends pyramid startup scripts in python/bin directory (pserve, prequest, pshell, pviews, proutes, ptweens) with additional info about the calling script.  The additional info is passed to the **main(global_config, **settings)** function as part of global_config dict during wsgi setup as '__script__'.
``global_config`` can contain custom values loaded from the ini file, so to avoid unwanted overwriting of a existing '__script__' value I've added a extra condition for comapatibility. 

In some cases it makes sense to call different code whether the application is created as server (bin/pserve) or for a single request (bin/prequest). For example setting up internal caching modules, maintenance or other background functions.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,17 @@ Features
 
 - Coverage reports in tests based on Python 3.11 instead of Python 3.8.
 
+- All scripts now pass a new option ``__script__`` when loading the WSGI app.
+  For example, ``pserve`` sets ``__script__ == 'pserve'``. This works for
+  ``pserve``, ``pshell``, ``prequest``, ``proutes``, ``ptweens``, ``pviews``,
+  as well as when using ``pyramid.paster.bootstrap`` directly.
+
+  When using ``plaster-pastedeploy`` to load an INI file, this option will
+  manifest as a new value passed into the ``global_conf`` arg of your
+  application factory, where you can use it as part of initializing your app.
+
+  See https://github.com/Pylons/pyramid/pull/3735
+
 Bug Fixes
 ---------
 
--- a/docs/narr/startup.rst
+++ b/docs/narr/startup.rst
@@ -101,7 +101,9 @@ Here's a high-level time-ordered overview of what happens when you press
    Note that the constructor function accepts a ``global_config`` argument,
    which is a dictionary of key/value pairs mentioned in the ``[DEFAULT]``
    section of an ``.ini`` file (if :ref:`[DEFAULT]
-   <defaults_section_of_pastedeploy_file>` is present).  It also accepts a
+   <defaults_section_of_pastedeploy_file>` is present) and the executing
+   script name ``__script__`` like ``pserve``, ``prequest`` or ``pshell``.
+   It also accepts a
    ``**settings`` argument, which collects another set of arbitrary key/value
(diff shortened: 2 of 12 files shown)


mmerickel (changes requested):
1. could you please tweak these to use `config_vars.setdefault('__script__', 'pshell')` as a simpler one-liner?
2. Is there a way to make this work with `pyramid.paster.bootstrap` as well, or only the scripts? I believe this use case is not addressed here and was the topic of the recent mailing list posts triggering this work.

adroullier:
1: no problem. 

2: ``pyramid.paster.bootstrap`` is called in the scripts ``proutes, pshell, ptweens, pviews``. In this case the script name is passed to ``bootstrap()`` as part of the ``options`` parameter (which is then passed to the callback get_app()). If ``bootstrap()`` gets called from other parts of the code there is no such info.
You are right, it makes sense to add info to ``boostrap()``, too. Maybe ``'__script__'='bootstrap'`` could be used for all these scripts (proutes, pshell, ptweens, pviews). It is more consistent.

mmerickel:
It could be the default value going back to your logic that won’t set it off already defined. Then individual scripts can still override it as you’re doing?

adroullier:
The scripts (except pserve, prequest) call ``pyramid.paster.bootstrap``, so the script name is set before bootstrap() is called and shouldn't be a problem.

I had a look at the email discussion and pyramid code again. The original question was if there is way to have more infos in the context of the ``ApplicationCreated`` event. The event is triggered in ``pyramid.config.Configurator.make_wsgi_app()``. The PR does not change this behaviour and  I don't think there is a way to change the event.

mmerickel:
The user would need to forward the setting from global _config into their app but at least it’s an automatic setting versus one they have to set/maintain in their config files.

mmerickel (changes requested):
What's the status of this from your perspective @adroullier? Trying to understand if you think this is done or if there's still a plan to handle the bootstrap case. I still think we should define `__script__` if bootstrap is used directly, not just via these p* scripts.

Rest of the PR looks great and I'd be happy to merge it.

adroullier:
Sorry I haven't mentioned it in my last comment. 
Actually I haven't found a use case or point where custom pyramid code is called after calling ``pyramid.paster.bootstrap()`` (and I think that is what you are referring to?). 
If ``bootstrap()``is called by custom code the developer knows the environment anyway before calling ``bootstrap()``. So adding the script name there means just adding a few lines of unnecessary code.
Or is there a mistake in my argumentation?

I can also add the code to set the name to 'bootstrap', like

``pyramid.paster.bootstrap()``  line 127
```
def bootstrap(config_uri, request=None, options=None):
    # ...
    if options is None:
        options = dict()
    options.setdefault('__script__', 'bootstrap')
    app = get_app(config_uri, options=options)
    env = prepare(request)
    env['app'] = app
    return env
```

mmerickel:
Well the goal is to still add that the app can differentiate the scenario from a normal startup using that `__script__` without relying on every user setting it themselves prior to calling `bootstrap`. I think it's worth adding that in the spirit of this PR.

mmerickel (approved):

mmerickel:
Fixes #3734.

PR #3737, opened 2023-11-29
Add 3.12 to CI, tox and metadata; use latest black and flake8

Jmennius:


src/pyramid/config/adapters.py
```diff
@@ -80,7 +80,7 @@ def _derive_predicate(self, predicate):
 
         if eventonly(predicate):
 
-            def derived_predicate(*arg):
+            def derived_predicate(*arg):  # noqa: F811
```
luhn, line 83: Probably better to fix the lint error rather than ignoring it.  Maybe:

```python
if not eventonly(predicate):
    return predicate

def derived_predicate(*arg):
     return predicate(arg[0])

return derived_predicate
```

Jmennius: Definitely, went for `else` clause.

stevepiercy (changes requested):
In addition to code and CI, would you please scan the documentation for Python support? I suspect searching for `python 3` or `3.11` will turn up a few instances.

stevepiercy (approved):
Thank you! I approve the docs stuff only.

Jmennius:
> In addition to code and CI, would you please scan the documentation for Python support? I suspect searching for `python 3` or `3.11` will turn up a few instances.

Done.
At this point, I would think that we may need a release notes/changes update as well? But I'm a bit confused about that because of difference in `2.0-branch` and `main`.

Should we drop Python 3.7?

PR #3750, opened 2024-02-07
use AssetResolver internally

mmerickel:
- abspath_from_asset_spec no longer uses pkg_resources directly
- ManifestCacheBuster now uses AssetResolver instead of pkg_resources directly
- rewrite static_view to use asset resolver

Changing the static_view is bw-incompat, and I'm not there yet with it being a good idea. I think it is - but hey.

The issue with AssetResolver in general is that it tries to import the package. This is bw-incompat. On previous versions, you could `add_static_view('does_not_exist:', 'static')` and it would work. You could then `config.override_asset(to_override='does_not_exist:', override_with='something_that_exists:')`. This change requires the initial `does_not_exist` package to be importable no matter what.

### DIFF
--- a/src/pyramid/asset.py
+++ b/src/pyramid/asset.py
@@ -1,7 +1,6 @@
 import os
-import pkg_resources
 
-from pyramid.path import package_name, package_path
+from pyramid.path import AssetResolver, package_name, package_path
 
 
 def resolve_asset_spec(spec, pname='__main__'):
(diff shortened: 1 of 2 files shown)


sbrunner:
The currently used setuptools version became to have some CVE is it possible to merge this pull request and create a new release?

robvdl:
The setuptools<82 thing was always a temporary workaround that was going to
blow up at some point, it was only a matter of time. I think we do need a
new release as well, for setuptools 83.

This is regarding CVE-2026-59890 I suppose?

sbrunner:
I completely agree!

Currently, this is the CVE, but there will inevitably be more in the future.

Issue #3758, opened 2024-04-06
Modify `route_prefix_context` to preserve leading slash


fmigneault:
## Feature Request

**Is your feature request related to an issue? Please describe.**

When `route_prefix_context` is used to apply a prefix on various routes, the supplied `route_prefix` gets stripped to remove any leading/training slashes. Provided that Pyramid handles `/prefix/route` and `prefix/route` the same way, it should not strip the leading slash (it is fine to strip the training one to allow addition of more routes after). 

This was highlighted a while back in https://github.com/Pylons/pyramid/issues/2143, but not addressed further. 

When working with utilities like https://github.com/Cornices/cornice and https://github.com/Cornices/cornice.ext.swagger, it is possible to define a `Service` that uses the `prefix_route` to obtain the correct path for the OpenAPI rendering. Because `route_prefix_context` strips the path, it is not generating the intended path.

**Describe the solution you'd like**

Modify this line:
https://github.com/Pylons/pyramid/blob/72f61853beda8e21b669c3520e43fe3e5b224ba3/src/pyramid/config/routes.py#L602
```diff
- route_prefix = route_prefix.strip('/')
+ route_prefix = route_prefix.rstrip('/')
```

When using `rstrip` and `with config.route_prefix_context("/ogcapi")`, the routes below get rendered correctly.
They behave and respond exactly the same way as if the path were stripped from the leading `/`.

![image](https://github.com/Pylons/pyramid/assets/19194484/1f76abea-9b71-49d1-807f-9f16ca47e713)

With the current `strip`, all leading `/` are lost, creating inconsistent paths:

![image](https://github.com/Pylons/pyramid/assets/19194484/e1cf5066-3f05-4b2c-aca0-43ceb8ecfa66)

**Describe alternatives you've considered**

Override the `pyramind.config.Configurator.route_prefix_context` method.
While this works, it is a dirty workaround as I have to duplicate the code to keep all the other `self.begin()`, etc. calls. 

**Additional context**
This is where the supplied `route_prefix` is relevant in the package that uses it for eventual OpenAPI generation : 
https://github.com/Cornices/cornice/blob/abfcfd9eeefaf281237ba84a3b90ef5ee0698aa9/src/cornice/pyramidhook.py#L169-L183

mmerickel:
[...]

PR #3759, opened 2024-04-12
fix `route_prefix` leading slash handling

fmigneault:
Fixes https://github.com/Pylons/pyramid/issues/3758

In summary, applies the proposed `strip` -> `rstrip` modification to `route_prefix` handling such that the resulting `route_prefix` in the configurator keeps the leading `/` if explicitly specified. If not specified (i.e.: `route_prefix="prefix"` instead of  `route_prefix="/prefix"`), it results in the same behavior as before.

When merging/nesting multiple paths/routes onto each other, insertion of the `/` is applied automatically as needed (as it was before, but considering `/prefix` must be kept instead of `prefix`). 

Added exhaustive test combinations for combining `config.include(route_prefix=...)`, `with config.route_prefix_context(route_prefix=...)` with direct/nested configurators and with/without explicit leading slash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ env*/
 venv/
 .cache/
 .python-version
+pyramid-*/
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,9 @@ Features
 Bug Fixes
 ---------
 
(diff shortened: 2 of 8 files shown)


fmigneault:
@mmerickel FYI

Issue #3761, opened 2024-06-15
`test_util` fails with Python 3.13


mattiaverga:
Two tests in `test_util` started to fail in Fedora Linux after the move to Python 3.13.0b2:
```
============================= test session starts ==============================
platform linux -- Python 3.13.0b2, pytest-7.4.3, pluggy-1.3.0
rootdir: /builddir/build/BUILD/python-pyramid-2.0.2-build/pyramid-2.0.2
configfile: setup.cfg
plugins: cov-4.0.0
collected 2635 items
tests/test_asset.py ...........                                          [  0%]
tests/test_authentication.py ........................................... [  2%]
........................................................................ [  4%]
................................................                         [  6%]
tests/test_authorization.py ........................                     [  7%]
tests/test_csrf.py ........................................              [  9%]
tests/test_decorator.py ....                                             [  9%]
tests/test_encode.py ...............                                     [  9%]
tests/test_events.py .......................................             [ 11%]
tests/test_exceptions.py ...........                                     [ 11%]
tests/test_httpexceptions.py ........................................... [ 13%]
................                                                         [ 13%]
tests/test_i18n.py ............................................          [ 15%]
tests/test_integration.py .............................................. [ 17%]
........................................................................ [ 20%]
......                                                                   [ 20%]
tests/test_location.py ..                                                [ 20%]
tests/test_paster.py ........                                            [ 20%]
tests/test_path.py ..................................................... [ 22%]
............................                                             [ 23%]
tests/test_predicates.py ............................................... [ 25%]
........................                                                 [ 26%]
tests/test_registry.py ....................................              [ 27%]
tests/test_renderers.py ................................................ [ 29%]
..............                                                           [ 30%]
tests/test_request.py .................................................. [ 32%]
.....                                                                    [ 32%]
tests/test_response.py ...............                                   [ 32%]
tests/test_router.py ................................................... [ 34%]
.........                                                                [ 35%]
tests/test_scripting.py .............                                    [ 35%]
tests/test_security.py ...............................................   [ 37%]
tests/test_session.py .................................................. [ 39%]
..........................................                               [ 40%]
tests/test_settings.py ...................                               [ 41%]
tests/test_static.py ................................................... [ 43%]
...                                                                      [ 43%]
tests/test_testing.py .................................................. [ 45%]
.............................                                            [ 46%]
tests/test_threadlocal.py ........                                       [ 46%]
tests/test_traversal.py ................................................ [ 48%]
[...]

PR #3762, opened 2024-06-17
Allow InstancePropertyHelper to accept properties with names on Python 3.13+

hroncok:
Fixes https://github.com/Pylons/pyramid/issues/3761

### DIFF
--- a/src/pyramid/util.py
+++ b/src/pyramid/util.py
@@ -110,7 +110,8 @@ class InstancePropertyHelper:
         if name is None:
             if not hasattr(callable, '__name__'):
                 raise ValueError(
-                    'missing __name__, must specify "name" for property'
+                    'missing __name__, must specify "name" for property '
+                    'on Python < 3.13'
                 )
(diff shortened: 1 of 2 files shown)


mmerickel:
We don't currently have 3.13 in the test suite so I can't trust the cicd here. Perhaps you could add the config for that to this PR and focus it on getting Pyramid working in 3.13?

As a side note, isn't `cgi` removed in 3.13 such that webob doesn't work on 3.13 so you're hosed anyway until we fix that?

hroncok:
> Perhaps you could add the config for that to this PR and focus it on getting Pyramid working in 3.13?

I can try, but no promises.

> As a side note, isn't cgi removed in 3.13 such that webob doesn't work on 3.13 so you're hosed anyway until we fix that?

I use [legacy-cgi](https://pypi.org/project/legacy-cgi/) for that.

hroncok:
The updated tox config works locally. Hopefully, it also works on the CI (needs your approval).

PR #3766, opened 2024-08-30
Fix p* scripts to produce an error when given a bad config_uri

kpinc:
When a bad settings URI (usually a path to a non-existant file) is
  supplied to one of pyramid's command line programs, and in general
  when there is a reason why a settings file cannot be used, produce
  an error message instead of a traceback.

  Applications which import pyramid.paster must handle the
  plaster.exceptions.PlasterError exception themselves.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,14 @@ Bug Fixes
 
   See https://github.com/Pylons/pyramid/pull/3742
 
+- When a bad settings URI (usually a path to a non-existant file) is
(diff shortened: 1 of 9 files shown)


kpinc:
I have changed the github workflow to get this to pass CI.  This change is unrelated to the purpose of the patch.  (macos-11 is no longer offered.)

kpinc:
To see what this fixes, try:`pserve /tmp/nonexistantfile`
You get a traceback without this patch.

PR #3779, opened 2025-02-18
add HttpHeaderCSRFStoragePolicy

silenius:
Add a `HttpHeaderCSRFStoragePolicy` to persists the CSRF token in an HTTP header

### DIFF
--- a/src/pyramid/csrf.py
+++ b/src/pyramid/csrf.py
@@ -161,6 +161,55 @@ class CookieCSRFStoragePolicy:
         )
 
 
+@implementer(ICSRFStoragePolicy)
+class HttpHeaderCSRFStoragePolicy:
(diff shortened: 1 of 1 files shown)


luhn:
This CSRF policy pulls from `request.headers['X-CSRF-Token']` by default and `check_csrf_token` (the module-level function) pulls from the same, so it ends up a tautology.

luhn:
Hard to say, I don't really understand where or how one might use this policy.  Can you give a sample use case?

silenius:
I just want to store the CSRF token in an header rather than in the session or in a cookie.  I'm using JWT token and it is often recommended to store the JWT token in a secure onlyHTTP same-site cookie and the CSRF token in header

PR #3783, opened 2025-05-27
updated build setup to modern pyproject.toml form.

ChrisBarker-NOAA:
I've moved everything from setup.py and setup.cfg to pyproject.toml.

It seems to build fine :-)

There doesn't seem to be a clear way to specify test and docs requirements, so I added:

`requirements_test.txt' and `requirements_docs.txt`

Or maybe tox takes care of that.

I also didn't update any pinned versions....

[though  I changed min Python version to 3.9, and added 3.13 to the tox test list]

I'm not set up to properly run tox, so maybe there's something more to do there, but I kept it as it was, other than the python versions in the test matrix.


pyproject.toml
```diff
@@ -2,6 +2,84 @@
...
+pserve = "pyramid.scripts.pserve:main"
+pshell = "pyramid.scripts.pshell:main"
+proutes = "pyramid.scripts.proutes:main"
+pviews = "pyramid.scripts.pviews:main"
+ptweens = "pyramid.scripts.ptweens:main"
+prequest = "pyramid.scripts.prequest:main"
+pdistreport = "pyramid.scripts.pdistreport:main"
+
+[project.entry-points."paste.server_runner"]
+tomatoes = "spam:main_tomatoes"
```
merwok, line 66: example left in?

pyproject.toml
```diff
@@ -2,6 +2,84 @@
...
+readme = {file = "README.rst", content-type = "text/x-rst"}
+
+requires-python = ">=3.9"
+dependencies = [
+    "hupper >= 1.5",  # ignore_files support
+    "plaster",
+    "plaster_pastedeploy",
+    "setuptools",
+    "translationstring >= 0.4",  # py3 compat
+    "venusian >= 1.0",  # ``ignore``
```
merwok, line 29: These inline comments are not very clear

pyproject.toml
ChrisBarker-NOAA: oops. thanks.

I'm a bit surprised that that didn't give me an error when built, but oh well.

pyproject.toml
ChrisBarker-NOAA: I copy and pasted those from the setup.py -- they should be reviewed an updated for the present needs, but I didn't feel qualified to do that.

pyproject.toml
merwok: Makes sense to me: nothing happens at build time except the entry points being recorded for run-time use.

It would be different for scripts entry-points.

ChrisBarker-NOAA:
OK -- all looking good. Note that the gitHub workflows aren't running, so could be broken -- that needs approval from a maintainer.

Also -- it looks like the docs are being built on RTD, and also in a gitHub workflow -- is one of these obsolete?

Finally -- I suspect the tox config could use some updating, but that's not my wheelhouse.

.readthedocs.yaml
```diff
@@ -10,8 +10,8 @@ formats:
...
   - epub
 python:
+
   install:
     - method: pip
-      path: .
-      extra_requirements:
-        - docs
+      path: .[doc]
+    # - requirements: requirements_docs.txt
```
stevepiercy, line 17:
```suggestion
```

stevepiercy:
> Also -- it looks like the docs are being built on RTD, and also in a gitHub workflow -- is one of these obsolete?

If I understand correctly, the workflow is for testing that docs will build, whereas the RTD config is for deployment of docs after merging the PR.

ChrisBarker-NOAA:
hmm -- looks like the workflow still needs to be approved, so I can't see if the CI will work.

But in any case, it's a tox config issue, I I really don't know tox.

I did add the test dependencies, and that has helped some (locally) -- though it seem that tox should be able to get those from the pyproject.toml,. if we can find the right configuration.

ChrisBarker-NOAA:
> If I understand correctly, the workflow is for testing that docs will build, whereas the RTD config is for deployment of docs after merging the PR.

Ahh -- that makes sense, yes.

pyproject.toml
```diff
@@ -2,6 +2,101 @@
...
+        "Programming Language :: Python :: Implementation :: CPython",
+        "Programming Language :: Python :: Implementation :: PyPy",
+        "Framework :: Pyramid",
+        "Topic :: Internet :: WWW/HTTP",
+        "Topic :: Internet :: WWW/HTTP :: WSGI",
+        "License :: Repoze Public License",
+]
+
+[tool.setuptools.package-data]
+mypkg = ["*.txt", "*.rst"]
```
merwok: Is `mypkg` a paste error?

pyproject.toml
```diff
@@ -2,6 +2,103 @@
...
+wsgiref = "pyramid.scripts.pserve:wsgiref_server_runner"
+cherrypy = "pyramid.scripts.pserve:cherrypy_server_runner"
+
+[project.entry-points."pyramid.pshell_runner"]
+python = "pyramid.scripts.pshell:python_shell_runner"
+
+
+[project.urls]
+Homepage = "https://trypyramid.com"
+Documentation = "https://docs.pylonsproject.org/projects/pyramid/en/latest/"
```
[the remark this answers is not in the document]
mmerickel: I would want the links to be correct in each branch that we release so I think I'd want the instructions in the RELEASING document to be updated to include maintaining these links in each branch.

tox.ini
```diff
@@ -16,7 +22,7 @@ extras =
 setenv =
     COVERAGE_FILE=.coverage.{envname}
 
-[testenv:py312-cover]
+[testenv:py313-cover]
 commands =
     python --version
     pytest --cov {posargs:}
```
mmerickel: similar to other changes that I just merged into main this is a little out of date now and needs to be updated. Specifically the envlist at least and let's leave the py312-cover to match what the CI config is currently doing.

tox.ini
```diff
@@ -1,13 +1,19 @@
...
 
 isolated_build = true
 
 [testenv]
+deps =
+    pytest
+    pytest-cov
+    webtest
+    zope.component
+    coverage
```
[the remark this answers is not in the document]
mmerickel: At the very least the extras = testing down below on line 23 should already be handling these dependencies, right? So this "deps" section can just be removed? This is related to improperly renaming the new extra as "test" instead of "testing".

pyproject.toml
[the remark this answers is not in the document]
ChrisBarker-NOAA: OK -- I put a hard-coded version number in for the Documentation.

since the version is also hard-coded in the pyproject.toml, this isn't so bad.

I also added a little script to check that it's correct -- you could put this in the CI to catch errors.

`check_docs_url.py`

Or update it to actually set the Documentation url (though that's a bit trickier as the build in tomllib can't write toml :-(.

.readthedocs.yaml
```diff
@@ -10,8 +10,7 @@ formats:
   - pdf
   - epub
 python:
+
   install:
     - method: pip
-      path: .
-      extra_requirements:
-        - docs
+      path: .[doc]
```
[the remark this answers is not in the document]
ChrisBarker-NOAA: I think I just fixed this: t it's called "docs" in the tox file now, and:

```
[project.optional-dependencies]
docs = [
    "sphinx >= 8",
```

so:
```
python:
  install:
    - method: pip
      path: .[docs]
```

That seems to be how to do it -- here and in tox.

But I'm not sure how to test.

ChrisBarker-NOAA:
Getting there -- thanks!

Issue #3731, opened 2023-08-11
pkg_resources is deprecated


luhn:
As of [setuptools v67.5.0](https://setuptools.pypa.io/en/latest/history.html#v67-5-0), a deprecation warning is raised when importing pkg_resources:

> DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html

Apparently this has been replaced by [importlib.resources](https://docs.python.org/3/library/importlib.resources.html) and [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html), which is available since Python 3.7, as well as [packaging](https://packaging.pypa.io/en/stable/).

This may be fairly tricky to remove, as `src/pyramid/config/assets.py` digs pretty deep into the pkg_resources API, namely it registers `PackageOverrides` with `pkg_resources.register_loader_type`.  However I can't figure out *why* this call is necessary, I can't find where this loader would be used—Maybe something to do with `AssetResolver`?

The other issue is that pkg_resources would unzip the whole package into a tmp directory when calling `resource_filename`, whereas importlib just extracts that single file.  `pyramid.path.package_path` depends on this behavior.  Possible solutions:

- Deprecate `pyramid.path.package_path`, and by extension `pramid.path.caller_path`, `pyramid.asset.asset_spec_from_abspath`, and `pyramid.resource.resource_path_from_abspath`.
- Change contract so `pyramid.path.package_path` and it's dependents don't work with zip files (or returning zip "path")
- Write an implementation that extracts package to tmp directory, mimicking pkg_resources behaviior.

mmerickel:
The issue is around `config.override_asset` which overrides how `pkg_resources` works on a threadlocal basis such that any `pkg_resources` (and `AssetResolver`) invocations find the appropriate packages. When I have looked, `importlib.resources` does not have hooks to implement this functionality and is going to be a sticking point in what Pyramid can do about this issue without requiring setuptools.

I'm not really aware of places that importlib.metadata would be helpful in pyramid but those have been very easy to replace in any package where I've done it to drop setuptools for that purpose.

luhn:
It may be possible to override importlib machinery?  A custom `Loader` can provide a `get_resource_reader` method, which then can return a `ResourceReader` object.   https://docs.python.org/3.8/library/importlib.html#setting-up-an-importer

[...]

PR #3787, opened 2025-08-12
Constrain setuptools to <81

luhn:
`pkg_resources` is deprecated and `setuptools>=81` is slated to remove it entirely.  Preventing setuptools from upgrading to v81 will buy time while #3731 is unresolved.


luhn:
Closing in favor of #3795

PR #3792, opened 2025-08-25
Add `get_spec` to `IPackageOverrides`

luhn:
While working on #3791, I ran into an issue:  The cache buster was diving into the guts of `PackageOverrides`, stymying my attempts to refactor that.

https://github.com/Pylons/pyramid/blob/eb0e01a0fd7aff84c345b1c43bd4fa9d2fe04441/src/pyramid/config/views.py#L2338-L2347

In particular, `get_path` and `pkg_name` should be considered private members `PackageAssetSource`/`FSAssetSource`, because they are undocumented and aren't directly tested.

This PR is an attempt to formalize those into a public, tested method called `get_spec`.

I also added `get_spec` to the `IPackageOverrides` interface.  This seemed necessary, as if we want to call that method after querying for the utility, it should be in the interface.

This comes with some minor behavior changes as well: The original `_bust_asset_path` implementation did not check for file existence (it always took the first matching override source), whereas the new implementation does in order to match the behavior of overrides elsewhere.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,13 +78,10 @@ Backward Incompatibilities
 - Add `get_spec` method to `IPackageOverrides`.
   See https://github.com/Pylons/pyramid/pull/3792
 
-- When using a `cache buster`_ with `asset overrides`_, the cache buster will
+- When using a cache buster with asset overrides, the cache buster will
   find the first existing file in the override stack, rather than taking the
   first override regardless of whether the file exists or not.
   See https://github.com/Pylons/pyramid/pull/3792
-
-.. _cache buster: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#cache-busting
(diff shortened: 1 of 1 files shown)


CHANGES.rst
```diff
@@ -75,6 +75,17 @@ Backward Incompatibilities
...
+- Add `get_spec` method to `IPackageOverrides`.
+  See https://github.com/Pylons/pyramid/pull/3792
+
+- When using a `cache buster`_ with `asset overrides`_, the cache buster will
+  find the first existing file in the override stack, rather than taking the
+  first override regardless of whether the file exists or not.
+  See https://github.com/Pylons/pyramid/pull/3792
+
+.. _cache buster: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#cache-busting
+.. _asset overrides: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#overriding-assets
```
mmerickel: I don't think I want these links to exist in the CHANGES document. They'll be handy to add to the whats-new however which can be done with internal references to the docs.

PR #3767, opened 2024-10-26
Add Python 3.13, remove Python 3.8 and update CI Workflow

digitalresistor:
Takes care of the following things:

- Updates the CI workflow to include Python 3.13 and testing on a variety of more platforms/including newer ones
- Updates tox.ini to clean it up syntax wise
- Marks some tests that now pass on Python 3.13 as `skipif`

PR #3793, opened 2025-08-25
Fix Github Actions

luhn:
Builds are waiting for 24hrs then timing out.  I believe it's because the images we're using are no longer supported.


luhn:
There's definitely some discussion to be had here.

> Bumps images to `-latest` variants

Could have also done `ubuntu-24.04` etc but I think it's a good thing if our runners keep on the latest images.

> Removes architecture from matrix, defaulting to architecture of the runner.

Because macos runners are now arm64, we would have needed to add that to the architecture list and expand our exclusions, at that point excluded runners would outnumber the rest.  Seems better to just go with whatever architecture the runner OS is.

The only real thing we're losing here is x86 tests.  They were only supported on Windows anyways, and I'm not sure if x86 is even relevant anymore.

> Removes `--cov` from tests.

Not sure why this was there, I can't think what it adds that the dedicated coverage test doesn't already handle.

luhn:
Just realized this conflicts quite a bit with #3767, I'll defer to that one.

luhn:
#3767 has been merged, so closing this one.

PR #3815, opened 2026-03-29
Resolve asset APIs

luhn:
This adds `Configurator.resolve_asset` and `Request.resolve_asset` methods.  For #3731.

Considerations for `Configurator.resolve_asset`

- Opted to put implementation in `config/__init__.py`, so it's next to `maybe_dotted` and `absolute_asset_spec`.  Could also make sense to put in `config/assets.py`.
- Set `Configurator.asset_resolver` at init-time rather than instantiating `AssetResolver` on-demand when the method is invoked.

Considerations for `Request.resolve_asset`:

- Was not sure best place to put it.  There wasn't a relevant mixin, so I put it in the base class.  Could add potentially add `AssetMethodsMixin`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ unreleased
 Features
 --------
 
+- Add `Configurator.resolve_asset` and `Request.resolve_asset` methods.  See
+  https://github.com/Pylons/pyramid/pull/3745.
+
 Bug Fixes
 ---------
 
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,9 @@ unreleased
 Features
 --------
 
+- Add `Configurator.resolve_asset` and `Request.resolve_asset` methods.  See
+  https://github.com/Pylons/pyramid/pull/3745.
```
lgedgar: Should reference pull/3815 instead?

luhn: Whoops, not sure how that happened.

PR #3816, opened 2026-03-29
Reorganize resolvers

luhn:
This PR moves `Resolver` subclasses and `IAssetDescriptor` implementations to `pyramid.resolver`.  The remaining contents of `pyramid.path` gets moved to `pyramid._path`, so that `pyramid.path` can expose the same API without a circular dependency.

In a future change, `AssetResolver` passes `IPackageOverrides` into `PkgResourceAssetDescriptor`.  This means that `AssetResolver` now requires `registry`, which falls back to `get_current_registry()`.  This results in a circular dependency:  `pyramid.path` now imports `pyramid.threadlocal` imports `pyramid.registry` imports `pyramid.path`.

Ways I considered resolving this:

- Make `registry` a mandatory argument for `AssetResolver`.
- Make `registry` optional with no threadlocal fallback.  (overrides just won't work)
- Move `AssetResolver` et al to new package and remove from `pyramid.path` entirely.
- Move `AssetResolver` et al to new package and put `AssetResolver` import at the very bottom of `pyramid.path`—Almost works but importing `pyramid.resolver` first still gives circular import.
- Move `AssetResolver` et al to new package and use `pyramid._path` to keep `AssetResolver` in `pyramid.path`.

The last one (this PR) seems like the only feasible solution, even though `_path.py` is a bit icky.


luhn:
@mmerickel would we want to mark `pyramid.path.AssetResolver` et al as deprecated?  I'm hesitant because that's a pretty big deprecation, but if sometime in the future if we want to get from `_path.py` back to `path.py` that'll be necessary.

ChrisBarker-NOAA:
Thanks for pushing this forward! 

Let us know if it would be helpful to test this with our actual apps before we make it final...

luhn:
Resolved merge conflicts.

luhn:
@mmerickel, I was noodling on this a bit and thought of a couple other feasible solutions to the circular dependencies:

1.  Define constructor functions that import lazily, something like:

```python
def AssetResolver(self, package=CALLER_PACKAGE):
    from .resolver import AssetResolver as cls
    return cls(package)
```

Obviously this isn't a perfect replacement, but should cover basic usecases.

2. Leave a legacy implementation of `AssetResolver` in `pyramid.path`.  I really don't love this but it does give a better bw compat than 1.

I think I still prefer the solution in this PR, plus deprecating `pyramid.path.AssetResolver` if we can.

This is the hairiest PR by far, should be easier going on from here.

